PR #20, opened 2015-01-07
add possibility to get raw request body json

nizsheanez:


olivere:
Can you please elaborate why exactly you need access to `searchSource`?

If you e.g. need access to the generates request body, maybe we should add a method for that.
If you e.g. need to pass your own `SearchSource` instance, maybe we add a setter for it in `SearchService`.

I'm hesistant to open it up because then people might get a reference to it, start using `SearchSource.Size()` directly and complain why `SearchService.Size()` did change as well.

nizsheanez:
Exactly here I need something like method Source() which will return raw result json. I need possibility pass it to http response of my service

nizsheanez:
hello, i'm about this: https://github.com/olivere/elastic/commit/9d576a11aa5aadb59b33cd5550f3eb5daed55706#diff-059dc0366ae402fcb2b70107da19b22dR55

I mean Source() method not setter but getter. Like 
```
func Source() interface{}  {
   return s.searchSource.Source()
}
```

olivere:
All methods in `SearchService` are setters, and I want to keep it like that.

Can you create your query e.g. via `SearchSource` and set it via
`SearchService.SearchSource()`? Then you can use `SearchSource.Source()` to
get the raw query. `SearchService` forwards most calls to `SearchSource`
anyway, so most of them are for convenience.

Here's what I mean:

```go
ss := NewSearchSource()
fmt.Printf("%v", ss.Source())  // or encode to JSON
...
res, err := client.Search().SearchSource(ss).Do()
...
```

nizsheanez:
i agree what you talking about, but i see that `Source()` is setter: 
https://github.com/olivere/elastic/commit/9d576a11aa5aadb59b33cd5550f3eb5daed55706#diff-059dc0366ae402fcb2b70107da19b22dR55

olivere:
Yes. Most structs in Elastic are there to configure stuff, via setters, then execute (e.g. via `Do()` or `Source()`). [Builders](http://en.wikipedia.org/wiki/Builder_pattern), basically.

Can you elaborate what you're trying to do (and cannot do with Elastic)?

nizsheanez:
i'm trying write json of request to stdout. 
I think need to get result of `SearchSource.source.Source()` and pass it to `json.Marshal` 
And `client.Search().Source()` must return it (because all objects in Elastic has method `Source()` which return this kind of `interface{}` ready for `json.Marshal`)
But now method `Source()` is setter. 

Or you mean that i need to create my source, set it, and then set it. 

```go
ss := elastic.NewSearchSource()
search := client.Search().SearchSource(ss).Query(elastic.NewMatchAllQuery()).Fields("a","b")
rawResponse, err = json.Marshal(ss.Source())  // or encode to JSON
if needToReturnRawResponse {
   return rawResponse
} else {
   res, err := search.Do()
}
```

Just i remember you don't wan't to make getter for source:
```
I'm hesistant to open it up because then people might get a reference to it, start using SearchSource.Size() directly and complain why SearchService.Size() did change as well.
```

and this method `client.Search().SearchSource(ss)` - also open it. 

For me here only need possibility to call method  `SearchSource.Source()` through `SearchService`. Maybe better to keep `SearchSource` closed if now we don't have use case for to open it?

olivere:
I'm sorry, maybe I still don't quite understand your logging needs.

I thought the best for your use case would be to create your own `SearchSource`, print it to `os.Stdout` (or whatever you need to do with it), and then pass it into the `SearchService` via the setter `SearchService.SearchSource(*SearchSource)` and execute the search:

```go
// Log the request to stdout
ss := elastic.NewSearchSource().Query(elastic.NewMatchAllQuery()).Fields("a","b").From(0).Size(10)
data, _ := json.Marshal(ss.Source()) // encode SearchSource as JSON (before passing it to Search)
fmt.Fprintf(os.Stdout, "%s", string(data))
...
// Execute the search
res, err := client.Search().SearchSource(ss).Do()
...
```

Regarding your sample code, I'm a bit confused. Why do you set the search source first and then set the query? That will override the query in search source. The idea of `SearchSource` is to create the complete search query for you (i.e. its query, from/size, filters, sorting etc.). Many methods in `SearchService` just pass the parameter to `SearchSource`, e.g. `SearchService.Query(...)`:

```go
type SearchService struct {
  searchSource *SearchSource
  ...
}
...
func (s *SearchService) Query(query Query) *SearchService {
	s.searchSource = s.searchSource.Query(query) // pass the query to search source
	return s
}
```

Which method in `SearchService` did I open up? I just added two setters to `SearchService`, i.e. `SearchService.SearchSource()` and `SearchService.Source()`. 

P.S.: You can also cosntruct the request manually and pass it to `SearchService` via `SearchService.Source(interface{})`, in case you need to do that.

P.P.S.: I'm still thinking about adding a generic `log.Logger` to `Client`, but that probably won't help you.

nizsheanez:
"All methods in `SearchService` are setters, and I want to keep it like that." - aha, maybe it's good patter also. 

Actually, i agree with everything. It's enough. I will do as you described.

olivere:
Okay. BTW I didn't invent this stuff. `SourceSource` is basically [`SearchSourceBuilder`](https://github.com/elasticsearch/elasticsearch/blob/master/src/main/java/org/elasticsearch/search/builder/SearchSourceBuilder.java#L63) from Elasticsearch.

Cheers :-)

PR #34, opened 2015-01-25
Adding boosting query, plus test

mdzor:


olivere:
Thank you. That looks almost complete.

I have one issue though: According to the Elasticsearch Java source, there can be only one negative/positive query (see [BoostingQueryBuilder.java](https://github.com/elasticsearch/elasticsearch/blob/master/src/main/java/org/elasticsearch/index/query/BoostingQueryBuilder.java#L53)). Right? You accept a slice of queries. ES doesn't seem to accept an array, just an object (i.e. a single query). See the accompanying [BoostingQueryFilter.java](https://github.com/elasticsearch/elasticsearch/blob/master/src/main/java/org/elasticsearch/index/query/BoostingQueryParser.java) as well.

mdzor:
Agreed ! I just fixed it :+1:

olivere:
Sorry for the delay.

PR #39, opened 2015-02-11
Feature Get/Put mappings

tiago4orion:
Added MappingsService and CreateMappingsService.

Basic implementation of the mapping feature. I don't know if this is the best implementation. If you think that needs some improvements on the code, please, let me know and maybe I can improve that.

Usage:

```go

// Create the mappings struct
// Use .Body() to set this object mapping
// Use .BodyString() to set a raw JSON string mapping
mappings := map[string]TypeMappings{
    "tweet": TypeMappings{
        Properties: map[string]FieldProperty{
            "location": FieldProperty{
                Type: "geo_point",
            },
            "suggest_field": FieldProperty{
                 Type:                       "completion",
                 Analyzer:                   "simple",
                 Payloads:                   true,
                 PreserveSeparators:         true,
                 PreservePositionIncrements: true,
                 MaxInputLength:             50,
            },
            "tags": FieldProperty{
                 Type: "string",
            },
        },
    },
}

// Creating the mappings
mappingsOk, err := client.CreateMappings().Index("elastic-test").Type("tweet").Body(mappings).Do()

if err != nil || mappingsOk["acknowledged"] != true {
     panic(err)
}

// Getting the mappings
result_mappings, err := client.Mappings().Indices("elastic-test").Do()

if err != nil {
    panic(err)
}

eq := reflect.DeepEqual(mappings, result_mappings)

if !eq {
	panic(errors.New("Mappings are unequal!"))
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,8 +285,8 @@ Here's the current API status.
 - [x] Delete index
 - [x] Indices exists
 - [x] Open/close index
-- [ ] Put mapping
-- [ ] Get mapping
+- [x] Put mapping
+- [x] Get mapping
 - [ ] Get field mapping
 - [ ] Types exist
 - [ ] Delete mapping
--- a/client.go
+++ b/client.go
@@ -354,3 +354,13 @@ func (c *Client) ClusterHealth() *ClusterHealthService {
 func (c *Client) ClusterState() *ClusterStateService {
 	return NewClusterStateService(c)
 }
+
+// Mappings retrieves the mappings
+func (c *Client) Mappings() *MappingsService {
+	return NewMappingsService(c)
+}
+
+// CreateMappings create the mappings
+func (c *Client) CreateMappings() *CreateMappingsService {
+	return NewCreateMappingsService(c)
+}
--- a/create_mappings.go
+++ b/create_mappings.go
@@ -1,0 +1,189 @@
+package elastic
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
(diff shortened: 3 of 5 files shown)


olivere:
Hi there. Thanks for your contributions.

I see that you need the mappings API. That isn't something I haven't been looking at. I was trying to focus on search and runtime issues firstly, while cluster and index administration is still something we do on the command line (with tools like [es](https://github.com/olivere/es) or curl for example). Of course, that doesn't mean I want to include it, so thanks again for your work.

Let me explain why I refrain from adding things like `TypeProperty` or `FieldProperty` to Elastic.

The Elasticsearch API is huge. It has about 100 API endpoints, roughly 600 parameters spread all over the API, and god knows how many data structures. And it is moving fast. I don't see a way of re-implementing all these structs in a Go client as you did with `TypeProperty` and `FieldProperty`. If you look at the official clients, they don't do that either.

What you can expect from Elastic is to have access to the API endpoints and a (general) way of performing requests against an ES server. The official clients, again, even have a separation into a low-level and a high-level API. Elastic currently is a mix of both. I'm trying to align it more towards their structure. So you can now use the low-level API to call any API endpoint in ES by doing it like this:

req

tiago4orion:
Hi there, thanks for the quick reply!

I understand your position. I'm developing a service that requires strong control of elasticsearch mappings, then I need some mappings utility. If you want I can change that code for use interface{} instead... 

While the elastic do not support that, I will stick with this implementation.

I didn't know the "es" tool, very nice! I will take a look at it, maybe it can help me!

Thanks!!

PR #40, opened 2015-02-17
Simple support for source filtering

corsc:
I was trying to filter the fields that where returned from a Get() and found the Fields() was not helpful to as it changed the response format.

After some searching I found a way to retrieve only the fields I requested: http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-request-source-filtering.html

There is more work needed here to support all of the options but this seems like a nice start.

### DIFF
--- a/get.go
+++ b/get.go
@@ -22,6 +22,7 @@ type GetService struct {
 	routing    string
 	preference string
 	fields     []string
+	source     []string
 	refresh    *bool
 	realtime   *bool
 }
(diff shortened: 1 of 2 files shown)


olivere:
Thanks for the PR. Give me a little more time to review this stuff...

olivere:
Hi there! Let me know if [this commit](https://github.com/olivere/elastic/commit/258273ebe6b3597044e701979bc0b4b7fd10cada) doesn't address your problem. Cheers...

corsc:
Yup that will get it done. Thanks!

PR #41, opened 2015-02-19
Add a Get Mapping service

tam7t:
Add ability to get mapping from `elasticsearch`.

Usage:
```go
client, err := elastic.NewClient(http.DefaultClient)
if err != nil {
	panic(err)
}

mapping, err := client.GetMapping().Get(index).
	Debug(true).  // print request and response to stdout
	Pretty(true). // pretty print request and response JSON
	Do()
fmt.Printf(os.Stdout, "Got mapping: %s", mapping)
```

The mapping is just returned as a `string` - no JSON unmarshalling.

### DIFF
--- a/client.go
+++ b/client.go
@@ -171,6 +171,12 @@ func (c *Client) Ping() *PingService {
 	return NewPingService(c)
 }
 
+// GetMapping returns a service to get the mapping of an index.
+func (c *Client) GetMapping() *GetMappingService {
+	builder := NewGetMappingService(c)
+	return builder
+}
+
 // CreateIndex returns a service to create a new index.
 func (c *Client) CreateIndex(name string) *CreateIndexService {
 	builder := NewCreateIndexService(c)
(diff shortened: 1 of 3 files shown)


olivere:
Thanks for your work.

I'm in the middle of a larger refactoring regarding the services (I try to align more to the practices of the official ES clients). I will revisit the GetMapping service after that. Once that refactoring is done, you will get a lot more APIs/services (they can be auto-generated; see the [generate-api branch](https://github.com/olivere/elastic/tree/generate-api) in case you're interested).

tam7t:
Thanks I'll look out for that. Also the tests I wrote fail on 1.3.8 apparently due to a difference in whats returned with you request a mapping.

olivere:
The mapping API is now in with [this commit](https://github.com/olivere/elastic/commit/9b2a76d66d3a609a7e22ccf1bbcf1479bdd6ed0d). However, notice that it's not in master but the [client-refactorings branch](https://github.com/olivere/elastic/tree/client-refactoring) that'll become master eventually (hopefully by the end of the week, if tests on production are all green). If you want to test, you're welcome :-)

PR #42, opened 2015-02-20
enable user to set UseNumber option during decoding the search result

jiangjin:
First, `elastic` lib looks quite advanced, we'd like to use it. Thanks a lot! @olivere 

But when we try to use it, find a problem during decoding the json result from elasticsearch, usually for the "_timestamp" (for example: "_timestamp": 144654564565). We suggest to use decode.UseNumber() option, so we try to add this `useNumber` toggle to enable the lib user to choose this option.

### DIFF
--- a/search.go
+++ b/search.go
@@ -27,6 +27,7 @@ type SearchService struct {
 	preference   string
 	types        []string
 	debug        bool
+	useNumber    bool
 }
 
 // NewSearchService creates a new service for searching in Elasticsearch.
@@ -38,6 +39,7 @@ func NewSearchService(client *Client) *SearchService {
 		searchSource: NewSearchSource(),
 		debug:        false,
 		pretty:       false,
+		useNumber:    false,
 	}
 	return builder
 }
@@ -106,6 +108,13 @@ func (s *SearchService) Pretty(pretty bool) *SearchService {
 // when calling Do.
 func (s *SearchService) Debug(debug bool) *SearchService {
 	s.debug = debug
+	return s
+}
+
+// UseNumber enables the caller to unmarshal a number into an interface{}
+// as a Number instead of as a float64 when calling Do.
+func (s *SearchService) UseNumber(useNumber bool) *SearchService {
+	s.useNumber = useNumber
 	return s
 }
 
@@ -360,7 +369,14 @@ func (s *SearchService) Do() (*SearchResult, error) {
 	}
 
 	ret := new(SearchResult)
-	if err := json.NewDecoder(res.Body).Decode(ret); err != nil {
+
+	decoder := json.NewDecoder(res.Body)
+
+	if s.useNumber {
+		decoder.UseNumber()
+	}
+
+	if err := decoder.Decode(ret); err != nil {
 		return nil, err
 	}
 	return ret, nil


jiangjin:
@olivere i see,
i don't notice the branch `client-refactoring`, you mean the decoder inside client structure?
that looks pretty good. thanks!

olivere:
Yes, the decoder can be set with `client.SetDecoder(...)` now. I'd be happy if you could test the feature. It's available in [this branch](https://github.com/olivere/elastic/tree/client-refactoring) that will eventually become master. The decoder feature is also described in the [wiki](https://github.com/olivere/elastic/wiki), including an example.

jiangjin:
@olivere `client.SetDecoder(...)` works fine, and I like this design.
but for branch `client-refactoring`, the method to sniffer each ES node looks nice; but I have a problem to use it, in the beginning to set up sniffer, you use ES API `/_nodes/http` to request each ES node info. But in my case, we deploy each ES node into docker container, ES node app listening to 9200 port, but the port bind by docker container to the real host is different. ES API `/_nodes/http` still returns port as 9200, because in the app level ES only knows it is listening on 9200. So in this case sniffer doesn't work, maybe you know any solution to this? or we'd better to have option to disable sniffer. just a suggestion, what do u think?

olivere:
Oh, okay. Thanks for testing those edge cases.

Can you create a client with specifying host and port manually, like `NewClient(http.DefaultClient, "http://<ipaddr>:<port>")` where `<ipaddr>` and `<port>` are the IP address and port number of ES (inside the container)?

Otherwise: Can you please send me the output from `curl -XGET <esserver>/_nodes/http?pretty=1`, from inside your container please?

Thank you.

jiangjin:
@olivere actually I use NewClient(http.DefaultClient, "http://<ip>:<port>") where ip and port is the ip address of ES tribe node,

`curl -XGET <esserver>/_nodes/http?pretty=1` to tribe node, tribe node will forward to different es cluster, in my case the nodes of es cluster is inside of docker container. the output is like

```
{
  "cluster_name" : "es_tribe_node",
  "nodes" : {
    "node_1_id" : {
      "name" : "Judas Traveller",
      "transport_address" : "inet[/10.8.1.3:31001]",
      "host" : "8056816c2421",
      "ip" : "192.168.0.11",
      "version" : "1.4.3",
      "build" : "36a29a7",
      "http_address" : "inet[/10.8.1.3:9200]",
      "http" : {
        "bound_address" : "inet[/0:0:0:0:0:0:0:0%0:9200]",
        "publish_address" : "inet[/10.8.1.3:9200]",
        "max_content_length_in_bytes" : 104857600
      }
    },
    "node_2_id" : {
      "name" : "Glamor",
      "transport_address" : "inet[/10.8.1.5:31001]",
      "host" : "5e918b8fc139",
      "ip" : "192.168.0.54",
      "version" : "1.4.3",
      "build" : "36a29a7",
      "http_address" : "inet[/10.8.1.5:9200]",
      "http" : {
        "bound_address" : "inet[/0:0:0:0:0:0:0:0%0:9200]",
        "publish_address" : "inet[/10.8.1.5:9200]",
        "max_content_length_in_bytes" : 104857600
      }
    }
  }
}
```

olivere:
Hmm... I'm still struggling to understand the problem. Is it that ES inside the container reports an IP address on `/_nodes/http` that is unavailable from the outside? So the sniffing process uses an internal IP that won't be available from outside the container? That would mean that in your example, `192.168.0.11` is the IP of `node_1_id` from the outside, whereas `10.8.1.3` is the IP from inside the container?

That should be solvable, I guess, by setting `network.host` (or [some other network setting](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/modules-network.html)) so that ES binds to an IP address that is reachable from the outside as well (in your case e.g. `192.168.0.11` for the `node_1_id`).

Sorry for asking all this stuff... I'm afraid I'm not too familiar with Docker...

olivere:
I'm thinking of disabling the sniffing process by default. If the consumer of the client wants a sniff to happen, one could call that immediately after creating the client. So there wouldn't be too much magic going on by default.

I'll look into the official clients and see what they are doing by default.

olivere:
@jiangjin I have come up with solution that enables you to configure the client before starting up. I think it's more flexible, but it is breaking the API. Maybe you want to try it with your problem. [See here](https://github.com/olivere/elastic/wiki/Client) for how to use the new API for `NewClient` to set up and configure your client accordingly.

jiangjin:
@olivere the sniffer in Client will use "http_address" returned by calling "/_node/http" (which in my case, it is "inet[/10.8.1.3:9200]", so the ip addr can be reachable, but the port can't be accessed. Because in my case "10.8.1.3" is host ip, "31000" is the host port binding to "9200" for es container.)

the problem is during `NewClient`, I set the host using a elasticsearch tribe node info. But the sniffer calls "/_node/http" get the data node host info... but elasticsearch as in the app level, it doses't know its listening port 9200 be mapped to other port of the real host by docker.

olivere:
Hmm. Okay; I suppose the official clients would fail in this scenario, too. At least I haven't found anything that they do that could resolve this situation (the JavaScript client has the option to disable sniffing on startup, but the Ruby client does not).

Anyway: Sniffing can now be disabled as you see in my previous post. I'm confident that'll help.

jiangjin:
yep, for this case, currently this only solution is to disable the sniffer. So there is only connection to tribe node. I'm not clear about if the tribe node would distribute uniformly the load to each node of that es cluster. I've not yet found any doc.
thanks!

ak47wyh:
@olivere please help me
spring.data.elasticsearch.cluster-name=elasticsearch
spring.data.elasticsearch.cluster-nodes=127.0.0.1:9300

$ curl -XGET http://127.0.0.1:9200/_nodes/http?pretty
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   486  100   486    0     0    740      0 --:--:-- --:--:-- --:--:--   740{
  "cluster_name" : "elasticsearch",
  "nodes" : {
    "Ua1k4rWsTNKP6M5xObho1g" : {
      "name" : "Master",
      "transport_address" : "127.0.0.1:9300",
      "host" : "127.0.0.1",
      "ip" : "127.0.0.1",
      "version" : "2.3.4",
      "build" : "e455fd0",
      "http_address" : "127.0.0.1:9200",
      "http" : {
        "bound_address" : [ "127.0.0.1:9200" ],
        "publish_address" : "127.0.0.1:9200",
        "max_content_length_in_bytes" : 104857600
      }
    }
  }
}
NoNodeAvailableException[None of the configured nodes are available: []
]
	at org.elasticsearch.client.transport.TransportClientNodesService.ensureNodesAreAvailable(TransportClientNodesService.java:290)
	at org.elasticsearch.client.transport.TransportClientNodesService.execute(TransportClientNodesService.java:207)
	at org.elasticsearch.client.transport.support.TransportProxyClient.execute(TransportProxyClient.java:55)
	at org.elasticsearch.client.transport.TransportClient.doExecute(TransportClient.java:288)
	at org.elasticsearch.client.support.AbstractClient.execute(AbstractClient.java:359)
	at org.elasticsearch.client.support.AbstractClient$IndicesAdmin.execute(AbstractClient.java:1226)
	at org.elasticsearch.action.ActionRequestBuilder.execute(ActionRequestBuilder.java:86)
	at org.elasticsearch.action.ActionRequestBuilder.execute(ActionRequestBuilder.java:56)
	at org.elasticsearch.action.ActionRequestBuilder.get(ActionRequestBuilder.java:64)
	at com.didispace.domain.esTest2.test(esTest2.java:59)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.springframework.test.context.junit4.statements.RunBeforeTestMethodCallbacks.evaluate(RunBeforeTestMethodCallbacks.java:75)
	at org.springframework.test.context.junit4.statements.RunAfterTestMethodCallbacks.evaluate(RunAfterTestMethodCallbacks.java:86)
	at org.springframework.test.context.junit4.statements.SpringRepeat.evaluate(SpringRepeat.java:84)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.runChild(SpringJUnit4ClassRunner.java:252)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.runChild(SpringJUnit4ClassRunner.java:94)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.springframework.test.context.junit4.statements.RunBeforeTestClassCallbacks.evaluate(RunBeforeTestClassCallbacks.java:61)
	at org.springframework.test.context.junit4.statements.RunAfterTestClassCallbacks.evaluate(RunAfterTestClassCallbacks.java:70)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.run(SpringJUnit4ClassRunner.java:191)

olivere:
You seem to be using Java. This is a Go library.

If you're using Go, please create a gist with a working code example of what you're trying to do.

PR #44, opened 2015-02-22
Made HasActive public so I can wait for it to come alive again

foeken:
I am using a worker process that shoots things into ES. It needs to wait until the connection is alive again before continuing.

### DIFF
--- a/client.go
+++ b/client.go
@@ -40,7 +40,7 @@ type Client struct {
 
 	mu        sync.RWMutex // mutex for the next two fields
 	activeUrl string       // currently active connection url
-	hasActive bool         // true if we have an active connection
+	HasActive bool         // true if we have an active connection
 }
 
 // NewClient creates a new client to work with Elasticsearch.
@@ -101,7 +101,7 @@ func (c *Client) dumpResponse(resp *http.Response) {
 func (c *Client) NewRequest(method, path string) (*Request, error) {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
-	if !c.hasActive {
+	if !c.HasActive {
 		return nil, ErrNoClient
 	}
 	return NewRequest(method, c.activeUrl+path)
@@ -120,7 +120,7 @@ func (c *Client) pinger() {
 
 // pingUrls iterates through all client URLs. It checks if the client
 // is available. It takes the first one available and saves its URL
-// in activeUrl. If no client is available, hasActive is set to false
+// in activeUrl. If no client is available, HasActive is set to false
 // and NewRequest will fail.
 func (c *Client) pingUrls() {
 	for _, url_ := range c.urls {
@@ -132,14 +132,14 @@ func (c *Client) pingUrls() {
 			if err == nil {
 				defer res.Body.Close()
 				if res.StatusCode == http.StatusOK {
-					// Everything okay: Update activeUrl and set hasActive to true.
+					// Everything okay: Update activeUrl and set HasActive to true.
 					c.mu.Lock()
 					defer c.mu.Unlock()
 					if c.activeUrl != "" && c.activeUrl != url_ {
 						log.Printf("elastic: switched connection from %s to %s", c.activeUrl, url_)
 					}
 					c.activeUrl = url_
-					c.hasActive = true
+					c.HasActive = true
 					return
 				}
 			} else {
@@ -153,7 +153,7 @@ func (c *Client) pingUrls() {
 	// No client available
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	c.hasActive = false
+	c.HasActive = false
 }
 
 // ElasticsearchVersion returns the version number of Elasticsearch
--- a/client_test.go
+++ b/client_test.go
@@ -42,8 +42,8 @@ func TestFindingActiveClient(t *testing.T) {
 	if len(client.urls) != 2 {
 		t.Fatalf("expected 2 default client urls, got: %v", client.urls)
 	}
-	if !client.hasActive {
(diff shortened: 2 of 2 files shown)


olivere:
Opening up these things is probably not a good idea. It may introduce races etc.

Also, `hasActive` is going away in the near future. It is replaced by a more robust mechanism introduced in [this branch](https://github.com/olivere/elastic/tree/client-refactoring). That will eventually be merged into master.

foeken:
But you provide no way to see if the connection is alive right? Nor has that been introduced in that branch... The feature request is a reasonable one in my opinion, perhaps a discussion might be more fruitful if you did not shoot it down right away :disappointed:

olivere:
I see what you are after. But introducing a problem to get a feature is not the right decision, don't you think?

I have the same problem in my code, and I solve it on the client side. I watch for errors, time out, and retry. Maybe you can do the same?

olivere:
BTW: Here's a snippet of code we use to process bulk requests. They can fail, and we simply try again. Maybe this helps tackling your problem.

```go
// bulkCommit commits bulk actions to Elasticsearch. When an error
// occurs, it retries the operation several times before failing.
// An error is returned via the Response channel.
func bulkCommit(r *IndexRequest) error {
	sleepSeconds := int64(1)
	numRetries := r.numRetries

	startTime := time.Now().UTC()
	reqid := utils.NewUUID()

	for {
		res, err := r.service.Do()
		if err != nil {
			logf("bulk [%s]: indexing failed with %v; %d more retries", reqid, err, numRetries)
			numRetries -= 1
			if numRetries < 0 {
				logf("bulk [%s]: indexing finally failed with %v", reqid, err)
				r.Response <- err
				return err
			}
			logf("bulk [%s]: indexing sleeps for %s", reqid, time.Duration(sleepSeconds)*time.Second)
			time.Sleep(time.Duration(sleepSeconds) * time.Second)
			sleepSeconds *= 2
		} else {
			if !res.Errors {
				logf("bulk [%s]: indexing succeeded with no errors in %s [took=%dms]", reqid, time.Now().UTC().Sub(startTime), res.Took)
			} else {
				logf("bulk [%s]: indexing failed with errors in %s [took=%dms]", reqid, time.Now().UTC().Sub(startTime), res.Took)
				for _, item := range res.Failed() {
					logf("bulk [%s]: item %s failed with: %v", reqid, item.Id, item.Error)
				}
			}
			break
		}
	}

	r.Response <- nil
	return nil
}
```

foeken:
I did not say I did not agree with the fact that this way it introduces a race condition, but that can be solved by adding a dedicated reader for the variable that can later be rewritten to look into the new mechanism. 

I do think that it would be nice to hook into the already existing code instead of having to check for the ErrNoClient code specifically after each call and build the timeout code in my worker too.

I had similar code that I wanted to replace by a guard that would hold everything at one spot...

foeken:
I prefer this above my worker:

```go
for {
  if p.Client.HasActiveConnection() {
    break
  } else {
    log.Println("[ES] Waiting for connection to recover.")
    time.Sleep(5 * time.Second) // or exponential back-off
  }
}
```

But I can also fake it by setting a variable whenever an `ErrNoClient` err is returned. But then I'd have to do another call to see if the connection has been recovered. I tried `Ping()` but that did not default to the client url but to the `default_url` so i had to find another call to use as a check. It just feels like the client should expose that state.

olivere:
While there might be a way to expose whether or not we have an active connection, there is no way to make this safe other than calling out to ES. The connection could be down again right after `HasActiveConnection()` reported true back to you.

Can you take a look into the official clients and see if they expose that functionality? Then I might review their implementation and try to find a way to get it in.

To use the ping service with another URL, set it via the [`URL(...)`](https://github.com/olivere/elastic/blob/master/ping.go#L53) function. To get the status of your cluster, use the [cluster nodes info API](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/cluster-nodes-info.html#cluster-nodes-info) or the [cluster health status](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/cluster-health.html).

foeken:
If I look at other libraries that do connection management internally (like AMQP) I see a pattern of a `notifyClose` channel and a `notifyConnect` channel. Would a solution like that be acceptable?

olivere:
I understand. However, I know that AMQP uses a persistent TCP connection to the server. With a persistent connection, you can detect if you have a connection or not. Elasticsearch on the other hand uses a HTTP request/response mechanism. You typically only see failure when requests go wrong.

foeken:
I think the main issue here is that you cannot use a check like this to really determine if the connection is up or down. I understand that. It's more that I want to link the state of my worker to the state of your pinger.

Currently I added something like this:

```go
// returns a boolean indicating if the connection is alive.
func (c Client) HasActiveConnection() bool {
  c.mu.Lock()
  defer c.mu.Unlock()
  return c.hasActive
}
```

But I am aware that adding this might cause confusion if other see this as a trustworthy check instead of a representation of the pinger state :)

olivere:
For your scenario, I'd implement a pinger in my own code that does e.g. a `HEAD /` request on your ES servers and report availability back with a channel. BTW: You can use the Ping service for that; just set the URL individually and set `HttpHeadOnly(true)` on the service. That shouldn't be too hard, and it is also very lightweight for the ES servers as well.

But in any case, to make your application robust, you should be prepared for errors to be returned anytime you interact with ES, either through Elastic or directly via HTTP.

foeken:
It's already quite robust, this was just a way to clean up the logs and not have the worker try when we can already know it won't work :)

I'll implement it myself for now :) Tnx!

PR #46, opened 2015-03-01
Restore FetchSourceContext feature

corsc:
This seems to have got lost in the shuffle to v2.

Thanks again!


olivere:
Oh noes. I suck at merging ;-)  Sorry for that.

PR #50, opened 2015-03-11
add geo distance filter for filtering point by distance

banux:
Hi

For a projet i need to filter location by distance.
I add the filter based on geo_polygon filter.
I test with lastest 1.3 and 1.4  official docker images and soon i production on 1.1.1.

Hope it is all good.

Regards
Bastien


olivere:
Thank you, @banux. I only fixed some minor stuff and completed the properties of the filter according to [GeoDistanceFilterBuilder.java](https://github.com/elastic/elasticsearch/blob/master/src/main/java/org/elasticsearch/index/query/GeoDistanceFilterBuilder.java).

When Travis think's it's okay, it should be ready in a few minutes.

PR #54, opened 2015-03-23
Added test that validates marshaling and unmarshaling of the aggregation...

SvenW:
...s object in the SearchResult struct, make sure it doesn't return a base64 encoded string.

Verifies that the branch solving issue-51 actually works


olivere:
Thank you. I moved the test into its own method.

I merged the branch into master. Travis is currently doing it's job.

PR #60, opened 2015-04-05
Fix example in readme

arcz:
Fix example


olivere:
A http.Client parameter is only necessary on version 1 of Elastic. Version 2 has a different way of configuring settings. The Wiki explains that. Are you still on version 1?

Issue #58, opened 2015-04-01
Problems on connect


olivere:
@dashaus, I copied it over from #57:

Hi,
I have the same problem here:

panic: main: conn db: no Elasticsearch node available

```
goroutine 1 [running]:
log.Panicf(0x84de50, 0x11, 0xc2080c7e90, 0x1, 0x1)
    /usr/local/go/src/log/log.go:314 +0xd0
main.init·1()
    /Users/emilio/go/src/monoculum/init.go:40 +0x348
main.init()
    /Users/emilio/go/src/monoculum/main.go:334 +0xa4

goroutine 526 [select]:
net/http.(*persistConn).roundTrip(0xc2088ad1e0, 0xc2086a9d50, 0x0, 0x0, 0x0)
20:30:13 app         |  /usr/local/go/src/net/http/transport.go:1082 +0x7ad
net/http.(*Transport).RoundTrip(0xc20806c000, 0xc2086f6000, 0xc20873ff50, 0x0, 0x0)
20:30:13 app         |  /usr/local/go/src/net/http/transport.go:235 +0x558
20:30:13 app         | net/http.send(0xc2086f6000, 0xed4f18, 0xc20806c000, 0x21, 0x0, 
20:30:13 app         | 0x0)
    /usr/local/go/src/net/http/client.go:219
20:30:13 app         |  +0x4fc
net/http.(*Client).send(0xc08b00, 0xc2086f6000, 0x21
20:30:13 app         | , 0x0, 0x0)
    /usr/local/go/src/net/http/client.go:142 +0x15b
20:30:13 app         | net/http.(*Client).doFollowingRedirects(0xc08b00, 0xc2086f6000, 0x97cd00, 0x0, 0x0, 0x0)
20:30:13 app         |  /usr/local/go/src/net/http/client.go:367 +0xb25
net/http.(*Client).Do(0xc08b00, 0xc2086f6000, 0xc20873fce0, 0x0, 
20:30:13 app         | 0x0)
    /usr/local/go/src/net/http/client.go
20:30:13 app         | :174 +0xa4
github.com/olivere/elastic.(*Client).sniffNode(0xc208659d10, 0xc208569920, 0x15
20:30:13 app         | , 0x0, 0x0, 0x0)
20:30:13 app         |  /Users/emilio/go/src/github.com/olivere/elastic/client.go:543
20:30:13 app         |  +0x16a
20:30:13 app         | 
github.com/olivere/elastic.func·014(0xc208569920, 0x15
[...]

PR #62, opened 2015-04-06
#58: increasing the timeout of sniffer

dashaus:


olivere:
Fixed in [this commit](https://github.com/olivere/elastic/commit/41814e5b8028921ab2ca93340361385253403396).

PR #67, opened 2015-04-26
Remove Reindexer.recreateTargetIndex

nwolff:
That method allowed the caller to specify the index settings for
number_of_shards and number_of_replicas, but otherwise prevented the
user from specifying a mapping.

Removed the method and the two settings.
Now the caller must setup the target index any way he likes before
invoking the reindexer (this is in conformity with the reindexer
package comment)

### DIFF
--- a/reindexer.go
+++ b/reindexer.go
@@ -29,8 +29,6 @@ type Reindexer struct {
 	sourceIndex, targetIndex   string
 	query                      Query
 	bulkSize                   int
-	shards                     int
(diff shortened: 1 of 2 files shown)


olivere:
You are correct. It has unexpected side-effects for users. It is better to leave it up to the end-user to set up the target index. Sorry for that.

nwolff:
Thanks for merging, and thanks for a library that tackles the huge feature-set of elasticsearch.

PR #68, opened 2015-04-26
Make reindexing work with child documents

nwolff:
The scanner now requests the _parent field (on top of the default _source field).

When the reindexer finds that field in the hit, it sets parent on the IndexRequest.

In index_test: Added a new 'comment' type that is a child of 'tweet' together with its mapping.
Added a comment instance to the test dataset.

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -32,6 +32,7 @@ type ScanService struct {
 	indices   []string
 	types     []string
 	keepAlive string
+	fields    []string
 	query     Query
 	size      *int
 	pretty    bool
@@ -88,6 +89,13 @@ func (s *ScanService) Scroll(keepAlive string) *ScanService {
 // available before expiration (e.g. "5m" for 5 minutes).
 func (s *ScanService) KeepAlive(keepAlive string) *ScanService {
 	s.keepAlive = keepAlive
+	return s
+}
+
+// Fields specifies the fields the scan query should load.
+// By default fields is nil so _source is loaded
+func (s *ScanService) Fields(fields ...string) *ScanService {
+	s.fields = fields
 	return s
 }
 
@@ -156,6 +164,9 @@ func (s *ScanService) Do() (*ScanCursor, error) {
 	}
 	if s.size != nil && *s.size > 0 {
 		params.Set("size", fmt.Sprintf("%d", *s.size))
+	}
+	if s.fields != nil {
+		params.Set("fields", strings.Join(s.fields, ","))
 	}
 
 	// Set body
--- a/index_test.go
+++ b/index_test.go
@@ -22,11 +22,17 @@ const (
 		"number_of_replicas":0
 	},
 	"mappings":{
-		"tweet":{
+		"_default_": {
 			"_timestamp": {
 				"enabled": true,
 				"store": "yes"
 			},
+			"_ttl": {
+				"enabled": true,
+				"store": "yes"
+			}
+		},
+		"tweet":{
 			"properties":{
 				"tags":{
 					"type":"string"
--- a/index_test.go
+++ b/index_test.go
@@ -39,6 +39,11 @@ const (
 					"payloads":true
 				}
 			}
+		},
+		"comment":{
+			"_parent": {
+				"type":	"tweet"
+			}
 		}
 	}
 }
@@ -58,6 +63,16 @@ type tweet struct {
 
 func (t tweet) String() string {
 	return fmt.Sprintf("tweet{User:%q,Message:%q,Retweets:%d}", t.User, t.Message, t.Retweets)
+}
+
+type comment struct {
+	User    string    `json:"user"`
+	Comment string    `json:"comment"`
+	Created time.Time `json:"created,omitempty"`
+}
+
+func (c comment) String() string {
+	return fmt.Sprintf("comment{User:%q,Comment:%q}", c.User, c.Comment)
 }
 
 func isTravis() bool {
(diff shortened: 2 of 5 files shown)


olivere:
Wonderful. Thank you. Give me some time to review.

Why did you disable the reindexer tests?

olivere:
Ah. Forget it. Tests are okay. I'm probably in weekend mode...

nwolff:
When adding this I looked at the source of stream2es, a tool that is able to copy child documents across indexes. They include _source and _routing in the scan request: https://github.com/elastic/stream2es/blob/86bcf684034d912e25cce754e8ea77b6fcb5ff9d/src/stream2es/es.clj#L96 and the reuse that data when indexing. I thought that adding _parent always made sense, but arbitrarily decided not to include _routing. After reading the ES documentation more attentively I now think that _routing should be added as well: https://www.elastic.co/blog/customizing-your-document-routing

I agree with you that the ScanService should take a Fields method, and the Reindexer should be the one telling the ScanService to include those fields in all reindexing operations.

Please tell me if you agree with the approach and I will add a commit to this PR.

olivere:
Thanks for clarifying.

I agree with your approach. Let me add two more thoughts.

The official Python client also has [helpers for search & scan and bulk indexing](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/helpers/__init__.py). It's a very good source, and it adds other fields as well in certain cases (just skimmed over it right now). Maybe we should have a look at what they're doing.

There is a [generate-api](https://github.com/olivere/elastic/tree/generate-api) branch which consumes the [REST API definition](https://github.com/elastic/elasticsearch/tree/master/rest-api-spec) from Elasticsearch and uses `go:generate` to generate service templates. It's not quite complete, could get some more love, but it's a good approximation to the final service. Plus: It adds a REST API compliant service. I just added it a few weeks ago, so services like `SearchService` were created before having the generator. Why am I saying this? Because you might use it to see what's also still missing in the `ScanService` besides the `Fields` setter.

nwolff:
Still in the process of writing tests for this PR, as part of the test I am using the GetService and found a bug, the fix is in PR: https://github.com/olivere/elastic/pull/69  
If you are able to merge it I would then like to rebase all the commits in this PR (68) on top of the new master but am not sure github will be able to continue associating the rebased commits to this PR?

olivere:
Already merged. Looks good. :-) If they only had a JSON spec for the response types.

nwolff:
So I rebundled the initial commits into one that includes the Fields method on ScanService.

I looked at https://github.com/olivere/elastic/tree/generate-api and saw search.gen.go with 40 attributes for SearchService, it didn't help me deciding which methods to add to the ScanService.

I looked at the python client and almost none of the fields they inject into the index-request can be obtained from a search or scan query https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/helpers/__init__.py#L30-L31 . This is not a bug, but is misleading.

I am currently finalizing a second commit for this PR, that allows the user to optionally take control of the mapping from SearchHit to BulkIndexRequest. This is similar to the expand_action_callback of the python client: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/helpers/__init__.py#L43

olivere:
Ha. You probably had to add them all to be compliant ;-)

Re the PR: No worries. Take your time. We don't need to rush anything. Your commits are high-quality. Thanks for your work until now.

olivere:
Re the `expand_action`: Isn't it for passing [these parameters](http://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-versioning) to the Bulk API? They don't have to be part of the response; they could've been added by the function passed into `expand_action_callback`.

nwolff:
You are right: I hadn't thought of the use of expand_action outside the scope of a reindexing.

nwolff:
OK so the PR is ready for review:
- The reindexer by default honors the _routing and _parent fields, I believes this is a sane default.
- If a user wants to do something different he can specify the fields the Scanner should return and/or specify the HitToIndexRequest function. There is a test-case that demonstrates copying the _ttl of each document from source to target. The HitToIndexRequest function can also be used to perform random transformation of data while reindexing, like for instance modifying felds for data migrations.

scan.go
```diff
@@ -157,6 +165,9 @@ func (s *ScanService) Do() (*ScanCursor, error) {
 	if s.size != nil && *s.size > 0 {
 		params.Set("size", fmt.Sprintf("%d", *s.size))
 	}
+	if s.fields != nil {
```
olivere: Maybe better to use `len(s.fields) > 0` here? If you use `service.Fields()` (no args), then it would result in `?fields=` in the query string. I don't know if ES would ignore that.

olivere:
The `HitToIndexRequest` function, enabling transformations, is really helpful in my use cases. Will review and merge tomorrow morning if it's looking good at second sight.

scan.go
olivere: Forget it. It'll work fine... I really shouldn't comment when I'm tired.

olivere:
I think you are right. It's called a `Reindexer` after all. So we go for option c) and leave your PR as is. If we want a transformation utility, we can write a `Transformer.

If you agree with that, I'll merge now.

nwolff:
I'd like to explore idea d) today because it may not just bring us extra flexibility, but also a much clearer separation of labor between the reindexer (provides the machinery) and the handler (do the actual work). 
I'd do this in another branch. At the end of the day you would merge either what we have now or design d).

olivere:
Okay. I like idea d). Much better than what I've had in mind. It's just that I don't want Reindexer to do too much as it might get too complex. Time for exploration, I guess :-)

nwolff:
Alernative d) is in https://github.com/olivere/elastic/pull/71

olivere:
Yes. I've seen that. Thank you. I'd prefer PR #71. What about you?  (Also see a last question regarding naming in #71).

nwolff:
I also prefer #71. I linked the two PRs together so that github will remember the discussions here brought us to #71

PR #69, opened 2015-04-27
Fix GetResult Fields type

nwolff:
Using Fields in GetService resulted in a marshalling error.

Fixed the type of Fields to map[string]interface{}.

In get_test:
- Added a test that request two fields and asserts their value.
- Simplified two tests that were testing more than the GetService,
and were issuing unnecessary Flush calls.

### DIFF
--- a/get.go
+++ b/get.go
@@ -212,12 +212,12 @@ func (b *GetService) Do() (*GetResult, error) {
 // -- Result of a get request.
 
 type GetResult struct {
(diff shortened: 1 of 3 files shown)


olivere:
Thank you. For reference: [`GetResult` in Elasticsearch](https://github.com/elastic/elasticsearch/blob/1.5/src/main/java/org/elasticsearch/index/get/GetResult.java#L45).

PR #71, opened 2015-04-28
Generic handler

nwolff:
The reindexer now takes a HitHandler than can emit any kind of BulkRequest
We provide a CopyToTargetIndex function to handle the common case of straighforward copying from one index to another.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1219,5 +1219,5 @@ func (c *Client) NodesInfo() *NodesInfoService {
 // http://www.elastic.co/guide/en/elasticsearch/guide/current/reindex.html
 // for more information about reindexing.
 func (c *Client) Reindex(sourceIndex, targetIndex string) *Reindexer {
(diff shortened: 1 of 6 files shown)


olivere:
:thumbsup: Much better than before. Flexible and simpler design. Is `Reindexer` still a good name for this?

olivere:
If we're going to stay with `Reindexer`, maybe it's better to use `ReindexerFunc` for symmetry? (There also is a `ReindexerProgressFunc`)

olivere:
Okay, I see that you've renamed the callback. I think it's better now.

Are you still working on some details? Otherwise I'll merge it.

nwolff:
I think it's good like this. Thank you for pushing the enveloppe

olivere:
It was a pleasure working with you. Thank you.

nwolff:
Same here, cheers!

Issue #74, opened 2015-05-21
Function score doesn't require a function and can specify a weight


autochthe:
As of 1.4, elasticsearch has added a weight parameter to function score functions.  Futhermore, function score queries do not require a score function if they have both a filter and a weight.

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html

olivere:
Hi there! Sorry for the delay. I'm currently on OOO and will review your PR ASAP. Thanks for working on this.

olivere:
I'm currently looking into this. If you have the time, maybe we could figure this out together.

I model the Go code after the Java client as I found the documentation misleading or incomplete sometimes (it probably has to). In other words: The Java code is the gold standard for me. [Here's the code](https://github.com/elastic/elasticsearch/tree/master/src/main/java/org/elasticsearch/index/query/functionscore) that implements the Function score feature in Java. Important classes are [FunctionScoreQueryBuilder](https://github.com/elastic/elasticsearch/blob/master/src/main/java/org/elasticsearch/index/query/functionscore/FunctionScoreQueryBuilder.java) which basically is [FunctionScoreQuery](https://github.com/olivere/elastic/blob/master/search_queries_fsq.go) in Elastic and [ScoreFunctionBuilders](https://github.com/elastic/elasticsearch/blob/master/src/main/java/org/elasticsearch/index/query/functionscore/ScoreFunctionBuilders.java) which is what you find in [search_queries_fsq_score_funcs.go](https://github.com/olivere/elastic/blob/master/search_queries_fsq_score_funcs.go).

However, when comparing the Java code with your implementation, I see that in Java the weights are only applied to the score functions whereas in your PR you add them to the `FunctionScoreQuery`. Do we agree that it is the correct thing to add them to the score functions?

BTW: I already have a patch available here (if we agree). So no need to change your PR. It fixes some other issues as well (like the `min_score` field that came with 1.5.0+, deprecation of `FactorFunction` and addition of `WeightFactorFunction`).

olivere:
Closed this for now. Feel free to open a new issue if this doesn't work for you. Cheers.

PR #75, opened 2015-05-22
Support function score weight parameter

autochthe:
Closes olivere/elastic#74


autochthe:
Note that first commit's build failed only on one node for a reason unrelated to this PR

olivere:
Should be achieved with [this commit](https://github.com/olivere/elastic/commit/b2de90d6773a73454428f4a44241b2fb860b6387).

PR #78, opened 2015-05-27
Allow sorting the results of an elasticsearch scroll

nwolff:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html

- Added two methods (a simple one + a comprehensive one) to specify the sort.
- Only set the search_type to scan if there is no sort applied.
- A unit test that shows a scroll request with a sort

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -34,6 +34,7 @@ type ScanService struct {
 	keepAlive string
 	fields    []string
 	query     Query
+	sorts     []SortInfo
 	size      *int
 	pretty    bool
 }
@@ -104,6 +105,24 @@ func (s *ScanService) Query(query Query) *ScanService {
 	return s
 }
 
+// Sort the results by the given field, in the given order.
+// Use the alternative SortWithInfo to use a struct to define the sorting.
+// See http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-request-sort.html
+// for detailed documentation of sorting.
+func (s *ScanService) Sort(field string, ascending bool) *ScanService {
+	s.sorts = append(s.sorts, SortInfo{Field: field, Ascending: ascending})
+	return s
+}
+
+// SortWithInfo defines how to sort results.
+// Use the Sort func for a shortcut.
+// See http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-request-sort.html
+// for detailed documentation of sorting.
+func (s *ScanService) SortWithInfo(info SortInfo) *ScanService {
+	s.sorts = append(s.sorts, info)
+	return s
+}
+
 func (s *ScanService) Pretty(pretty bool) *ScanService {
 	s.pretty = pretty
 	return s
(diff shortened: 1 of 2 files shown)


olivere:
LGTM. Will merge soon. Thanks.

nwolff:
Am having a hard time understanding why there are currently 2 services in the codebase:
- https://github.com/olivere/elastic/blob/master/scroll.go#L16
- https://github.com/olivere/elastic/blob/master/scan.go#L29

I was familiar with the ScanService so that's where I added the Sort and SortInfo methods. So now the "ScanService" is performing an elasticsearch scroll (but not an elasticsearch scan): this makes things even more confusing to consumers of the API.

Please shed some light on the goal of the two different services, and on the apparent naming mismatch.

olivere:
To be honest, I was asking myself the same question recently :-/

I guess it all started by a misunderstanding of mine regarding the scroll/scan situation. Also, when I wrote it, I was probably looking for two slightly different kinds of APIs: One would return search results (like search does), the other would be iterator-like (for bulk operations on large amounts of data).

It is a mess and it would need a serious cleanup. I'm sure I could merge the two without losing any of its functionality.

I have pushed it off for a while, but now that you found it too, I should really spend some time refactoring. Maybe I'll find some time at the weekend.

Sorry.

olivere:
BTW: The comments in [client.go](https://github.com/olivere/elastic/blob/master/client.go#L1072) give some insights into my thoughts. Reading them now, I'm even more confused... oh my.

```
// Scan through documents. Use this to iterate inside a server process
// where the results will be processed without returning them to a client.
func (c *Client) Scan(indices ...string) *ScanService {
  ...
}

// Scroll through documents. Use this to efficiently scroll through results
// while returning the results to a client. Use Scan when you don't need
// to return requests to a client (i.e. not paginating via request/response).
func (c *Client) Scroll(indices ...string) *ScrollService {
  ...
}
```

nwolff:
Thank you for clarifying. FWIW I think the two overlapping services are only a minor annoyance. 
Not sure it warrants a big refactoring that would introduce backward incompatible changes.

olivere:
Hmm... maybe procrastinate even more until ES 2.0 is out. ;-)  There will probably be backwards-incompatible  changes with ES 2.0 anyway which will give me an opportunity for refactoring.

nwolff:
ES 2.0: good plan.
Are you going to merge this commit?

olivere:
It already is, isn't it?

nwolff:
OK thanks, was confused by the github message: "This pull request is closed, but the nwolff:sort-in-scan branch has unmerged commits.".

olivere:
I think I squashed the merge from a local "feature branch" so GitHub couldn't figure out what happened. Old habits die hard... anyway, it's merged. And you're in the fresh [CONTRIBUTORS](https://github.com/olivere/elastic/blob/master/CONTRIBUTORS) file as well. Cheers :-)

PR #82, opened 2015-06-02
use analyzer in match query

deoxxa:


olivere:
Whoops. Sorry, missed that.

deoxxa:
Easy mistake, easy fix - thanks for merging it so quickly!

PR #83, opened 2015-06-10
Fix memory leak NewTicker -> After

cep21:
My long running application is showing a memory leak that profiles to the elastic search client.  Calling time. NewTicker() without calling Stop() on the ticker will create a memory leak.  Replace with After() to resolve.

### DIFF
--- a/client.go
+++ b/client.go
@@ -512,7 +512,7 @@ func (c *Client) sniffer() {
 	for {
 		c.mu.RLock()
 		timeout := c.snifferTimeout
-		ticker := time.NewTicker(c.snifferInterval)
+		ticker := time.After(c.snifferInterval)
 		c.mu.RUnlock()
 
 		select {
@@ -520,7 +520,7 @@ func (c *Client) sniffer() {
 			// we are asked to stop, so we signal back that we're stopping now
 			c.snifferStop <- true
 			return
-		case <-ticker.C:
+		case <-ticker:
 			c.sniff(timeout)
 		}
 	}
@@ -678,7 +678,7 @@ func (c *Client) healthchecker() {
 	for {
 		c.mu.RLock()
 		timeout := c.healthcheckTimeout
-		ticker := time.NewTicker(c.healthcheckInterval)
+		ticker := time.After(c.healthcheckInterval)
 		c.mu.RUnlock()
 
 		select {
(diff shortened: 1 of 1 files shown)


cep21:
My pprof:

```
(pprof) top30 -cum
91.51MB of 127.35MB total (71.86%)
Dropped 735 nodes (cum <= 0.64MB)
Showing top 30 nodes out of 67 (cum >= 6.50MB)
      flat  flat%   sum%        cum   cum%
         0     0%     0%   117.35MB 92.15%  runtime.goexit
   91.51MB 71.86% 71.86%    91.51MB 71.86%  time.NewTicker
         0     0% 71.86%    85.01MB 66.75%  github.com/olivere/elastic.(*Client).healthchecker
         0     0% 71.86%     9.40MB  7.38%  main.(*Server).main
```

From healthchecker
```
         .          .    676:// healthchecker periodically runs healthcheck.
         .          .    677:func (c *Client) healthchecker() {
         .          .    678:	for {
         .          .    679:		c.mu.RLock()
         .          .    680:		timeout := c.healthcheckTimeout
         .    85.01MB    681:		ticker := time.NewTicker(c.healthcheckInterval)
         .          .    682:		c.mu.RUnlock()
         .          .    683:
         .          .    684:		select {
         .          .    685:		case <-c.healthcheckStop:
         .          .    686:			// we are asked to stop, so we signal back that we're stopping now
         .          .    687:			c.healthcheckStop <- true
         .          .    688:			return
         .          .    689:		case <-ticker.C:
         .          .    690:			c.healthcheck(timeout, false)
         .          .    691:		}
         .          .    692:	}
         .          .    693:}
```

olivere:
Hi! I have long-running applications as well but didn't find this issue yet, so thanks for reporting.

I'm not sure about the fix though as [time.After](https://golang.org/src/time/sleep.go?s=3012:3046#L99) simply does `return NewTimer(d).C`. So it won't stop the ticker as well, or will it?

I think the right fix is to insert a `defer ticker.Stop()` after creating the ticker. Do you agree?

olivere:
Sorry. I was too fast: I didn't see the ticker being created inside the for loop (which is because `c.healthcheckTimeout` could have been changed). But the cleanup question remains: Using `time.After` won't stop the leak, right?

cep21:
Thanks for looking at this.  time.After creates a Timer which is different than a ticker.  Tickers forever consume resources as they forever tick.  A timer relinquishes its resource after the time duration.

Note the documentation for http://golang.org/src/time/tick.go?s=706:740#L11 says "Stop the ticker to release associated resources" while there is no such caveat for NewTimer http://golang.org/src/time/sleep.go?s=1986:2018#L56

olivere:
Oh well. I was confusing `time.NewTicker` with `time.NewTimer`. I should have stayed in bed. Will merge in a second.

Thanks for your patience :-)

PR #87, opened 2015-06-24
fix percolation on an existing document

sacheendra:
Percolation on an existing document did not work because id was not being included in the url.


olivere:
Thanks for reporting.

PR #88, opened 2015-06-26
Use capacity instead of length

jun06t:
This doesn't cause error, however this currently generates a bit strange path.
So I fixed as follows.

current path
```
http://localhost:9200/my_blog/,,,,titles,contents,tags,comments/_search
```

fixed path
```
http://localhost:9200/my_blog/titles,contents,tags,comments/_search
```

### DIFF
--- a/search.go
+++ b/search.go
@@ -67,7 +67,7 @@ func (s *SearchService) Index(index string) *SearchService {
 // Indices sets the names of the indices to use for search.
 func (s *SearchService) Indices(indices ...string) *SearchService {
 	if s.indices == nil {
-		s.indices = make([]string, 0)
+		s.indices = make([]string, 0, len(indices))
 	}
 	s.indices = append(s.indices, indices...)
 	return s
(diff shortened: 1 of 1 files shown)


olivere:
Thanks for reporting.

I'm not sure I understand the issue. Can you please send me the code that generates this URL?

```
http://localhost:9200/my_blog/,,,,titles,contents,tags,comments/_search
```

jun06t:
Thanks for reply :)
The following is the sample code which cause above problem.

```
package main

import (
        "log"
        "os"

        "github.com/olivere/elastic"
)

func main() {
                elastic.SetInfoLog(log.New(os.Stderr, "Info", log.LstdFlags)),
                elastic.SetURL("http://localhost:9200"),
                elastic.SetSniff(false),
        )
        if err != nil {
                panic(err)
        }

        query := elastic.NewMatchQuery("value", "test")
        resp, err := client.Search().
                Index("my_blog").
                Types("titles", "contents", "tags", "comments").
                Query(&query).
                Do()
        if err != nil {
                // Handle error
                panic(err)
        }

        log.Println(resp)
}
```

```
$ go run sample.go 
Info2015/06/29 11:30:18 POST http://localhost:9200/my_blog/,,,,titles,contents,tags,comments/_search [status:200, request:0.005s]
```

olivere:
Oh. I see. Thank you.

PR #93, opened 2015-07-01
Update Aggregation struct definitions to take and pass instance via reference

autochthe:
Closes olivere/elastic/#91

### DIFF
--- a/search_aggs_avg.go
+++ b/search_aggs_avg.go
@@ -20,45 +20,44 @@ type AvgAggregation struct {
 	subAggregations map[string]Aggregation
 }
 
-func NewAvgAggregation() AvgAggregation {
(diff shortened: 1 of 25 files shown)


autochthe:
Note that only failed check is the following test in [job 69095606](https://travis-ci.org/olivere/elastic/jobs/69095606)
```
--- FAIL: TestReindexerPreservingTTL-2 (0.09s)
	reindexer_test.go:239: elastic: Error 503 (Service Unavailable)
```

olivere:
These errors happen from time to time with Travis (mostly timing issues from my experience). Restarting the failing job often makes it succeed.

autochthe:
I have restarted the job and all checks have passed!

Issue #94, opened 2015-07-01
Properly unmarshall keyed Range Aggregation


autochthe:
The `rangeAggregationEntry` is currently not exported, which makes the json Marshaller unable to generate a proper json document when marshalling a RangeAggregation.

Fix will be to rename `rangeAggregationEntry` to `RangeAggregationEntry`, which should not break any user code since the type is currently unexported and therefore private to the module.

olivere:
Hi. I'll keep this open and mark it as an enhancement. Marshaling is out of scope for Elastic at the current stage. It's primary use case it to unmarshal and work with results returned from Elasticsearch (see #51 and #64). What is your use case for marshaling data?

autochthe:
This is not a marshalling data marshalling issue, but the way that the library builds the JSON query that is sent to elasticsearch.  My use case for this is that I want to submit a range aggregation query to my elasticsearch cluster.  This is not an enhancement.  The range aggregation query functionality is currently broken.  This is a bug that will persist until the RangeAggregationEntry type is exported from the library (ie. renamed from rangeAggregationEntry).

autochthe:
Ohhhh, how embarrasing.  So I discovered that the problem is in fact none of these things, but that the SearchHits.Aggregations.Range(string) function is not propertly Unmarshalling the "buckets" object in the response from elasticsearch.  The current code seems to expect "buckets" to be an array of AggregationBucketRangeItems objects, while in fact the response is a map of keys to AggregationBucketRangeItems (see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-range-aggregation.html).

I can write a fix for this, but I've been so off base on what the problem was for this whole time that I just want to run this by you and see what you think first.

autochthe:
Ah ok, so I think I've got it now.  The response where "buckets" is a map only happens when you are making a *keyed* Range Aggregation.  So the right thing to do seems to be to make a SearchHits.Aggregations.KeyedRange(string) function that properly unmarshalls the keyed version of the reponse.

olivere:
Don't worry. We'll figure that out. It might take some time, but we'll figure it out eventually.

It might be very possible that there's a bug, especially with the Aggregations. Also: While the Elasticsearch documentation is very nice and informative, my trusted source is the [Elasticsearch Java source code](https://github.com/elastic/elasticsearch). While documentation may sometimes lack certain information, fields etc., the source code never lies. 

But back to your issue.

[...]

PR #95, opened 2015-07-01
Export type `RangeAggregationEntry`

autochthe:
rename `rangeAggregationEntry` to `RangeAggregationEntry`

Closes olivere/elastic#94


autochthe:
This turns out to be unnecessary as the bug is due to another problem

PR #102, opened 2015-07-14
fix boostingQuery to receive multiple valuess

thekostya:
in positive and negative clauses

```json
"boosting": {
                 "positive": [
                          {"term": {"sourceLang": "en"}},
                          {"term": {"targetLang": "ru"}}                     
                 ],
                 "negative": [
                         {"term": {"sourceLang": "ru"}},
                         {"term": {"targetLang": "en"}}                                          
                 ],
                 "negative_boost": 0.2
             }
```

### DIFF
--- a/search_queries_boosting.go
+++ b/search_queries_boosting.go
@@ -10,10 +10,10 @@ package elastic
 // http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-boosting-query.html
 type BoostingQuery struct {
 	Query
-	positiveClause Query
-	negativeClause Query
-	negativeBoost  *float64
-	boost          *float64
+	positiveClauses []Query
+	negativeClauses []Query
+	negativeBoost   *float64
+	boost           *float64
 }
 
 // Creates a new boosting query.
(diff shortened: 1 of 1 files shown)


olivere:
Hello.

According to the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-boosting-query.html) and the [Java API](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/index/query/BoostingQueryBuilder.java#L40), there can be only one positive and negative query.

Have you tried to wrap them inside a [boolean query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html)?

thekostya:
I know documentation doesn't contain anything about it, but it works.

POST /.../.../_search
{
    "query": {
        "bool": {
            "must": [
                {
                    "boosting": {
                        "positive": [
                            {"term": {"services.sourceLang": "en"}},
                            {"term": {"services.targetLang": "ru"}}
                        ],
                        "negative": [
                            {"term": {"services.sourceLang": "ru"}},
                            {"term": {"services.targetLang": "en"}}
                        ],
                        "negative_boost": 0.2
                    }
                }
            ]
        }
    }
}

olivere:
It might even work. The problem is: It is not officially supported as far as I can see. So suppose I merge your PR and the guys at Elastic silently refactor the API. Now your query stops working over night.

I try to be defensive. But you are correct: The docs are vague sometimes. That's why I try to use the Java API as reference. But sometimes the Java API has more features than documented. E.g. the index close API accepts one or more index names while the REST API spec only states one. In that case, the REST API spec seems more if a safe bet.

Why not play it safe and wrap the array in a bool query. It's even more expressive as you are forced to explicitly state if you want the term queries to both match (must) or if one match is acceptable (should). Whereas in your array of term queries: Is that a "must" or a "should" match?

thekostya:
Ok, i understand your point of view. It will be safe to skip my PR.

PR #103, opened 2015-07-17
Add support for search request source filtering

gdm85:
Upstream documentation: https://www.elastic.co/guide/en/elasticsearch/reference/master/search-request-source-filtering.html

I tried to model it around existing code in Explain()

### DIFF
--- a/cluster-test/cluster-test.go
+++ b/cluster-test/cluster-test.go
@@ -17,7 +17,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/olivere/elastic"
+	"github.com/gdm85/elastic"
 )
 
 type Tweet struct {
--- a/example_test.go
+++ b/example_test.go
@@ -12,7 +12,7 @@ import (
 	"reflect"
 	"time"
 
-	"github.com/olivere/elastic"
+	"github.com/gdm85/elastic"
 )
 
 type Tweet struct {
--- a/search.go
+++ b/search.go
@@ -110,6 +110,25 @@ func (s *SearchService) TimeoutInMillis(timeoutInMillis int) *SearchService {
 	return s
 }
 
+// XSource is a list of fields to extract and return from the _source field.
+// With no parameters specified, it will completely disable returning _source.
+func (s *SearchService) XSource(xSource ...string) *SearchService {
+	s.searchSource = s.searchSource.XSource(xSource...)
+	return s
+}
+
+// XSourceInclude is a list of fields to extract and return from the _source field.
+func (s *SearchService) XSourceInclude(xSource ...string) *SearchService {
+	s.searchSource = s.searchSource.XSource(xSource...)
+	return s
+}
+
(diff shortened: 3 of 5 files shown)


olivere:
Hi and thanks for the commit.

In the Java API, there is the [`FetchSourceContext`](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/search/fetch/source/FetchSourceContext.java#L35) for this, which Elastic [already has](https://github.com/olivere/elastic/blob/master/fetch_source_context.go). [Here's an example](https://github.com/olivere/elastic/blob/30b70b8736e737554e35e7dd86d37781cb39fac9/search_source_test.go#L68) of source filtering with `FetchSourceContext`.

However, as the public API comes with these fields as well, I'm thinking about adding it to Elastic. It's sort of a duplicate, but it's not clear from Elasticsearch documentation that you can use something like `FetchSourceContext` for the same thing.

gdm85:
Thanks @olivere for your feedback. I had seen FetchSourceContext, but being new to ElasticSearch I couldn't make up my mind if it was the same as what I was looking for.

As I understand, I can set the FetchSourceContext on the SearchSource, so functionality is already there, closing.

gdm85:
@olivere reopening this as I am unable to use FetchSourceContext with a call like:

```
	searchService := client.Search().
		Query(query).
		Size(size)
```

olivere:
You can use `SearchSource` to build a search, as [illustrated here](https://github.com/olivere/elastic/blob/master/search_test.go#L561). The [SearchService](https://github.com/olivere/elastic/blob/master/search.go) uses the `SearchSource` under its cover. However, you are right in that `SearchService` does forward calls to the `FetchSource` of `SearchSource`.

Let's keep it open and I'll see if I find time to fix that today.

gdm85:
Yeah, but searchSource is not exported (for good reasons too, probably) on a ``SearchService``. I'll leave it to you, there are probably better ways that what I concocted here on this PR.

olivere:
It doesn't have to be exported. You use it like this:

```go
fsc := elastic.NewFetchSourceContext(true).Include("obj1.*", "obj2.*").Exclude("*.description")
src := elastic.NewSearchSource().Query(elastic.NewMatchAllQuery()).FetchSourceContext(fsc)
res, err := client.Search().Index("products").SearchSource(src).Do()
if err != nil {
  t.Fatal(err)
}
```

But you're right. I will add `XSourceInclude` etc. as soon as I get the time. It's in other services as well, so it's a bit larger than your PR. But thanks for bringing it up.

gdm85:
@olivere thanks for your work and providing me these pointers and examples, much appreciated :+1:

olivere:
BTW: You can now direcly access `FetchSource` and `FetchSourceContext` in `SearchService`. So things got a little bit easier.

I think you can use source filtering with 1.7 now in all detail. If not, let me know of the use-cases that don't work for you.

PR #109, opened 2015-07-29
Fixed issue with fuzzy query in completion suggestions

zakthomas:
Found a minor issue with fuzzy query in completion suggestion. 

File : suggester_completion_fuzzy
Line : 143
Code:
	if q.fuzziness != nil {
		suggester["fuzziness"] = q.fuzziness
	}

Suggestion:
    if q.fuzziness != nil {
        suggester["fuzzy"] = map[string]interface{}{"fuzziness": q.fuzziness}
    }

Kindly review it and give suggestions.

### DIFF
--- a/suggester_completion_fuzzy.go
+++ b/suggester_completion_fuzzy.go
@@ -139,9 +139,9 @@ func (q FuzzyCompletionSuggester) Source(includeName bool) interface{} {
 	}
 
 	// Fuzzy Completion Suggester fields
-	if q.fuzziness != nil {
-		suggester["fuzziness"] = q.fuzziness
-	}
+    if q.fuzziness != nil {
+        suggester["fuzzy"] = map[string]interface{}{"fuzziness": q.fuzziness}
+    }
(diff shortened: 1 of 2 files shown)


olivere:
Thanks for your PR. Actually, it's a bit more complicated because there is a [whole structure underneath](https://github.com/elastic/elasticsearch/blob/4c981ff4bfc250080d521af105b5e8589c9fc517/core/src/main/java/org/elasticsearch/search/suggest/completion/CompletionSuggestionFuzzyBuilder.java#L113) the "fuzzy" field now. I will have a look but it might take some days. It's hard to stay on top of the Elasticsearch API.

olivere:
I just pushed release v2.0.4. Would you mind to double-check my changes? Thank you.

Notice that you should use the import path `gopkg.in/olivere/elastic.v2` as the master branch will soon target Elasticsearch 2.0.

olivere:
Re-open if something is still wrong for you.

PR #110, opened 2015-07-29
Make accessing terms stats facets in results possible.

ghaering:
There was no way to to access terms stats facets. Adding the additional fields
and their json mappings is an easy solution to fix this.

### DIFF
--- a/search.go
+++ b/search.go
@@ -454,11 +454,17 @@ type SearchFacet struct {
 	Entries []searchFacetEntry `json:"entries"`
 }
 
-// searchFacetTerm is the result of a terms facet.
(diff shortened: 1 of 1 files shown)


olivere:
Hi! Thanks for the PR. Seems legit :-)

For self-reference: [Source for terms stats for type strings in 1.7](https://github.com/elastic/elasticsearch/blob/1.7/src/main/java/org/elasticsearch/search/facet/termsstats/strings/InternalTermsStatsStringFacet.java#L270)

olivere:
I've just pushed v2.0.3. Notice that you should use the new import path `gopkg.in/olivere/elastic.v2` as the master branch will target Elasticsearch 2.0 soon.

Thanks again.

Issue #119, opened 2015-08-24
Bug with subsequent calls to ScanCursor.Next()


guilherme-santos:
I'm getting a strange follow error when use [Reindexer](https://github.com/olivere/elastic/blob/master/reindexer.go):
```
Error 400 (Bad Request): ElasticsearchIllegalArgumentException[Can't specify parent if no parent field has been configured]
```
I started to debug and I found something. When your module try to insert a document at some point, it add "_parent" and "_routing", but this fields is not present in the original document. 

I'm not sure if you expect [this lines](https://github.com/olivere/elastic/blob/master/scan.go#L305) will be cleared before unmarshall, but it wont. When you call ```ScanCursor.Next()``` ```c.Results``` already be populated, and something strange happen in ```json.Unmarshal``` that add "_parent" and "_routing" to my new hits.

To fix this bug I inserted right before, this couple of lines:
```
c.Results = new(SearchResult)
c.Results.ScrollId = body
```

olivere:
Thanks for reporting! I assume you're working with Elastic 2 (targeting Elasticsearch 1.x), right?

Can you try to extract a complete example/failing test, please? Maybe the existing unit tests in [`reindexer_test.go`](https://github.com/olivere/elastic/blob/master/reindexer_test.go) are a good start.

olivere:
I'm not exactly sure what's going on. But can you try to set `ScanFields(...)` to an `_source` only and try again:

```go
...
reindexer = reindexer.ScanFields("_source")
...
```

Right now, it automatically includes `_source`, `_parent` and `_routing` by default.

olivere:
Just to be sure: Can you double-check your mapping? Does it have a parent-child relationship? If so, you [added a `_parent` to your child mapping](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-parent-field.html), did you?

guilherme-santos:
Answering you questions...
1. Yes, I'm using Elasticsearch 1.7.0 and "gopkg.in/olivere/elastic.v2" in my import
2. Actually I'm using other documents that have ```_parent``` and ```_routing```, for that I cannot set ```ScanFields()```, because default works to the most documents
3. My mappings is right.

Only to be more clear, I have company and employees relationship. some companies are correct in ```c.Results``` and employees are added correctly, but at some point the document's company try index passing ```_parent``` and ```_routing``` without have this fields at ```_source```.

olivere:
Thank you very much for this. I'll have a look.

[...]

PR #120, opened 2015-08-25
Bug #119

guilherme-santos:
I wrote this test to show that bug exists!


olivere:
Fixed in [f4b7e68](https://github.com/olivere/elastic/commit/f4b7e6802f83621ea8a31fde62b9e6eddecb7b11).

PR #136, opened 2015-10-01
Add optional format parameter to date range filter and query

brocaar:
I noticed that the ``format`` parameter was missing in your range query and filter implementations. See also https://github.com/elastic/elasticsearch/commit/462956fd478ba852d05373d9875c5fcd0dd40d28.

### DIFF
--- a/search_filters_range.go
+++ b/search_filters_range.go
@@ -13,6 +13,7 @@ type RangeFilter struct {
 	from         *interface{}
 	to           *interface{}
 	timeZone     string
+	format       string
 	includeLower bool
 	includeUpper bool
 	cache        *bool
@@ -28,6 +29,11 @@ func NewRangeFilter(name string) RangeFilter {
 
 func (f RangeFilter) TimeZone(timeZone string) RangeFilter {
 	f.timeZone = timeZone
+	return f
+}
+
+func (f RangeFilter) Format(format string) RangeFilter {
+	f.format = format
 	return f
 }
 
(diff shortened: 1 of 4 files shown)


olivere:
Thank you. LGTM.

olivere:
Interesting...

Although [`format` is documented here](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html#_date_options), there is no sign of it in the [Java source of 1.7](https://github.com/elastic/elasticsearch/blob/1.7/src/main/java/org/elasticsearch/index/query/RangeQueryBuilder.java). However, it is in the [source code of 2.0](https://github.com/elastic/elasticsearch/blob/2.0/core/src/main/java/org/elasticsearch/index/query/RangeQueryBuilder.java) and [elastic.v3-unstable already has it](https://github.com/olivere/elastic/blob/release-branch.v3/search_queries_range.go#L110).

I'd be surprised if it even works in 1.7?! Anyway, I [asked the Elasticsearch gurus](https://github.com/elastic/elasticsearch/issues/13885).

brocaar:
It is not in the builder classes (I couldn't find a reference to it), but it seems to be in the parser classes, e.g.:
https://github.com/elastic/elasticsearch/blob/1.7/src/main/java/org/elasticsearch/index/query/RangeFilterParser.java#L105. Maybe they forgot to implement it in the client library, but the server is able to parse and use it?

See also the changed files in https://github.com/elastic/elasticsearch/commit/462956fd478ba852d05373d9875c5fcd0dd40d28

olivere:
Oh, thanks. So they have tests for `format`. If they have tests, then it's probably just an oversight in the builder class (which is more than understandable given the size of the ES project).

Okay, so I will push the changes to Elastic in a minute.

Thanks again.

Issue #135, opened 2015-09-28
Implement Source Filtering


autochthe:
Elasticsearch has a way to return partial source documents that is more powerful than the fields parameter.  This would extremely useful functionality.  See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-source-filtering.html

autochthe:
I noticed there is an open PR #103.  I would like to note that for my specific use case, I wish to use filtering during a scan/scroll request, which is functionality supported by Elasticsearch.  I noticed that it is currently possible to perform a regular search request with a FetchSourceContext, but I don't think it is currently possible to set SearchSource manually with ScanService or ScrollService.

olivere:
Hi!

When implementing Elastic I tend to use the Java client as the "source of truth". When initially implementing source filtering I noticed that the Java client always uses the `FetchSourceContext` class under the hood. All functions related to source filtering in Java basically create an instance and forward to that class. So I implemented [`FetchSourceContext`](https://github.com/olivere/elastic/blob/master/fetch_source_context.go) and asked the client programmer set it up correctly and pass it into the services, e.g. [`Get`](https://github.com/olivere/elastic/blob/master/get.go#L88), [`MultiGet`](https://github.com/olivere/elastic/blob/master/multi_get.go#L153), [`MoreLikeThis`](https://github.com/olivere/elastic/blob/master/search_queries_more_like_this.go#L345), [`InnerHit`](https://github.com/olivere/elastic/blob/master/inner_hit.go#L89) etc.

So Elastic has source filtering, but it's probably not as visible as it should be. One example is your use case: How do I use source filtering with `Scan`? I'm not sure it works, but can you try this:

```go
// Init search source
src := elastic.NewSearchSource(...)
...

// Use scan service
scan := client.Scan(...)
scan = scan.Query(src)  // as SearchSource implements the Query interface this should be okay
...
``` 

It's a hack, I know. And I'm thinking of better ways, like directly offering to set the `SearchSource` in `Scan`.

However, I cannot see how to use source filtering with `Scroll` though. Do you have any examples?

autochthe:
[...]

PR #137, opened 2015-10-02
Add FetchSourceContext function to ScanService

autochthe:
Not essential, but a convenient function to have nevertheless.  See #135


olivere:
Already had that in my local changes, no worries.

Issue #124, opened 2015-08-31
How can I wait until elastic is ready?


aanm:
I'm connecting to elastic with the following code:
```go
		c, err := elastic.NewClient(
			elastic.SetURL("http://"+ip+":"+port),
			elastic.SetMaxRetries(10),
			elastic.SetSniff(false),
			elastic.SetErrorLog(l.New(fe, "", l.LstdFlags)),
			elastic.SetInfoLog(l.New(fo, "", l.LstdFlags)),
			elastic.SetTraceLog(l.New(ft, "", l.LstdFlags)),
		)
```
but I always get the following error:
```
2015/08/31 16:01:57 elastic: http://127.0.0.1:9200 is dead
```
Because elasticsearch isn't ready when the program reaches that code.
So is there any option that "waits" until elasticsearch is ready? I'm using `SetMaxRetries` but I'm not sure if there's also an option for timeout or something. Elastic is probably offline or not ready at all to receive connections so it would be great to have something that kept trying if it has failed and wait a couple seconds to try again.

olivere:
Hi there!  I think it'd be better if the application would be responsible for retrying. You know, applications are different: Some want to retry indefinitely, some have very hard limits and need to fail fast (or timeout).

Using `SetMaxRetries` will help. It's a very simple process actually. The retry interval doubles with every retry starting at a random value around 100ms. I've successfully avoided to make the retry interval configurable until now ;-)

If your servers fail regularly, you could try to set a lower [health check interval](https://github.com/olivere/elastic/wiki/Healthcheck) to quickly pick up new nodes as they come alive again.

But you should reconsider using an Elasticsearch server that fails regularly or is on a slow network link. What kind of use case do you have for Elasticsearch?

olivere:
Oh... I see you need to wait at startup... See [here](https://github.com/olivere/elastic/wiki/Client) and search for `SetHealthcheckTimeoutStartup`. That might help.

aanm:
Thanks a lot for the reply!
Yeah, I need to wait at startup but the `SetHealthcheckTimeoutStartup` doesn't seem to work. The problem is that elastic IP might be unreachable and on that case `SetHealthcheckTimeoutStartup` doesn't seem to have the effect I want, right?

olivere:
Hmm... you're right. The healthcheck doesn't cover that case that Elasticsearch is completely unavailable ([code](https://github.com/olivere/elastic/blob/release-branch.v2/client.go#L699)). I need to review this case.

[...]

PR #145, opened 2015-10-09
Use the provided http client to do the healthcheck on startup.

bbassingthwaite-va:
If a user provides a custom httpClient in the `NewClient`, we need to use that to do the healthcheck. In my case, I have added basic auth to the transport which is required to do a successful healthcheck.

### DIFF
--- a/client.go
+++ b/client.go
@@ -752,9 +752,8 @@ func (c *Client) startupHealthcheck(timeout time.Duration) error {
 	// If we don't get a connection after "timeout", we bail.
 	start := time.Now()
 	for {
-		cl := &http.Client{Timeout: timeout}
 		for _, url := range urls {
-			res, err := cl.Head(url)
+			res, err := c.c.Head(url)
 			if err == nil && res != nil && res.StatusCode >= 200 && res.StatusCode < 300 {
 				return nil
 			}


olivere:
You're right. Thanks for reporting. However, your PR silently [removes the `Timeout`](https://github.com/bbassingthwaite-va/elastic/commit/f53b556bfc635930c0ce34216e88529aae1893c5#diff-d0b7f86cea13f90d3cc0cc2338da8d5fL755) from the `http.Client` which we need here in this special case. I think we need to make a copy of `c.c` and then set the timeout.

Given the very common Basic Auth use case, I think it also might be worthwhile to [backport Basic Auth](https://github.com/olivere/elastic/blob/release-branch.v3/client.go#L262) from elastic.v3 into elastic.v2.

olivere:
Can you update to 2.0.13 (just `go get -u gopkg.in/olivere/elastic.v2`) and try again. It should now work both with your own `http.Transport` as well as with the backported option `elastic.SetBasicAuth` (see [the wiki](https://github.com/olivere/elastic/wiki/HttpTransport#http-basic-authentication)).

Please re-open an issue if it doesn't work for you. Thanks again.

bbassingthwaite-va:
@olivere Thanks for the quick feedback and fix.  This still doesn't work for my case since I am running this on App Engine, which requires that you use a special http.Transport that they provide.  I should have mentioned this in my original PR on top of my basic auth problem.

olivere:
Okay. I'm not familiar with Go on App Engine, sorry for that. The changes should do two things:

a) You should be able to use HTTP Basic Auth without using your own `http.Transport`.
b) It should [copy and reuse](https://github.com/olivere/elastic/blob/release-branch.v2/client.go#L770) the `http.Client`/`http.Transport` you specified via [`SetHttpClient`](https://github.com/olivere/elastic/blob/release-branch.v2/client.go#L239) in `startupHealthcheck `.

The latter should fix your original issue. Am I missing something?

bbassingthwaite-va:
Not sure why, but commenting out the `cl.Timeout = timeout` fixes my problem. I don't have it specified and it's using the default value.  Going to look into this further on why this is causing problems for me.

bbassingthwaite-va:
Not sure why but setting `elastic.SetHealthcheckTimeoutStartup(0)` but it makes it work for me.

For anyone wanting to use this library on AppEngine. This is the code that I used.

```go
type AppEngineTransport struct {
	Username string
	Password string
	Transport http.RoundTripper
}

func (t *AppEngineTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	rt := t.Transport
	if rt == nil {
		panic("transport must be supplied")
	}
	req.SetBasicAuth(t.Username, t.Password)
	return rt.RoundTrip(req)
}

transport := &AppEngineTransport{
    Username:  "username",
    Password:  "password",
    Transport: &urlfetch.Transport{Context: ctx, Deadline: time.Second * 30},
}
httpClient := &http.Client{Transport: transport}

elasticClient, err := elastic.NewClient(
    elastic.SetHttpClient(httpClient),
    elastic.SetURL(GetElasticIp(ctx)),
    elastic.SetSniff(false),
    elastic.SetHealthcheck(false),
    elastic.SetMaxRetries(1),
    elastic.SetHealthcheckTimeoutStartup(0),
)
```

We have a load balancer that sits in front of out nodes that serve the query/indexing traffic, so it's not necessary for us to do sniffing.

olivere:
Do you want me to include a flag to skip the initial healthcheck altogether? Seems a bit cleaner than a timeout of zero to me. Although I'd really like to know exactly what causes the behavior you describe...

olivere:
For reference: [`http.Client`](https://golang.org/pkg/net/http/#Client) specifies that a timeout of zero means: No timeout.

olivere:
Hmm... seems like [`urlfetch.Transport`](https://github.com/golang/appengine/blob/master/urlfetch/urlfetch.go) does not implement [`CancelRequest`](https://golang.org/pkg/net/http/#Transport.CancelRequest) which is required for `http.Timeout` to work.

bbassingthwaite-va:
Yes it looks like you are right. It would be awesome if I could skip the initial health check since we have that additional overhead of doing it on every single one of our requests.

olivere:
Why is that? Are you creating a new client for every request?

bbassingthwaite-va:
@olivere Yes, it is required since we are on AppEngine. The `urlfetch.Transport` we build is unique to every single request that we serve.

olivere:
Oh wow, that's pretty bad. Setting up a client is slow. It shouldn't matter in the normal use case because you typically only do it on startup, once. I'd love to have a better solution for App Engine then. What's the typical way to approach this on App Engine then? Can you point me to some articles?

Regarding the Timeout: Maybe it's even easier... maybe I can skip the healthcheck on startup altogether if the specified transport doesn't support CancelRequest. So you don't have to do anything.

olivere:
Also: When creating a new Client in every request, you're creating two Go routines... for every single request! That should cause some GC pressure on App Engine when your application is busy.

Maybe we could create a lightweight `NewClientForAppEngine(appengine.Context)` that doesn't do too much. Even better would be a way to keep a Client around between requests. Any chance the latter is going to fly with App Engine?

bbassingthwaite-va:
I like the idea with `NewClientForAppengine`, but the `appengine.Context` is specific to a request. To get an `appengine.Context` you would do `ctx := appengine.NewContext(request)`, Where request is a `*http.Request`. So we would have to supply the context on any requests that `elastic` wants to make. 

I think the best solution is that `NewClientForAppengine` is extremely lightweight. It spawns 0 goroutines and does no go over the wire to do health checks on initialization. This is the optimal solution for our case since it makes no sense to monitor the health of the connections for a single request. 

Although shouldn't this be possible with just disabling healthcheck and sniffing?

```go
// In NewClient 
if c.healthcheckEnabled {
   if err := c.startupHealthcheck(c.healthcheckTimeoutStartup); err != nil {
		return nil, err
	}
}
.....
if c.snifferEnabled {
  go c.sniffer()       // periodically update cluster information
}
if c.healthcheckEnabled {
  go c.healthchecker() // start goroutine periodically ping all nodes of the cluster
}

````

olivere:
I agree in that the Goroutines could be only started when sniffer and/or healthchecks are enabled. But the initial healthcheck on startup is for checking that you get a valid client. Otherwise, even if you had no Elasticsearch servers anywhere, you could create a new Client and it would only fail on the first request (see #124).

olivere:
I pushed the branch [`client-appengine`](https://github.com/olivere/elastic/tree/client-appengine) that will disable all sanity checks in `NewClient` and make creation of a `Client` cheap when you disable the sniffer and the healthchecker via `SetSniff(false)` and `SetHealthcheck(false)`. Can you check that branch and tell me if everything works smoothly for you?

bbassingthwaite-va:
I checked out the branch and everything is working :+1:

olivere:
Okay. I'm in the process of merging this into elastic.v2 and elastic.v3. It will be released in version 2.0.14 and should be available on `gopkg.in` soon.

I also added a Wiki page for [App Engine](https://github.com/olivere/elastic/wiki/App%20Engine) and pasted your sample code there. Feel free to add more details there if you have them.

PR #151, opened 2015-10-27
Add BulkSize method to get the bulk request body size in bytes

cleitonmarx:
I’m using this method to send bulk request based in the bulk body size

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -15,28 +15,30 @@ import (
 )
 
 type BulkService struct {
-	client *Client
-
+	client   *Client
 	index    string
 	_type    string
 	requests []BulkableRequest
 	//replicationType string
 	//consistencyLevel string
-	timeout string
-	refresh *bool
-	pretty  bool
+	bulkSize int
+	timeout  string
+	refresh  *bool
+	pretty   bool
 }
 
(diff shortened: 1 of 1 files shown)


olivere:
Thanks for collaborating. I'm not sure if I want this to be part of the core client. What do you use it for?

cleitonmarx:
I'm using bulk size to know the bulk request body size and control when I want to send the request. It's an alternative approach to NumbersOfActions function. 
Regards, 
Cleiton 
On Oct 28, 2015 5:54 AM, Oliver Eilhard <notifications@github.com> wrote:Thanks for collaborating. I'm not sure if I want this to be part of the core client. What do you use it for?

olivere:
I understand. Hmm... I'm trying to keep elastic aligned to the Java API. The Java API doesn't have a concept like this. I see how this can be useful in certain use cases, but I think I refrain from adding it to core. Do you see a way of adding this to your application without modifying elastic?

Furthermore, the name `bulkSize` (to represent the request body size) clashes with the concept of "bulk size" used in the clients to represent the number of "requests to send in one batch"; perhaps using `bodySize` or `requestBodySize` would be a better fit?

PR #152, opened 2015-10-29
Add method to set context queries for a suggest field

DanielHeckrath:
As shown in the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/suggester-context.html) it's possible to set a context for suggest fields. 
It should be possible to set this context when indexing a new document


olivere:
Thanks. I also added a [test for the `SuggestField`](https://github.com/olivere/elastic/blob/release-branch.v3/suggest_field_test.go).

Will backport into v2. It's missing there as well.

PR #153, opened 2015-10-30
parse update result when the body contain error info and status

foreversmart:
I found when i update an index not exist. the elastic server will get this error 
`{
  "error": "DocumentMissingException[[xxxxx][1] [room][56334545f6576e24c6000001_56334545f6576e24c6000005]: document missing]",
  "status": 404
}`

but i can't find the define in the UpdateResult, so i can't get the all info of the result.

so I add these two fields in the UpdateResult struct Error for get error info, Status for get Status info.

### DIFF
--- a/update.go
+++ b/update.go
@@ -20,6 +20,8 @@ type UpdateResult struct {
 	Id        string     `json:"_id"`
 	Version   int        `json:"_version"`
 	Created   bool       `json:"created"`
+	Error     string     `json:"error"`
+	Status    int        `json:"status"`
 	GetResult *GetResult `json:"get"`
 }
 


olivere:
Are you using v2 or v3? If v3, isn't this covered by the behavior described in the [CHANGELOG for v3](https://github.com/olivere/elastic/blob/release-branch.v3/CHANGELOG-3.0.md#http-status-404-not-found)?

foreversmart:
i'm using v2

by the way can i use v3 with elasticsearch1.7?

olivere:
Okay. No. v2 is for ES 1.x, v3 is for ES 2.x.

Let me review your change. I'm a bit hesitant as 404 is not an error in v2, so I'd probably need to add it to all structs that could potentially result in a 404. That's obviously not good so I'm trying to find a better solution.

foreversmart:
in this case the 404 status code is not return in the elastic error but int the elastic result. and i look the v3 code again but did't find any JSON filed in the UpdateResult can parse these error info.

olivere:
In v3, if the server responds with 404, you get an error. See [checkResponse in errors.go](https://github.com/olivere/elastic/blob/release-branch.v3/errors.go#L29), which is called with each response from ES.

foreversmart:
yes, you are right~, but how should i do in v2? thanks a lot.

olivere:
Also, the error you get is not part of an `UpdateResponse` but a general ES error as [`UpdateResponse` does not have an `error` field](https://github.com/elastic/elasticsearch/blob/1.7/src/main/java/org/elasticsearch/action/update/UpdateResponse.java#L31).

I think you should make sure that the document/index exists in the first place.

olivere:
I think we have two different issues at play here:

1. If you do an `Update` an get a 404, elastic.v2 should give you an error. Which it doesn't right now, so you don't have a chance to find out if the update failed. I'm thinking of a solution for that case. Adding two fields to `UpdateResponse` is not a solution but a workaround.
2. If you want the `Update` to do an insert if the document does not exist, you should be able to do [Upserts](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/docs-update.html#upserts), probably with [`doc_as_upsert`](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/docs-update.html#_literal_doc_as_upsert_literal) in your case (here is a [test for that in elastic.v2](https://github.com/olivere/elastic/blob/release-branch.v2/update_test.go#L202)).

foreversmart:
yes, i get the two way, but the second way is two slow for me, because the upsert data you should prepared is not needed in most time and the prepare time is too much. so maybe the best solution is update a index and get the failed info and upsert it.

olivere:
Can you test the [pr-153](https://github.com/olivere/elastic/tree/pr-153) branch and see if that change correctly solves your issue. The commit changes the outcome of a failed `Update`. When a 404 is returned, you get an error which contains the  error message and details (see [here](https://github.com/olivere/elastic/blob/pr-153/errors_test.go#L34) how to extract that information).

foreversmart:
I check your code again and again, sadly find this code
`if (res.StatusCode >= 200 && res.StatusCode <= 299) || res.StatusCode == http.StatusNotFound {
 		return nil
}`
in v2

and
`if req.Method == "HEAD" && res.StatusCode == 404 {
		return nil
}`

in v3 

so the 404 not found error return nil directly!

olivere:
Yes. As I said, 404 is **not** an error in v2 but **is** an error in v3 (unless it is a `HEAD` request).

foreversmart:
i know your point。so the best idea is pack these error info in Response？

olivere:
Did you test the [pr-153](https://github.com/olivere/elastic/tree/pr-153) branch already? See [this commit](https://github.com/olivere/elastic/commit/bb3d84dfc6becd02a945ce9349ce351aae5609fd).

foreversmart:
I test it, it's just what i want. thanks very much~. 
by the way will you update this on gopkg.in?

olivere:
On the road now. Will update today, but it'll take 3-4 hours probably...

olivere:
Just pushed 2.0.16. Use `go get -u gopkg.in/olivere/elastic.v2` to update. Thanks for reporting the issue!

foreversmart:
thank you very much to help me solve this problem

PR #154, opened 2015-10-30
Support multiple indices in the Exists API.

tomarus:
Hi I've added multiple index support to the exists-api. Please review this and if you could add it that would be nice.


olivere:
The Exists API does not support multiple indices.

Issue #170, opened 2015-11-25
Not all requests use POST instead of GET


gdm85:
I just noticed #167, nice feature addition, however I am not sure this is complete.

I noticed that calls for settings, aliases for example are not using it:
```
GET http://localhost:9200/_all/_settings [status:200, request:0.033s]
elastic: http://localhost:9200 is dead
time="2015-11-25T11:32:47+01:00" level=fatal msg="Get http://localhost:9200/[>4096 bytes omissis]/_aliases: EOF"
```

This in turn makes the `elastic.SetSendGetBodyAs("POST")` less effective.

olivere:
Can you explain exactly what you do to retrieve settings/aliases (so I can try to reproduce where this happens)?

gdm85:
Pretty sure it's a call to `client.Aliases()`, however I can report this back in about 40 hours

olivere:
Didn't have time to look into it. But all calls with GET and a body should be changed to POST in Client.PerformRequest. However, GET calls *without* a body will still use GET!

gdm85:
@olivere you're right, I am afraid this is a type of query that doesn't benefit from using GET with body.

The only way one could workaround the issue here is making multiple calls to make sure the URI stays below 4096 bytes (the default for ES' natty), then merge the results.

Would you likely accept a PR adding a feature like this? I would start specifically from this call and check if others need it to.

Pros:
* if you have load balancers/proxy you want to stay below a specific (standard-friendly) amount of bytes for your GET requests URIs
Cons:
* for loops in the code when request URI is too long, makes code a tad less elegant probably

olivere:
I'm afraid I probably don't understand the root issue yet. Can you elaborate?

From the perspective of a client, I'd like to stay within the limits of the official clients. If you e.g. find a behavior in the official clients for Ruby, Python, Perl etc. and elastic behaves different, I'm very likely willing to change that. But I'm probably not moving to far off track. E.g. elastic should always be "just a client", not some sort of "do-it-all" library. I hope the explanation will give you an idea :-)

olivere:
@gdm85 It looks like you have problems with maximum length of the URL, do you? Can you can increase [`http.max_initial_line_length`](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-http.html) in your cluster?

gdm85:
[...]

PR #177, opened 2015-11-30
Fix tracelog issue with large requests/responses

gdm85:
There is a limitation in bufio that prevents current trace dumps to work:
```
time="2015-11-30T14:50:01+01:00" level=info msg="ES trace 2015/11/30 14:50:01 GET /_all/_settings HTTP/1.1" 
time="2015-11-30T14:50:01+01:00" level=info msg="Host: x.x.x.x:9200" 
time="2015-11-30T14:50:01+01:00" level=info msg="User-Agent: elastic/2.0.20 (linux-amd64)" 
time="2015-11-30T14:50:01+01:00" level=info msg="Accept: application/json" 
time="2015-11-30T14:50:01+01:00" level=info msg="Accept-Encoding: gzip" 
time="2015-11-30T14:50:01+01:00" level=info msg= 
time="2015-11-30T14:50:01+01:00" level=info msg= 
time="2015-11-30T14:50:01+01:00" level=info msg="ES trace 2015/11/30 14:50:01 HTTP/1.1 20
0 OK" 
time="2015-11-30T14:50:01+01:00" level=info msg="Content-Length: 191369" 
time="2015-11-30T14:50:01+01:00" level=info msg="Content-Type: application/json; charset=UTF-8" 
time="2015-11-30T14:50:01+01:00" level=info msg= 
time="2015-11-30T14:50:01+01:00" level=error msg="Error while reading from Writer: bufio.Scanner: token too long" 
time="2015-11-30T14:50:01+01:00" level=info msg="ES info 2015/11/30 14:50:01 GET http://x.x.x.x:9200/_all/_settings [status:200, request:0.065s]" 
time="2015-11-30T14:50:01+01:00" level=fatal msg="elastic: Error 400 (Bad Request)" component="cmd/es-index-deploy" hostname=xxx
```

You can find the constant precisely at line 71 here: https://golang.org/src/bufio/scan.go (being 64K)

When the user is troubleshooting issue #170 he/she can't get any help from log traces because of this problem. By using the trace method I introduce here, will instead be able to see the large request and provide/use it as a testcase.


olivere:
@gdm85 TBH I'm not sure I want to enable this by default. Don't misunderstand me: I know what you're after, and I see how it can be valuable. However, it can also result in GB of log file data without even noticing. Maybe by introducing a verbosity level? I personally think elastic is already too verbose by default.

gdm85:
@olivere as it is now, it has an (undocumented) limitation/feature that goes more or less like this:
* bufio has a 64K scan limitation, so anything larger than 64K will result in an error being printed instead of the actual HTTP requests/responses

You could either document this in a KNOWN_BUGS file, or allow this to be configured, or leave to the user the problem of the GBs in the logs.

I do not think it is the job of the client library to take care of verbose output when you enable such verbose output, but rather if the client library can help in "seeing" what goes on behind curtains, that shall be the biggest priority as without it you simply are stuck, and need to use tcpdump/wireshark instead.

olivere:
@gdm85 Can you write a test that illustrates the issue in elastic and how your PR changes that?

gdm85:
@olivere sure, I will send it in a few hours

gdm85:
I am having some issues at reproducing the problem; I am progressively increasing the generated response size until I can reproduce it.

olivere:
Same here. I tried to reproduce it by writing to `bytes.Buffer` but it logged just fine at any arbitrary size. So I asked to see if you could find a reproducible test.

olivere:
@gdm85 Sorry to bother you, but I'm cleaning up issues. Did you have success in reproducing the issue recently?

gdm85:
No you're right, sorry for not coming back on this.
I will open it again if problem is found and there is a testcase.

PR #179, opened 2015-12-02
Add Term Vectors API for Elasticsearch 1.x (elastic.v2)

t2y:
I found elastic doesn't support Term Vector API now. So, I have made the mappings to call the *termvector* API. This is my first contribution to elastic. It might not be satisfied with contributing guidelines. If so, tell me how to do completely and I'll fix it.

Term Vector API has improved Elasticsearch 2.x. For example, it renamed _termvector_ to _termvector**s**_ . This PR is for Elasticsearch 1.x.

* https://www.elastic.co/guide/en/elasticsearch/reference/1.7/docs-termvectors.html

### DIFF
--- a/client.go
+++ b/client.go
@@ -1381,3 +1381,11 @@ func (c *Client) WaitForGreenStatus(timeout string) error {
 func (c *Client) WaitForYellowStatus(timeout string) error {
 	return c.WaitForStatus("yellow", timeout)
 }
+
+// TermVector returns information and statistics on terms in the fields
+// of a particular document.
+func (c *Client) TermVector(index, typ, id string) *TermvectorService {
+	builder := NewTermvectorService(c)
+	builder = builder.Index(index).Type(typ).Id(id)
+	return builder
+}
--- a/client.go
+++ b/client.go
@@ -1384,8 +1384,8 @@ func (c *Client) WaitForYellowStatus(timeout string) error {
 
 // TermVector returns information and statistics on terms in the fields
 // of a particular document.
-func (c *Client) TermVector(index, typ, id string) *TermvectorService {
+func (c *Client) TermVector(index, typ string) *TermvectorService {
 	builder := NewTermvectorService(c)
-	builder = builder.Index(index).Type(typ).Id(id)
+	builder = builder.Index(index).Type(typ)
 	return builder
 }
--- a/termvector.go
+++ b/termvector.go
@@ -44,91 +44,91 @@ func NewTermvectorService(client *Client) *TermvectorService {
 	}
 }
 
-// Index is documented as: The index in which the document resides..
+// Index is documented as: The index in which the document resides.
 func (s *TermvectorService) Index(index string) *TermvectorService {
 	s.index = index
 	return s
 }
 
-// Type is documented as: The type of the document..
+// Type is documented as: The type of the document.
 func (s *TermvectorService) Type(typ string) *TermvectorService {
 	s.typ = typ
 	return s
 }
 
-// Id is documented as: The id of the document..
+// Id is documented as: The id of the document.
 func (s *TermvectorService) Id(id string) *TermvectorService {
 	s.id = id
 	return s
 }
 
-// Doc is documented as: The document to analyze..
+// Doc is documented as: The document to analyze.
 func (s *TermvectorService) Doc(doc map[string]string) *TermvectorService {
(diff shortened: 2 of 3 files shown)


olivere:
@t2y Thanks a lot. This is looking really good.

Just a few words regarding the implementation of services. First, I try to use the Java client as the root source rather than the documentation. The official documentation often lacks details where the Java source even shows off the generated JSON structure for every edge case. I also use the Java source for creating the responses. E.g. here's the code for [TermVectorRequestBuilder](https://github.com/elastic/elasticsearch/blob/1.7/src/main/java/org/elasticsearch/action/termvector/TermVectorRequestBuilder.java) and [TermVectorResponse](https://github.com/elastic/elasticsearch/blob/1.7/src/main/java/org/elasticsearch/action/termvector/TermVectorResponse.java#L177) for 1.7. Maybe you're missing some information in the response (didn't look it up).

Second, did you see that the [generate-api branch](https://github.com/olivere/elastic/tree/generate-api/generator) has a Go generator for services. It takes the [REST specification](https://github.com/elastic/elasticsearch/tree/master/rest-api-spec) and generates the Go services. Well, it doesn't do it right 100% of the time (it doesn't get the GET/POST/PUT right sometimes), but at least it pumps out all properties available according to the REST specification. So if there's a service missing, I try to use the Go generator to create a 90% template for me.

Again, thanks for the PR. I will try to merge soon.

t2y:
Thank you for detailed reply.

> I also use the Java source for creating the responses. (snip) Maybe you're missing some information in the response (didn't look it up).

Oh, that's a good method. I'll check and update seeing the Java source.

> did you see that the generate-api branch has a Go generator for services. It takes the REST specification and generates the Go services.

Yes. I originally made the source file from the generator, then I updated a few line and the response object mappings. That's a notable tool. I learnt Go's potential using it.

t2y:
@olivere 

I implemented some fields in the response object and some parameters for API referring Java source. Is there other tasks?

olivere:
@t2y Seems okay. Only thing I'd change is to use `map[string]string` instead of `*map[string]string` and then test for `len(doc) > 0` instead of `doc != nil`.

t2y:
@olivere Thank you for reviewing. I fixed it.

t2y:
> Can you also fix the documentation here as well as in the other places? It should read as an English sentence.

@olivere Which document should I fix? I'm not good at English writing, but I'll try if it's boilerplate task.

t2y:
> Can you please also add a test for buildURL and the two endpoints (see other tests)?

Thanks. I'll write a test referring other test cases.

olivere:
@t2y Open [this link](https://github.com/t2y/elastic/commit/fd75d19d896a70f0f083eb767c7a0946d601beaf). My comments are inlined so you see the exact location where I added notes.

Re: documentation. It's just a formatting issue, e.g. instead of `Index is documented as: The index in which the document resides..` it should read as `Index in which the document resides.`

t2y:
@olivere I got this error on Travis CI, but this error doesn't happened on my local environment. Do you know how to fix?

```
--- FAIL: TestTermVectorWithDoc (0.07s)
termvector_test.go:70: elastic: Error 503 (Service Unavailable):
  NoShardAvailableActionException[[elastic-test][0] null]; nested:
  IllegalIndexShardStateException[[elastic-test][0] CurrentState[POST_RECOVERY]
  operations only allowed when started/relocated]; 
```

olivere:
@t2y Travis is wonky sometimes. Often a [small sleep interval helps](https://github.com/olivere/elastic/blob/4572b761a1de1341ce47b2cf591f4b9ac1e2217a/update_test.go#L320), sometimes manually refreshing or flushing after indexing data. Sometimes you can even restart the test and it works on the 2nd run. Don't worry about ES errors too much. Non-ES errors are more important. I will run tests again when merging the PR. If I can't get it to run reliably on Travis, I'm even considering to [disable it on CI](https://github.com/olivere/elastic/blob/2219d4c999f6dba3d68c280e35c5523c731e55c2/scan_test.go#L382).

t2y:
@olivere I see. Now, I don't fix the test code. Please add sleep interval or disable it on CI when you think so.

olivere:
Merged and should be available in 2.0.23. Thanks for your support.

t2y:
Thanks a lot!

PR #182, opened 2015-12-04
Add services to get, put and delete warmers

DanielHeckrath:
This change adds the missing services for get, put and delete requests against the warmer api.

I'm not a 100% sure about the implementation of the get url. The response differs slightly if there is a type present in the url. I've used * as the fallback if no type is set for the request, which results in the second case in the following example. I think, while the response we get back is different, that we don't lose any important information.

Example: we have 2 indices (index_1, index_2) and a warmer (warmer_1) in index_1.

If we use the url `/_all/_warmer` there response would be
```
{
  "index_1": {
    "warmers": {
      "warmer_1": {
        "types": [
          "type_1"
        ],
        "source": {
          "query": {
            "match_all": {}
          }
        }
      }
    }
  },
  "index_2": {
    "warmers": {}
  }
}
```

We could also get the warmers with the following url `/_all/*/_warmer/_all`

```
{
  "index_1": {
    "warmers": {
      "warmer_1": {
        "types": [
          "type_1"
        ],
        "source": {
          "query": {
            "match_all": {}
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -1287,6 +1287,21 @@ func (c *Client) PutMapping() *PutMappingService {
 	return NewPutMappingService(c)
 }
 
+// GetWarmer gets a mapping.
+func (c *Client) GetWarmer() *GetWarmerService {
+	return NewGetWarmerService(c)
+}
+
+// PutWarmer registers a warmer.
+func (c *Client) PutWarmer() *PutWarmerService {
+	return NewPutWarmerService(c)
+}
+
+// DeleteWarmer deletes a warmer.
+func (c *Client) DeleteWarmer() *DeleteWarmerService {
+	return NewDeleteWarmerService(c)
+}
+
 // -- cat APIs --
 
 // TODO cat aliases
--- a/indices_delete_warmer.go
+++ b/indices_delete_warmer.go
@@ -1,0 +1,142 @@
+// Copyright 2012-2015 Oliver Eilhard. All rights reserved.
+// Use of this source code is governed by a MIT-license.
+// See http://olivere.mit-license.org/license.txt for details.
+
+package elastic
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http/httputil"
+	"net/url"
+	"strings"
+
+	"github.com/olivere/elastic/uritemplates"
+)
+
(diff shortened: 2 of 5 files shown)


olivere:
@DanielHeckrath Thanks for your contribution. I will give you feedback as soon as I find some time.

Just a few words regarding the implementation of services and my general approach. First, I try to use the [Java client](https://github.com/elastic/elasticsearch) as the root source rather than the documentation. The official documentation often lacks details where the Java source even shows off the generated JSON structure for every edge case. I also use the Java source for creating the responses. E.g. here's the code for [get/put/delete warmers](https://github.com/elastic/elasticsearch/tree/2.1/core/src/main/java/org/elasticsearch/action/admin/indices/warmer), the [GetWarmersRequestBuilder](https://github.com/elastic/elasticsearch/blob/2.1/core/src/main/java/org/elasticsearch/action/admin/indices/warmer/get/GetWarmersRequestBuilder.java) and [GetWarmersResponse](https://github.com/elastic/elasticsearch/blob/2.1/core/src/main/java/org/elasticsearch/action/admin/indices/warmer/get/GetWarmersResponse.java) (all for 2.1).

Second, you might have noticed the [generate-api branch](https://github.com/olivere/elastic/tree/generate-api/generator). It contains a Go generator for services that takes the [(version-specific) REST specification](https://github.com/elastic/elasticsearch/tree/master/rest-api-spec) of the ES endpoints and generates the Go services. It doesn't do it right 100% of the time (it doesn't get the GET/POST/PUT right sometimes), but at least it pumps out all properties available according to the REST specification. So if there's a service missing, I try to use the Go generator to create a template that is 90% complete. E.g. here's the official REST spec for [indices.get_warmer (for ES 2.1)](https://github.com/elastic/elasticsearch/blob/2.1/rest-api-spec/src/main/resources/rest-api-spec/api/indices.get_warmer.json).

olivere:
@DanielHeckrath I just pushed the Warmer API for elastic.v3. I changed the `buildURL` a bit and added some more tests. Getting the URLs right is hard sometimes.

Thank you very much for your contribution.

PR #193, opened 2016-01-04
fix sniffing for ES managed by AWS

SeUkKim:
ES managed by AWS does not support sniffing = true by default.


SeUkKim:
it causes "no Elasticsearch node available" error when you using ES managed by AWS.

SeUkKim:
i find SetSniff func.

PR #207, opened 2016-01-22
Scan cursor mutex

avleen:
We wanted to get some performance gains around scan and scroll, and needed to add a couple of mutexes to make `ScanCursor.Next()` be more thread safe.
The change doesn't impact performance when only one thread is calling `Next()`, but they helped get us a ~40% performance improvement by letting multiple threads call `Next()` at the same time.


avleen:
Ah nuts. sorry, I think I have a bad merge here?

PR #208, opened 2016-01-22
Add a couple of mutexes in Next(), to make this more thread safe.  Th…

avleen:
We wanted to get some performance gains around scan and scroll, and needed to add a couple of mutexes to make `ScanCursor.Next()` be more thread safe.
The change doesn't impact performance when only one thread is calling `Next()`, but they helped get us a ~40% performance improvement by letting multiple threads call `Next()` at the same time.

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/url"
 	"strings"
+	"sync"
 
 	"github.com/olivere/elastic/uritemplates"
 )
@@ -285,20 +286,24 @@ func (s *ScanService) Do() (*ScanCursor, error) {
 type ScanCursor struct {
 	Results *SearchResult
 
-	client      *Client
-	keepAlive   string
-	pretty      bool
-	currentPage int
+	client       *Client
+	keepAlive    string
+	pretty       bool
+	currentPage  int
+	fetch_mutex  *sync.Mutex
+	return_mutex *sync.Mutex
 }
 
 // newScanCursor returns a new initialized instance
 // of scanCursor.
 func NewScanCursor(client *Client, keepAlive string, pretty bool, searchResult *SearchResult) *ScanCursor {
 	return &ScanCursor{
-		client:    client,
-		keepAlive: keepAlive,
-		pretty:    pretty,
-		Results:   searchResult,
+		client:       client,
+		keepAlive:    keepAlive,
+		pretty:       pretty,
+		Results:      searchResult,
+		fetch_mutex:  &sync.Mutex{},
+		return_mutex: &sync.Mutex{},
 	}
 }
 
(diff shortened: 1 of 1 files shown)


avleen:
Build failed, possibly due to a travis-ci error. Closing and re-opening to trigger another build.

olivere:
I will have a look. I had a new "Iterator"-style pattern on the roadmap to merge Scan and Scroll but it didn't make it into 3.0. Maybe I should pick it up again...

Can you provide some insight into what you do with scan-and-scroll? It seems like you're starting up goroutines to concurrently pull from the "cursor". What I would've done is to start one goroutine for pulling and many others for processing, typically via a channel. My gut feeling is that I/O is the problem here, not the workers processing data. But I haven't tested that.

What got you to a 40% performance improvement?

avleen:
Yup!
We started with exactly what you suggest: One thread constantly pulling from Elasticsearch, and sticking the data on a channel. Elasticsearch was able to send data back faster than we could process it - all our servers have 10gbit networking, all the data is in memory, etc. So I/O is never close to being a limitation fortunately.

The result was one CPU core being 100% busy pulling data from elasticsearch, unmarshalling the JSON, and then finally putting it on a channel for more processing.

Next we tried to have multiple threads pulling data from the same cursor at the same time.
This was better, but introduced two race conditions until we added these mutexes. 
Now we get 4-5 CPU cores which are all more than 50% busy at the same time.

The entire cost is in processing the JSON :-)

olivere:
@avleen Interesting. I didn't expect that. Thanks for sharing.

Okay, I'll review your changes and make `ScanService` safe from races. I'm a bit busy so it could take a few days.

avleen:
Not a problem - very much appreciate the work you and others have done on this.
If there's anything I can do to help, let me know!

avleen:
One other place we found the JSON processing to be expensive and had to be forced in-line in a single thread, was in the decoder for `PerformRequest()`, but I couldn't think of a good way to split that out and make it thread safe. Fortunately the mutex around `PerformRequest()` was able to hide that for now.

That one seems like a ton more work to me (I'm just pretending to be a software engineer), so I didn't tackle it as the gains I was able to make were good enough for now. There be dragons there :-)

olivere:
@avleen Just looked into this issue. I see no way of making this safe from data races without a breaking change. The main issue is that the `ScanCursor` has a public field `Results` that cannot be protected. Your implementation also has some issues. E.g. the `fetch_mutex` could be in a locked state when you're leaving [here](https://github.com/avleen/elastic/blob/20253f16aa496bb8a9afd1b81b70d958895ff44c/scan.go#L369).

I guess I have to pick up that Iterator idea again, and make that one safe for usage from different goroutines.

olivere:
BTW: `ScrollService` has been refactored in 3.0.48. It should now be safe to use from multiple goroutines. See [this example in the Wiki](https://github.com/olivere/elastic/wiki/ScrollParallel) to illustrate scrolling in parallel. Maybe you can give it a try?

avleen:
Thanks Oliver! I'll be able to refactor my stuff to try in a few weeks.

Issue #211, opened 2016-02-01
release-branch.v3 elastic.RangeQuery syntax broken on ES 2.x


adamalix:
Currently, [release-branch.v3](https://github.com/olivere/elastic/tree/release-branch.v3) uses the old range query syntax.  See the new [Range Query DSL Syntax here](https://www.elastic.co/guide/en/elasticsearch/reference/2.x/query-dsl-range-query.html#query-dsl-range-query).

We should replace:
- `RangeQuery.From()` with `RangeQuery.Gte()`
- `RangeQuery.To()` with `RangeQuery.Lte()`

Remove:
- `RangeQuery.IncludeLower()`
- `RangeQuery.IncludeUpper()`

Add:
- `RangeQuery.Lt()`
- `RangeQuery.Gt()`

olivere:
@adamalix Thanks a lot for your support. I'm rather busy at the moment, so it may take a few days.

I try not to remove existing code though (unless it's really unsupported) as that might break clients.

olivere:
@adamalix RangeQuery should resemble the API of the Java client (see [RangeQuery](https://github.com/elastic/elasticsearch/blob/2.x/core/src/main/java/org/elasticsearch/index/query/RangeQueryBuilder.java)).

Elastic documentation is good. I try to take the Java API as reference though :-)

adamalix:
@olivere cool. will fix up PR to reflect.

olivere:
@adamalix Thanks a lot. I promise I'll look into it as soon as I find some time.

olivere:
See discussion on [PR-214](https://github.com/olivere/elastic/pull/214).

Issue #196, opened 2016-01-09
range query instead of gte, it uses include_lower, upper, to


sundarv85:
Hello,

When I use the code
```
	dateFilter := NewBoolQuery()
	dateFilter = dateFilter.Filter(NewRangeQuery("ts").Gte("now-24h/d"))
	dateFilter = dateFilter.Filter(NewRangeQuery("ts").Gte("now-24h"))
```

I was only expecting the `gte` keyword to be used. But I see
```
                         "from": "now-24h/d",
                        "include_lower": true,
                        "include_upper": true,
                        "to": null 
```

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html also explains only about `gte`. Is there a reason why you convert to these values. Also, I was trying to find in ES Reference of range query. Is this mentioned somewhere in docs. 

Full Query below:

```
{
  "bool": {
    "_name": "Test",
    "filter": {
      "bool": {
        "filter": [
          {
            "term": {
              "singer": "Michael"
            }
          },
          {
            "term": {
              "kind": "rock"
            }
          }
        ]
      }
    },
    "must": {
      "has_child": {
        "inner_hits": {
          "fields": "ts",
          "name": "ts",
          "size": 1,
          "sort": [
            {
              "ts": {
                "order": "desc"
              }
            }
          ]
        },
        "query": {
          "bool": {
            "filter": {
              "bool": {
                "filter": [
                  {
                    "range": {
                      "ts": {
                        "from": "now-24h/d",
                        "include_lower": true,
                        "include_upper": true,
                        "to": null
                      }
                    }
                  },
                  {
                    "range": {
                      "ts": {
                        "from": "now-24h",
                        "include_lower": true,
                        "include_upper": true,
                        "to": null
                      }
                    }
                  }
                ]
              }
            },
            "must": {
              "match_all": {}
            }
          }
        },
        "score_type": "sum",
        "type": "music"
      }
    }
  }
}
```

olivere:
AFAICS the behavior is semantically correct, isn't it?

[...]

PR #214, opened 2016-02-02
updates range query syntax for elastic 2.x

adamalix:
Elasticsearch documentation omits the `from` and `to` fields on
RangeQuery. This change brings the produced queries in line with the
documentation and proxies old methods through the `Gte` and `Lte`
methods.

https://www.elastic.co/guide/en/elasticsearch/reference/2.x/query-dsl-range-query.html

- Proxy `From(interface{})` method on elastic.RangeQuery through
  `Gte(interface{})`.
- Proxy `To(interface{})` method on elastic.RangeQuery through
  `Lte(interface{})`.
- Adds `Lt(interface{})` method on elastic.RangeQuery. Has the same
  effect of `RangeQuery{}.To().IncludeUpper(false)`.
- Adds `Gt(interface{})` method on elastic.RangeQuery. Has the same
  effect of `RangeQuery{}.From().IncludeLower(false)`.
- Change internal behavior of `IncludeLower(bool)` method on
  elastic.RangeQuery.
- Change internal behavior of `IncludeLower(bool)` method on
  elastic.RangeQuery.
- New tests for updated bounds checking behavior.
- Fixes tests in:
  * search_aggs_bucket_filter_test.go
  * search_aggs_bucket_filters_test.go
  * search_queries_bool_test.go
  * search_queries_boosting_test.go
  * search_queries_nested_test.go
  * search_queries_not_test.go
  * search_queries_range_test.go

Addresses #211

### DIFF
--- a/search_queries_range.go
+++ b/search_queries_range.go
@@ -9,79 +9,84 @@ package elastic
 // For details, see
 // https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html
 type RangeQuery struct {
-	name         string
-	from         interface{}
-	to           interface{}
-	timeZone     string
-	includeLower bool
-	includeUpper bool
-	boost        *float64
-	queryName    string
-	format       string
+	name      string
+	gt        interface{}
+	gte       interface{}
+	lt        interface{}
+	lte       interface{}
+	timeZone  string
+	boost     *float64
+	queryName string
+	format    string
 }
 
 // NewRangeQuery creates and initializes a new RangeQuery.
 func NewRangeQuery(name string) *RangeQuery {
-	return &RangeQuery{name: name, includeLower: true, includeUpper: true}
+	return &RangeQuery{name: name}
 }
 
-// From indicates the from part of the RangeQuery.
+// From indicates the `gte` part of the RangeQuery.
 // Use nil to indicate an unbounded from part.
 func (q *RangeQuery) From(from interface{}) *RangeQuery {
-	q.from = from
-	return q
+	return q.Gte(from)
 }
 
-// Gt indicates a greater-than value for the from part.
-// Use nil to indicate an unbounded from part.
-func (q *RangeQuery) Gt(from interface{}) *RangeQuery {
-	q.from = from
-	q.includeLower = false
-	return q
-}
-
-// Gte indicates a greater-than-or-equal value for the from part.
-// Use nil to indicate an unbounded from part.
-func (q *RangeQuery) Gte(from interface{}) *RangeQuery {
-	q.from = from
-	q.includeLower = true
-	return q
-}
-
-// To indicates the to part of the RangeQuery.
-// Use nil to indicate an unbounded to part.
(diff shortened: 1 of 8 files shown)


search_queries_range.go
```diff
@@ -9,79 +9,84 @@ package elastic
...
+		q.lte = nil
+	}
 	return q
 }
 
 // IncludeUpper indicates whether the upper bound should be included or not.
 // Defaults to true.
 func (q *RangeQuery) IncludeUpper(includeUpper bool) *RangeQuery {
-	q.includeUpper = includeUpper
+	if includeUpper {
```
adamalix: it's unclear if this method (and [IncludeLower](https://github.com/olivere/elastic/pull/214/files#diff-e0951bfd14ee4008141d6c8a228a4178R70) should now do nothing or if it should mutate the query.

olivere:
@adamalix I looked at it again. But quite frankly: I can't figure out what you're trying to achieve. 

AFAICS the code in [search_queries_range.go](https://github.com/olivere/elastic/blob/release-branch.v3/search_queries_range.go) matches the implementation of [RangeQueryBuilder.java](https://github.com/elastic/elasticsearch/blob/2.x/core/src/main/java/org/elasticsearch/index/query/RangeQueryBuilder.java) in the 2.x branch of Elasticsearch. The only change your PR provides is a different presentation of ranges on the wire but comes at the cost of losing compatibility with the Java API. I think the latter is far more important than having a nice representation on the wire. Don't you agree?

But maybe I'm missing something... Can you explain what issues your PR addresses?

adamalix:
@olivere I just wanted to match the wire format to the elasticsearch docs.  If the format used by elastic matches the official Java client wire format, then it's probably safe to leave it as such.

adamalix:
More explanation: I got confused debugging some issues I was having with RangeQuery and wanted to make sure that I was sending data correctly, so I wrote changes to match the ES documentation.

sundarv85:
@olivere  @adamalix  I was also bitten by this. The ES documentation seems to be different from Java/elastic API though in the end they are both the same. I had logged a similar issue #196

olivere:
@adamalix Okay, I get your point. Unfortunately the official Elasticsearch documentation is more of a high-level introduction into the REST API. There's a lot more going on under the covers which you can only see if you take a deep dive into the source code. Don't get me wrong: The documentation is A+ but it's incomplete and ambiguous nonetheless. As Linus said: "Talk is cheap. Show me the code." ;-)

Changing the "on-the-wire" format also comes at a significant cost. By changing that, I make debugging easier. True. However, I'm no longer in line with the Java API. Some dark edges in Elastic are already out-of-sync and I wish I haven't done that. So, I'm sorry, I think I will keep the API as is.

Still: [Thank you for your contribution](https://github.com/olivere/elastic/commit/f93749a7dd6645ed8d0a89ad83d517a80f001540). You're contributions are always welcome.

adamalix:
@olivere totally understood. thanks for explaining your reasoning!

PR #215, opened 2016-02-02
missing value for terms aggregation

shalinlk:
Missing value in terms aggregation has been added. Count of docs which misses the aggregating field will be counted under the given name. 

ES Feature : https://www.elastic.co/guide/en/elasticsearch/reference/2.x/search-aggregations-bucket-terms-aggregation.html#_missing_value_12

### DIFF
--- a/search_aggs_bucket_terms.go
+++ b/search_aggs_bucket_terms.go
@@ -30,6 +30,7 @@ type TermsAggregation struct {
 	showTermDocCountError *bool
 	includeTerms          []string
 	excludeTerms          []string
+	missingTerm           string
 }
 
 func NewTermsAggregation() *TermsAggregation {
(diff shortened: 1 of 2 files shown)


olivere:
@shalinlk I found that it's not that simple. All aggregations derived from [ValuesSourceAggregationBuilder](https://github.com/elastic/elasticsearch/blob/2.x/core/src/main/java/org/elasticsearch/search/aggregations/ValuesSourceAggregationBuilder.java#L44) in Java have that "missing field" property, and it can not only be a string but any type. So I had to fix all of them: Histogram, DateHistogram, Range, Terms etc.

[This commit](https://github.com/olivere/elastic/commit/aaba83e3fd9849159da1d646d0b8a5be1182a8b8) has all necessary changes (and adds the [Sampler Aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-sampler-aggregation.html) as well). I will release a new version to `gopkg.in` today.

Thanks for your support.

shalinlk:
Thanks for the fix and sorry for missing others. :-(

PR #216, opened 2016-02-03
Return an error if setting the request body fails

idubinskiy:
I passed a type with a custom `MarshalJSON` to `BodyJson` and was getting the error `elastic: Error 400 (Bad Request): failed to parse, document is empty [type=mapper_parsing_exception]`. It took me a little while to debug and figure out that the custom `MarshalJSON`, and thus the `json.Marshal` in `setBodyJson`, was returning an error.

Since `setBodyJson` and `setBody` already return the error, this propagates it up to `PerformRequest`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -965,7 +965,11 @@ func (c *Client) PerformRequest(method, path string, params url.Values, body int
 
 		// Set body
 		if body != nil {
-			req.SetBody(body, gzipEnabled)
+			err = req.SetBody(body, gzipEnabled)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


olivere:
@idubinskiy Oh, that makes sense. Sorry for missing that, and thanks for your contribution.

olivere:
@idubinskiy Thanks for the PR. I merged it into v2 and v3. Will release a new version on `gopkg.in` today.

idubinskiy:
Thanks!

PR #218, opened 2016-02-04
Use an interface for logging

### LINKED ISSUE #217: Custom Logger
Is there any reason to enforce the use of a log.Logger?

As it is not an interface, it does not allow to log with a custom logger implementing the same methods, such as logrus/glog/others...

Why not exposing a logger interface that anyone could implement?

Bests.

kri5:
This PR changes the log.Logger usage to an interface
so that you can now use your own custom logger 

This fixes #217.

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,7 +9,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"log"
 	"math/rand"
 	"net/http"
 	"net/http/httputil"
@@ -112,9 +111,9 @@ type Client struct {
 	mu                        sync.RWMutex  // guards the next block
 	urls                      []string      // set of URLs passed initially to the client
 	running                   bool          // true if the client's background processes are running
-	errorlog                  *log.Logger   // error log for critical messages
-	infolog                   *log.Logger   // information log for e.g. response times
-	tracelog                  *log.Logger   // trace log for debugging
+	errorlog                  Logger        // error log for critical messages
+	infolog                   Logger        // information log for e.g. response times
+	tracelog                  Logger        // trace log for debugging
 	maxRetries                int           // max. number of retries
 	scheme                    string        // http or https
 	healthcheckEnabled        bool          // healthchecks enabled or disabled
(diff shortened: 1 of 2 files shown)


olivere:
@kri5 I actually think we could do that. We could tighten the interface even further, as I don't use the `...ln` and `Fatal...` functions anywhere IIRC.

The reason I'm hesitating is that I'm not sure how it will integrate into the various logging solutions out there. Some need key/value pairs in their interface (e.g. kit/log), but I guess that could be solved with a small wrapper. Some have log levels which is [controversial among Go developers](http://dave.cheney.net/2015/11/05/lets-talk-about-logging). I'm still looking for the idiomatic way of integrating logging into a library like Elastic.

kri5:
@olivere I think the interface is already pretty narrowed down. not much to implement to get logs.

olivere:
@kri5 Just discussing the issue on the Gophers channel with some valuable insights.

I think the general approach to integrating logging into a library in Go is to "just don't": Return errors and leave logging up to the implementer. But Elastic already does logging (and is way too chatty anyway) so we have to deal with it.

I did a quick grep over what kind of information Elastic is logging, and most of it is just informative: Response times, outcome, node health etc. Most of this is metrics/instrumentation, not logging (or actionable feedback). So in the long run, I'd love to make Elastic silent and only return errors to leave it up to the implementer to deal with it. In addition to that, reporting response times etc. will probably need an interface into a metrics/instrumentation library, e.g. [kit/metrics](https://godoc.org/github.com/go-kit/kit/metrics). I will create a separate issue for that.

To summarize:
I will try to follow your approach but trim down the interface to just one function so that you can provide a wrapper for your custom logger. It will probably look like this:

```go
type Logger interface {
  Printf(format string, v ...interface{})
}
```

So there's just one function to implement in your custom wrapper. [Here's an example](https://github.com/go-kit/kit/blob/master/loadbalancer/zk/logwrapper.go).

Later on, Elastic will be changed to properly separate logging from telemetry.

olivere:
Logging/telemetry issue is #221

olivere:
Thanks for your contribution.

PR #226, opened 2016-02-10
Improved documentation to better reflect functions' action.

kf6nux:
Current documentation suggests the function overwrites the search restriction.  Updated documentation to specify the function adds restrictions rather than overwriting them.

(This is not applicable to elastic release v3 as both the API and docs are already correct.)

### DIFF
--- a/search.go
+++ b/search.go
@@ -73,7 +73,7 @@ func (s *SearchService) Indices(indices ...string) *SearchService {
 	return s
 }
 
-// Type restricts the search for the given type.
+// Type adds a search restriction for the given type.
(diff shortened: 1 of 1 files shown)


olivere:
@kf6nux Oops. I must have missed this PR. You are right: v3 changed this (as [described here](https://github.com/olivere/elastic/blob/release-branch.v3/CHANGELOG-3.0.md#multiple-calls-to-variadic-functions)). And yes: Documentation can always be better.

olivere:
Fixed this evening in the new release. Thanks for reporting and fixing.

PR #235, opened 2016-02-26
fix incorrect logger for joining cluster

rschmukler:
a connection joining should probably be an info message as opposed to an error


olivere:
Yes, of course. Thanks for reporting. :-)

Will be fixed in the next release.

Issue #231, opened 2016-02-24
BulkService "After" function requests parameter is empty


dimfeld:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

[X] elastic.v2 (for Elasticsearch 1.x)
[ ] elastic.v3 (for Elasticsearch 2.x)

### Please describe the expected behavior
BulkService `After` requests parameter contains the requests that were sent.

### Please describe the actual behavior
The requests parameter is a zero-length slice.

Here are some simple logs I added in my code to illustrate the problem:
```
I0224 23:21:14.877035   29669 esindexer.go:77] requests contains 0 items
I0224 23:21:14.877098   29669 esindexer.go:78] response.Items contains 2000 items
```

### Any steps to reproduce the behavior?
Run a bulk request with an After step. Note that requests is empty.

I think this is because the end of `BulkService.Do` calls `reset` which resets the requests array, but `bulkWorker.commit` still references `w.service.requests` when calling `afterFn`. Since `Do()`and `reset` create a new list for `requests` instead of modifying it, this is probably as simple as saving a copy of `w.service.requests` before calling `Do()` and then using that saved copy in the call to `afterFn`. I'm not completely sure though. What do you think?

olivere:
Hi there! Sorry for the late reply. Looks like an error.

It's correct that the Bulk service resets the given requests slice. The official clients do the same. But of course, it shouldn't be reset before calling `After`.

TBH I don't quite like the `BulkProcessor` API as it is now. The error case with the `After` and `Before` callbacks are cumbersome and resemble the Java approach; they're not very Go-like. The `After` callback e.g. desperately needs a pointer to its `BulkProcessor` to be able to re-add failed documents. BulkProcessor is nice for fire-and-forget bulk indexing though.

Having said all that: Yes, it's a bug and I will fix it. However, expect some more changes to the Bulk Procesor.

dimfeld:
Yeah, I can see where you're coming from. I built my `After` function around an object that has a reference to the client, which works out well, but having the client included in the function call would be even better. I'm using glide for vendoring so will be able to adapt to any API changes on my own time :) Thanks for building this; aside from this minor hiccup it's working great!

olivere:
[...]

PR #238, opened 2016-02-29
Make sure bulk processor after callback gets requests on success

idubinskiy:
The bulk service resets its `requests` field after success, which means the bulk processor needs to store them locally so that the `requests` argument in the after callback is not empty on success.

### DIFF
--- a/bulk_processor.go
+++ b/bulk_processor.go
@@ -463,6 +463,9 @@ func (w *bulkWorker) commit() error {
 		w.p.beforeFn(id, w.service.requests)
 	}
 
+	// Store requests for after callback
+	reqs := w.service.requests
+
 	// Commit bulk requests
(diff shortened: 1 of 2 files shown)


olivere:
Thank you so much. I didn't have the time to fix it over the last weekend. I hope I find some time over the next days.

olivere:
See also #231.

idubinskiy:
Ah, I neglected to check the open issues. This commit should fix that bug, at least.

olivere:
Should be merged into 3.0.24 and released this evening. Will also backport to 2.x. Thanks again for reporting.

idubinskiy:
Thanks! Didn't add the tests?

olivere:
@idubinskiy Oops. My bad. I was in a hurry, but I add them as I write this.

idubinskiy:
Thanks!

Issue #243, opened 2016-03-10
Unable to connect to node which is not at the base URL.


shendaras:
### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[x] elastic.v3 (for Elasticsearch 2.x)

### Please describe the expected behavior
I am able to connect to an elasticsearch node running at http://localhost/elastic/

### Please describe the actual behavior
Currently, the path is stripped from the URL as part of canonicalize. This causes requests to go to http://localhost/ directly.

### Any steps to reproduce the behavior?
Create an elasticsearch node and have it accessible via a URL with a path.

olivere:
Looked into the official clients. They support this as well. Fixed in [this commit](https://github.com/olivere/elastic/commit/df416f02670ed201b1d04e9b198e14bc1f4c9478), also [backported to v2](https://github.com/olivere/elastic/commit/e713be938499714ee063069b868040d39c96cffd).

Thanks for being a contributor.

PR #244, opened 2016-03-10
Allow nodes which do not run at the base URL.

shendaras:
Previously, canonicalize removed the path component of the URL before returning.  This prevented connections to elasticsearch nodes which do not run at the base URL (e.g., "http://localhost/elastic/").  Without this change, any query to an elasticsearch node would have the wrong path and give strange results.  Change canonicalize to preserve the path.

### DIFF
--- a/canonicalize.go
+++ b/canonicalize.go
@@ -12,7 +12,7 @@ import "net/url"
 // protocol http or https.
 //
 // Example:
-// http://127.0.0.1:9200/path?query=1 -> http://127.0.0.1:9200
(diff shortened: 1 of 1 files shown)


olivere:
I haven't seen people using ES with a base path other than `/` in the last 4 years... :-)  but technically that should be okay. I will look into whether the official clients allow this because I want to stay in line with their design decisions (it played out several times).

shendaras:
I understand; the patch is simple and I can maintain it elsewhere if you feel like this isn't a fit for this project.  I agree that it isn't common -- I hadn't run into it until just recently -- but apparently some people do this.

olivere:
Closed (see #243).

PR #245, opened 2016-03-11
Make search_type optional, and allow sorting to be provided in Scroll requests.

samcday:
ScrollService is currently unusable against Elasticsearch 2.2 clients - many queries seem to result in an empty resultset when the deprecated `?search_type=scan` is used. This PR makes the search type customizable.

I also added support for providing sort criteria.

That said, it feels almost like `Scroll()` shouldn't be a top-level member of the Client anymore, but rather something you call off the `SearchService`, WDYT?

### DIFF
--- a/scroll.go
+++ b/scroll.go
@@ -15,20 +15,24 @@ import (
 
 // ScrollService manages a cursor through documents in Elasticsearch.
 type ScrollService struct {
-	client    *Client
-	indices   []string
-	types     []string
-	keepAlive string
-	query     Query
-	size      *int
-	pretty    bool
-	scrollId  string
+	client     *Client
+	indices    []string
+	types      []string
+	keepAlive  string
+	query      Query
+	sorts      []SortInfo
+	sorters    []Sorter
+	size       *int
+	pretty     bool
+	scrollId   string
+	searchType string
 }
 
 func NewScrollService(client *Client) *ScrollService {
 	builder := &ScrollService{
-		client: client,
-		query:  NewMatchAllQuery(),
+		client:     client,
+		query:      NewMatchAllQuery(),
+		searchType: "scan",
 	}
 	return builder
 }
@@ -83,6 +87,33 @@ func (s *ScrollService) ScrollId(scrollId string) *ScrollService {
 	return s
 }
 
+func (s *ScrollService) SearchType(searchType string) *ScrollService {
+	s.searchType = searchType
+	return s
+}
+
+// Sort adds a sort order.
+func (s *ScrollService) Sort(field string, ascending bool) *ScrollService {
+	s.sorts = append(s.sorts, SortInfo{Field: field, Ascending: ascending})
+	return s
+}
+
+// SortWithInfo adds a sort order.
+func (s *ScrollService) SortWithInfo(info SortInfo) *ScrollService {
+	s.sorts = append(s.sorts, info)
+	return s
+}
+
+// SortBy	adds a sort order.
+func (s *ScrollService) SortBy(sorter ...Sorter) *ScrollService {
+	s.sorters = append(s.sorters, sorter...)
+	return s
+}
+
+func (s *ScrollService) hasSort() bool {
+	return len(s.sorts) > 0 || len(s.sorters) > 0
+}
+
 func (s *ScrollService) Do() (*SearchResult, error) {
 	if s.scrollId == "" {
 		return s.GetFirstPage()
@@ -129,8 +160,9 @@ func (s *ScrollService) GetFirstPage() (*SearchResult, error) {
 
 	// Parameters
 	params := make(url.Values)
-	// TODO: ES 2.1 deprecates search_type=scan. See https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_21_search_changes.html#_literal_search_type_scan_literal_deprecated.
-	params.Set("search_type", "scan")
+	if s.searchType != "" {
+		params.Set("search_type", s.searchType)
+	}
 	if s.pretty {
 		params.Set("pretty", fmt.Sprintf("%v", s.pretty))
 	}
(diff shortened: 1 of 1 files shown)


olivere:
Yes, the situation about Scroll and Scan service is due to a design decision I took many years ago. It wasn't right in hindsight, but everything seems easy looking back :-)

But I'm surprised... why is Scan unusable in 2.2? They [deprecated `search_type=scan`](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_21_search_changes.html#_literal_search_type_scan_literal_deprecated) in 2.1 which usually means they remove it in the next major release. Do you have a test case that fails with 2.2?

Back to the Scroll and Scan issue. I think we should try hard to be in line with the ES REST API. If you want to scroll by searching with the `scroll` parameter, then that should be possible (and it hopefully is). The Search and Scroll services mainly sit on top of the Search service and are a helper to make iterating easy for the caller (e.g. you don't have to do all the juggling with the pagination token). Scroll and Scan mainly differ in that one returns one result after another (by getting a chunk of the results and automatically retrieving the next when required) and the other is giving you back the chunks as a whole.

In other issues and PRs I mentioned that I want to merge Scroll and Scan and turn it into an Iterator-style pattern that is e.g. easily consumable by many readers. This is still on the roadmap.

I was surprised to hear that some users out there are seeing JSON serialization in Go as the bottleneck (!). One solution to that could be to do it concurrently. However, I didn't have the time yet to dig into this. PRs welcome ;-)

sundarv85:
Hi @olivere 

> I was surprised to hear that some users out there are seeing JSON serialization in Go as the bottleneck (!)

Are there some issues related to this. Just interested in knowing more about it.

Thanks

olivere:
Hi @sundarv85. @avleen brought this up in #208.

samcday:
@olivere thanks for the background. Wasn't trying to criticise the decision, I only use a very small surface area of this impressively large Elasticsearch library :)

The reason I suggested merging the Scroll / Scan into SearchService is because I want to issue fairly complex queries and scroll through them. It turns out that that what I really want is what backs the SearchService - `SearchSource`. Is there a reason why it hasn't been plumbed into `ScrollService`? I would love to be able to specify the following when I scroll:

 * Aggregations
 * Field names
 * terminateAfter param

olivere:
@samcday No worries. I didn't read any criticism into your comments. Even if there were, it'd probably be just constructive criticism ;-)

Is this what you have in mind? (It should work, but I just wrote it down from my mind...)

```go
query := elastic.NewMatchAllQuery()
src := elastic.NewSearchSource(query).Sort("price", true)
scanner := client.Scan("my-index").Scroll("1m").SearchSource(src)
cursor, err := scanner.Do()
if err != nil {
	return err
}

for {
	docs, err := cursor.Next()
	if err == elastic.EOS {
		break
	}
	if err != nil {
		return err
	}

	for _, hit := range docs.Hits.Hits {
		// Process results here
	}
}
```

That should work, no?

However, if you can, don't sort. Why? If you ask for sorting, ES has to get all documents from the shards, bring them into order, then return a slice of them. From a performance perspective, that's rather bad. While not directly related to scrolling, there is a [breaking change regarding `from` and `size` in 2.1](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_21_search_changes.html#_from_size_limits) which tackles a similar problem. The [documentation of the Scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/2.2/search-request-scroll.html) also mentions that there is an optimization path in ES for scrolling (use sort order = `_doc`). Scan service has a similar [special case for optimization](https://github.com/olivere/elastic/blob/release-branch.v3/scan.go#L229) if you *don't* want sorting.

So if you don't need the guarantee of getting the documents in order, then don't. Scrolling will still give you a the guarantee to return each document once and only once.

samcday:
Oh, wow. Upon closer examination of `ScanService` it seems to already be doing exactly what I was making `ScrollService` do in this PR ... heh. Why does `ScrollService` exist at all? Is it deprecated?

olivere:
@samcday No, it's not deprecated. The `ScanService` returns search results in chunks while the `ScrollService` yields one after the other. The both exist because, well, I messed it up, really :-)  Hope I will fix it one day and merge it into a new service that can do both, is safe for multiple readers etc. (like I said a few comments before). I try hard not to introduce breaking changes. So removing or merging them is out of scope until the next major release. Adding things is okay, of course.

Notice that `SearchSource` contains all the things you can do with searching and is used (at least) under the covers of `SearchService` and `ScanService`. It resembles [`SearchSourceBuilder`](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/search/builder/SearchSourceBuilder.java) in the Java API.

samcday:
I'm still a little confused, sorry :(

| The `ScanService` returns search results in chunks while the `ScrollService` yields one after the other.

Can you elaborate on this? As far as I understand it, a "scan" in Elasticsearch returns results in an arbitrary order (but will still yield all results once everything has been walked) whereas a scroll will return results in a consistent order (which can be made more efficient by sorting on `_doc`). However, if you setup a `ScanService` with a sort order, then the `search_type=scan` param is not provided, which results in a scrolling request, with all the same cursor iteration stuff that `ScrollService` has, except `ScanService` already has `SearchSource` plumbed in, and thus supports a richer control over the querying. So, to summarise, it looks like there is *zero* reason to use `ScrollService` instead of `ScanService`, right?

olivere:
Yes, I agree it's confusing.

Short answer: Just use `ScanService`.

Long answer:

1. Both `ScanService` and `ScrollService` wrap the `_search` endpoint. So they DO the same.
2. Both use the scroll token to paginate the results. Again, no difference.
3. They differ in that `ScanService` returns [each result individually](https://github.com/olivere/elastic/blob/release-branch.v3/scan.go#L303) while `ScrollService` [yields a `SearchResult`](https://github.com/olivere/elastic/blob/release-branch.v3/scroll.go#L86), i.e. a slice of the results which you have to decode.
4. `ScrollService` [always uses `search_type=scan`](https://github.com/olivere/elastic/blob/release-branch.v3/scroll.go#L133), which is deprecated as of 2.1.
5. `ScanService` is more versatile in that it understands sorting and many more things.

I hope, one day, to get the best of both worlds (plus additional features like safety for multiple readers) into a single service, e.g. `IteratorService`.

HTH

olivere:
BTW: `ScrollService` has been refactored in 3.0.48. See [this example in the Wiki](https://github.com/olivere/elastic/wiki/ScrollParallel) to illustrate scrolling in parallel.

PR #247, opened 2016-03-18
Backport the TimeZone option for QueryStringQuery

stephenkubovic:
## What Changed & Why

Added support for setting the `"time_zone"` option for a query string query, via the `TimeZone` setter.  This option is supported in Elasticsearch 1.5.x and up, and we need to use it against an Elasticsearch 1.7 installation.  I've tried to mirror the time zone implementation from `release-branch.v3` as close as possible.

## Testing
There are not any tests for the existing options, so I have just ensured that [search_queries_query_string_test.go](./search_queries_query_string_test.go) still passes.  I've also used the fork in our code to ensure the option is set properly.

## Documentation
https://www.elastic.co/guide/en/elasticsearch/reference/1.7/query-dsl-query-string-query.html

### DIFF
--- a/search_queries_query_string.go
+++ b/search_queries_query_string.go
@@ -38,6 +38,7 @@ type QueryStringQuery struct {
 	rewrite                   string
 	minimumShouldMatch        string
 	lenient                   *bool
+	timeZone                  string
 }
 
 // Creates a new query string query.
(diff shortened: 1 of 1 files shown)


olivere:
Thanks for reporting. Merged in [this commit](https://github.com/olivere/elastic/commit/c03bd8ca6051b01de8ea02637a9055bc97d36c27). New version coming out later this weekend.

Issue #221, opened 2016-02-05
Logging and instrumentation


olivere:
The general approach to integrating logging into a library in Go is to "just don't": Return errors and leave logging up to the implementer. But Elastic already does logging (and is way too chatty anyway) so we have to deal with it.

A quick grep over what kind of information Elastic is logging shows up that most of it is just informative: Response times, outcome, node health etc. Most of this is telemetry, not logging (or actionable feedback). So in the long run, I'd love to make Elastic silent and only return errors and leave it up to the implementer to deal with it. In addition to that, reporting response times etc. will probably need an interface into a metrics/instrumentation library, e.g. [kit/metrics](https://godoc.org/github.com/go-kit/kit/metrics).

This issue is reserved for thoughts and discussion on the separation of logging and telemetry in Elastic.

vsiv:
Would be interesting if clients can pass a logger that the library can use for info/warns/errors. Developers can then decide whether they want to access logs and what to do with them.

olivere:
You can pass an instance that implements the `Logger` interface when creating a new client. It is described in the Wiki under [Logging](https://github.com/olivere/elastic/wiki/Logging). But it doesn't differentiate between info/warning/error.

vsiv:
Yes, I meant if the logger was aware of levels, that would be interesting and helpful to log for lib users. Happy to see this issue tracked. Thanks.

olivere:
In my apps I try to stay away from log levels, but it's a matter of personal taste.

Out of the blue, I feel like we could create a `LeveledLogger` interface with funcs for the various levels, or a single func that accepts the log level as the first argument. Right now, elastic only uses a trace log (which is used for debugging only), an info log, and an error log. So we could do something like this:

```go
type LeveledLogger interface {
  Errorf(format string, v ...interface{})
  Infof(format string, v ...interface{})
  Tracef(format string, v ...interface{})
}
```

One could set activate the leveled logger via `elastic.SetLeveledLogger(...)` on startup.

What I don't like is that we now have two interfaces for logging.

vsiv:
I too am not too stoked about the 2 logging interfaces. A similar option but more domain centric is something like what [ozzo-routing](https://github.com/go-ozzo/ozzo-routing) does... a logger for domain specific task. In elastic it could be network related error or an elastic error and we can treat them differently as opposed to classifying everything in standard log levels (unless its lib level tracing for debugging I suppose)

This might make it easier to troubleshoot as well.

```
router.Use(
[...]

PR #250, opened 2016-03-22
EstimatedSizeInBytes optimization

r--w:
This simple change, reduced number of GCs by half:
screenshot at mar 22 11-48-47

If you like it please port to the v3 version.

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -25,8 +25,6 @@ type BulkService struct {
 	timeout string
 	refresh *bool
 	pretty  bool
-
-	sizeInBytes int64
 }
 
 func NewBulkService(client *Client) *BulkService {
@@ -39,7 +37,6 @@ func NewBulkService(client *Client) *BulkService {
 
 func (s *BulkService) reset() {
 	s.requests = make([]BulkableRequest, 0)
-	s.sizeInBytes = 0
 }
 
 func (s *BulkService) Index(index string) *BulkService {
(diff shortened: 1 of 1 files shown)


olivere:
Looks good at first sight. I have to take a close look if this has side-effects with BulkProcessor.

olivere:
I implemented your proposal in [this commit](https://github.com/olivere/elastic/commit/56384553f13a5157c520f4ad78dd303ac8950030). Can you test it in your environment again? A new client will be released soon.

olivere:
@r--w Off-topic question: As you are handling quite a large amount of data with ES & Elastic and are actively monitoring your cluster, are you interested in giving me feedback on telemetry/instrumentation/metrics in Elastic (#221)?

It's not started yet, but the plan is to implement a mechanism into Elastic to expose interesting metrics, like response time etc. Implementation could be similar to the [`Logger`](https://github.com/olivere/elastic/blob/release-branch.v3/logger.go) interface, so that is (or can be made) easily consumable by standard monitoring solutions like Prometheus or simply expvar. I'm thinking of something like the [`metrics`](https://github.com/go-kit/kit/tree/master/metrics) package in [Go kit](http://gokit.io/), albeit simpler and without adding new dependencies into Elastic.

r--w:
Hi, thank you, I'll test It on Tuesday as well as #221 :)

olivere:
@r--w Had to change it once again; there was a regression that caused immense slow-down in bulk processor. See #252 and #253 for details.

r--w:
Hi, everything looks good. There is no significant difference between my version and the latest from github. Thanks.

screenshot at mar 29 11-27-08

PR #253, opened 2016-03-27
Cache calls to Source. Prevent unnecessary string creation

### LINKED ISSUE #252: Bulk Processor: Change in EstimatedSizeInBytes calculation causes immense slowdown

[X] elastic.v2 (for Elasticsearch 1.x)
[X ] elastic.v3 (for Elasticsearch 2.x)

Performance should about the same as before commit 56384553f13a5157c520f4ad78dd303ac8950030.

Bulk index performance is about 400x slower in my usage.

I'm using 8 workers with a BulkActions setting of 2000. The documents themselves are on the small side, with approximate 8 fields each. I'm also running with GOGC=500 which helps a lot with reducing percentage of time in GC here.

Since commitRequired is called every time a new request comes in, the commit referenced above effectively changes the algorithmic complexity from O(n), performing 1 size calculation on each request object, to O(n^2), recalculating for all queued requests each time a new request comes in. I narrowed it down to this particular commit, and in my usage case I see the performance approximately 400x slower, from 20000-30000 documents indexed per second down to 60-70.

dimfeld:
When `BulkableRequest.Source()` is called, the underlying objects
now cache the result, preventing future calls from going through the
JSON marshaling process yet again.

`estimatedSizeInBytes` now acts directly on the values returned
by Source instead of calling String to avoid creating a bit of garbage. 
In addition, this does undo the change I mentioned in #252 as well, so that `estimatedSizeInBytes` is only called once per `BulkableRequest`. 

In my testing, these changes combined give a 20-30% increase in documents indexed per second compared to code at commit 629ba26e184081ae0b8d577ea45a253a88dd8615 (before the change to `estimatedSizeInBytes`). 

Fixes #252.

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -118,10 +118,10 @@ func (s *BulkService) estimateSizeInBytes(r BulkableRequest) int64 {
 	lines, _ := r.Source()
 	size := 0
 	for _, line := range lines {
-		size += len(line)
+		// +1 for the \n
+		size += len(line) + 1
 	}
-	// +1 for the \n
-	return int64(1 + size)
+	return int64(size)
 }
 
 // NumberOfActions returns the number of bulkable requests that need to
--- a/bulk.go
+++ b/bulk.go
@@ -36,6 +36,8 @@ type BulkService struct {
 	timeout string
 	refresh *bool
 	pretty  bool
+
+	sizeInBytes int64
 }
 
 // NewBulkService initializes a new BulkService.
@@ -49,6 +51,7 @@ func NewBulkService(client *Client) *BulkService {
 
 func (s *BulkService) reset() {
 	s.requests = make([]BulkableRequest, 0)
+	s.sizeInBytes = 0
 }
 
 // Index specifies the index to use for all batches. You may also leave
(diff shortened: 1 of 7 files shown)


dimfeld:
Oh, and to give some more context, my use of the API is for a high-speed bulk indexing utility. It does not run persistently, but can run for up to an hour or two at a time for large jobs.

olivere:
@dimfeld Nice. Thanks a lot. I just released 3.0.28--2.0.36 will follow soon ([commit](https://github.com/olivere/elastic/commit/6ba207c39aef08e469d928c50a882a1bb5154f70)).

If I just had some more time for the project, I'd be looking into using [`sync.Pool`](https://golang.org/pkg/sync/#Pool) to reduce GC pressure in JSON serialization ([example](https://github.com/camlistore/camlistore/blob/master/pkg/httputil/httputil.go#L273-L304)) and perhaps even [ffjson](https://github.com/pquerna/ffjson) for the "hot paths"  in Elastic, e.g. Bulk and Search.

It's always interesting how people use Elastic and Elasticsearch. I will add a section in the issue template. My experience is that at least 50% use it for something different than just "plain old search".

r--w:
@olivere JSON marshaling definitely needs some more optimization, currently almost all "hot" parts are related to the JSON marshaling (but it's not your fault ;)
screenshot at mar 29 11-33-42

olivere:
@r--w Thanks for sharing. Some guys over at Etsy came to a similar conclusion [here]( https://github.com/olivere/elastic/pull/208#issuecomment-174590023). Maybe I need to invest some time into speeding up JSON serialization, at least for Bulk and Search operations.

dimfeld:
In addition to ffjson, I came across https://github.com/mailru/easyjson recently which is a similar concept, using code generation, but purports to be faster. Haven't checked it out in detail though.

PR #255, opened 2016-03-31
fix filtered query allow empty query

ishare:
filtered query allows without query part, only filters.


olivere:
@ishare Fixed in 2.0.37 which I just release. Thanks for being a contributor.

PR #262, opened 2016-04-11
Create ClientOptionFunc to set custom request headers

jdbalistreri:
The Elasticsearch cluster I'm calling with the elastic package is going to start enforcing that custom headers are set on each request. I've updated the elastic package to allow custom headers to be passed to a Client as a ClientOptionFunc.

### DIFF
--- a/client.go
+++ b/client.go
@@ -108,30 +108,31 @@ type Client struct {
 	conns   []*conn      // all connections
 	cindex  int          // index into conns
 
-	mu                        sync.RWMutex  // guards the next block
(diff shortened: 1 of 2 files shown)


olivere:
Hi there. You can do that already by using your own `HttpClient`. Here's an [example of it](https://github.com/olivere/elastic/wiki/HttpTransport#http-basic-authentication).

I see the value/simplicity of adding header management directly into to the client. But I refrain from adding this because we might end up replicating all the things you could do with a HTTP request.

Please reply if this doesn't work for you.

PR #275, opened 2016-05-06
searchResult.Hits contains multiple values in v.3

sun-kuet:
searchResult.Hits contains below three values although there might be no result. 
These three values are: 
      "total": 0,
      "max_score": null,
      "hits": []

So, this not going to be Nil. So, devs may try other options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ for _, item := range searchResult.Each(reflect.TypeOf(ttyp)) {
 fmt.Printf("Found a total of %d tweets\n", searchResult.TotalHits())
 
 // Here's how you iterate through results with full control over each step.
-if searchResult.Hits != nil {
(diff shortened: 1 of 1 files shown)


olivere:
You're correct. It's not `nil` for a reason IIRC (you'd have way to many if clauses otherwise):

```go
if res != nil && res.Hits != nil && res.Hits.TotalHits > 0 { ... }
```

The "best" way to check is to do this:

```go
if res.TotalHits() > 0 { ... }
```

But I will check again next week when I'm back from vacation.

sun-kuet:
Did you review it ?

olivere:
Need to change it in probably 4 different places: gh-pages, v3, v2, Wiki... Will do later.

Thanks for your support.

PR #276, opened 2016-05-06
Adds GeoShape query searching for points and polygons

zombor:


olivere:
Sorry for the delay. Will review next week when I'm back from my vacation.

A variable declaration like `[][][]float64` looks horrible to my eyes... Can you try to change it to a `GeoShapeQuery` that resembles [`GeoShapeQueryBuilder`](https://github.com/elastic/elasticsearch/blob/2.3/core/src/main/java/org/elasticsearch/index/query/GeoShapeQueryBuilder.java) from the Java API? It's some more work, but it built all queries around the builders in the Java API.

zombor:
it looks like the constructor for `GeoShapeQueryBuilder` in the java api takes in a shape. You want it to mimic that? I think the `[][][]float64` would still live somewhere in some form since that's how polygons work.

When you get back I'd love to discuss the design in more detail. GeoShape is tricky!

olivere:
Yes, I agree. It's tricky. Maybe we can  e.g. reuse GeoPoint as it's already there for other Geo queries.

Will be back next week.

olivere:
Just looked at it again. ShapeBuilder is quite some work. Now I remember why I skipped it until now ;-)

I currently think the best way to deal with it is to make a generic `GeoShapeQuery` (like you already did) and provide a set of builders for the various shapes (unlike the `SetPoint` / `SetPolygon` that your PR provides). Something along the lines of:

```go
// GeoShape provides a geographic shape, e.g. a Circle or a Polygon.
// Use the builders in the geo package to create one of these.
type GeoShape interface {
  Source() (interface{}, error)
}

// Shape provides the shape for the GeoQuery.
func (q *GeoShapeQuery) Shape(shape GeoShape) *GeoShapeQuery {
  q.shape = shape // used in GeoShapeQuery.Source to generate the JSON shape data
  return q
}

...

package geo

// Circle implements GeoShape and returns the shape of a circle.
type Circle struct { ... }

func (c *Circle) Source() (interface{}, error} { ... }

// Polygon implements GeoShape and returns the shape of a polygon.
type Polygon struct { ... }

func (p *Polygon) Source() (interface{}, error} { ... }

```


I'll probably create a separate branch for this as its unlikely that I get it right on my first try.

olivere:
For self-reference: [GeoQuery documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-shape-query.html).

zombor:
Sounds good. Pretty similar to what I implemented, except the shape builders are not in the query type.

Just a note/nitpick, there's no `Circle` type in the GeoJSON standard.

olivere:
Yes. Pretty similar. Elastic is really no rocket science :-)

Re: Circle: Didn't know that. At least ES has a [`CircleBuilder`](https://github.com/elastic/elasticsearch/blob/2.3/core/src/main/java/org/elasticsearch/common/geo/builders/CircleBuilder.java).

zombor:
> Re: Circle: Didn't know that. At least ES has a CircleBuilder.

Interesting. In the main site docs, it says it only supports GeoJSON (`Similar to the geo_shape type, the geo_shape Filter uses GeoJSON to represent shapes.`), but it looks like they implemented their own [additional types](https://github.com/elastic/elasticsearch/blob/2.3/docs/reference/mapping/types/geo-shape.asciidoc#circle) that geojson lacks. TIL.

olivere:
The documentation site for ES is really really good but it's not the best place if you're looking for implementation details. From time to time I find something is missing or ambiguous. That's why I take the Java API as reference for Elastic.

aubm:
Hey, any news about that one?
I kind of need it for a project ^^

olivere:
It's still missing, sorry. It's more work than I though initially.

olivere:
@aubm A workaround is to use [`RawStringQuery`](https://github.com/olivere/elastic/blob/release-branch.v3/search_queries_raw_string.go) and generate the underlying query yourself. It's not ideal, of course. But if you only need a limited set of Geo features, that might help. See the [tests](https://github.com/olivere/elastic/blob/release-branch.v5/search_queries_raw_string_test.go) for some examples.

aubm:
Good to know, thanks you for the trick! :)

olivere:
@philippecarle Sorry, this has been put aside for far too long.

I wasn't happy with the PR because it solves only one specific use case. Elasticsearch has [`ShapeBuilders`](https://github.com/elastic/elasticsearch/blob/5.6/core/src/main/java/org/elasticsearch/common/geo/builders/ShapeBuilders.java) for all kinds of shapes (circles, multi-points etc.). And I didn't have the time to build a proper `ShapeBuilder` for all these shapes in Elastic.

So the current workaround is to use [`RawStringQuery`](https://github.com/olivere/elastic/blob/release-branch.v3/search_queries_raw_string.go) and build your own shape queries.

If you want to help, we basically need to create Go code for all these shapes listed [here](https://github.com/elastic/elasticsearch/tree/5.6/core/src/main/java/org/elasticsearch/common/geo/builders).

philippecarle:
Hi @olivere don't be sorry, I think that implementing this kind of feature is quite tough. I'd love to help, but 1) I'm quite new to Elastic so it won't be very easy for me to get the whole scope of geo search and 2) reading the ES Java code to create the related Go code :thinking: ? But I will give it a try !  I'm currently using `RawStringQuery` but it's not very satisfying as soon as you have multiple parameters / options to pass or not...

olivere:
@philippecarle I see. Actually, if it makes things simpler for you: You can use your own struct to be used as part of the DSL in Elastic. The only thing you need to implement is the [`elastic.Query`](https://github.com/olivere/elastic/blob/release-branch.v5/query.go) interface, which comes down to only one function: `Source`. The output of `Source` will be serialized into JSON and is used in the query. So if you have a more natural way to generate the query, you don't need to rely on `RawStringQuery`, but can build your own `Query` as well.

zombor:
Yeah, that's basically what I did in the code in this PR, built my own thing that generated the right JSON that ES could use. We've moved our project off of ES though, so I won't be able to contribute anything more than this at this time.

philippecarle:
Thanks a lot for the advice, it works like a charm. It is obviously my first version of it and it would deserve to create dedicated object to manage each key of the maps so that it could be marshalled and maintainable more easily but it definitely looks way less ugly than formatting json strings and passing vars to it. 
https://github.com/populin/popul.in/blob/develop/services/geography/elastic/geo_shape.go

olivere:
I'm closing this for now. It is too old, and I'd rather look for a more advanced version that has [builders for all the different shapes](https://github.com/elastic/elasticsearch/tree/master/server/src/main/java/org/elasticsearch/common/geo/builders). Creating a new followup issue.

PR #277, opened 2016-05-07
Feature/bulk delete request parent

j16r:
When deleting children records in ES, you must specify the id of the parent with the _parent field. This means that you are currently unable to delete children records using the BulkDeleteRequest interface. (unless there is a way with Routing?).

### DIFF
--- a/bulk_delete_request.go
+++ b/bulk_delete_request.go
@@ -18,6 +18,7 @@ type BulkDeleteRequest struct {
 	index       string
 	typ         string
 	id          string
+	parent      string
 	routing     string
 	refresh     *bool
 	version     int64  // default is MATCH_ANY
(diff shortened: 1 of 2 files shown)


olivere:
IIRC setting the routing and setting the parent is basically the same with child records, as the parent is used as the routing key for child objects. The reason for that is that ES needs to make sure that parent and child documents are stored in the same shard for performance reasons.

But it shouldn't be a problem to have both in a client like Elastic, as long as you don't set parent and routing to different values (which should fail on the server side).

I will review this when I'm back in the office after my vacation this week.

olivere:
Just for reference: [`DeleteRequest.parent`](https://github.com/elastic/elasticsearch/blob/2.3/core/src/main/java/org/elasticsearch/action/delete/DeleteRequest.java#L161).

olivere:
*sigh* Merged into master. Will fix this and merge into release-branch.v3 then.

olivere:
Published 2.0.43 (for ES 1.x) and 3.0.36 (for ES 2.x) that adds a `Parent` property to `BulkDeleteRequest`.

Notice that your PR was a diff to the master branch, which is defunct. The branches to work with are release-branch.v2 and release-branch.v3.

PR #281, opened 2016-05-11
Remove search_type=scan from Scroll, it causes no results to be returned

j16r:
I've just had to remove the search_type=scan parameter for scroll against my elasticsearch version (2.3.2) as its presence caused no results to be returned.

Was this just left in for backwards compatibility?

### DIFF
--- a/scroll.go
+++ b/scroll.go
@@ -129,8 +129,6 @@ func (s *ScrollService) GetFirstPage() (*SearchResult, error) {
 
 	// Parameters
 	params := make(url.Values)
-	// TODO: ES 2.1 deprecates search_type=scan. See https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_21_search_changes.html#_literal_search_type_scan_literal_deprecated.
-	params.Set("search_type", "scan")
 	if s.pretty {
 		params.Set("pretty", fmt.Sprintf("%v", s.pretty))
 	}
--- a/scroll_test.go
+++ b/scroll_test.go
@@ -53,7 +53,7 @@ func TestScroll(t *testing.T) {
 	if res.Hits.TotalHits != 3 {
 		t.Errorf("expected results.Hits.TotalHits = %d; got %d", 3, res.Hits.TotalHits)
 	}
-	if len(res.Hits.Hits) != 0 {
+	if len(res.Hits.Hits) != 1 {
 		t.Errorf("expected len(results.Hits.Hits) = %d; got %d", 0, len(res.Hits.Hits))
 	}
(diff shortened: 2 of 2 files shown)


j16r:
So I've changed the tests in this one, and this change makes sense to me - here is how I'm reading the code:

1. `Do` initial search on all results, `scrollID` is blank, this should call `GetFirstPage` and since `Size(1)` should return 1 hit
2. Create a new scroll with `scrollID` from previous search
3. `Do` should see a non blank `scrollID` and do `GetNextPage` which should attempt to return the remaining results, of which there are 2.

Am I misunderstanding the Scroll API or is correct?

olivere:
Does the test code in https://github.com/olivere/elastic/blob/release-branch.v3/scroll_test.go help answer your question?

I'm currently on vacation, so a short reply. I'll be back next week.

j16r:
Thanks for the quick reply, that's actually the part I'm having difficulty with. From my understanding, it should work as outlined in the steps above.

Returning no results for the first page seems a little unusual to me.

olivere:
Whether the first query returns any results or not is up to Elasticsearch, not specific to Elastic. You can simulate the API calls by issuing curl requests from the command line and compare with what Elastic returns (see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html).

In the end: Does it matter if the 1st query returns any results, as long as all documents are returned eventually?

j16r:
Well, yeah it kinda does - in my case it means doing 2 requests instead of 1 for a larger number of queries!

However that was not my motivation for making this pull request, with the `search_type=scan` my elasticsearch was returning no results. The `Scroll` API change was a side effect.

olivere:
Scroll API is for iterating through a large number of results, potentially hundreds of millions of documents. If N (=the size of documents returned per shard per request) is very large, N+1 becomes insignificant. If you have many requests returning a small number of documents, a filtered query, potentially executed in parallel or as a multi-search, seems like a better option to me. However, I'm not aware of your special use case.

Search type scan is deprecated as of 2.1 (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-type.html#scan) and will probably be removed in 5.0. I'm currently not aware of any problems with it. I will check when I have the time for it. A failing test case would be appreciated.

j16r:
This is indeed true, though I rarely know in advance if the set of things I want to search is going to hit that limit - and querying to find out would be yet more requests.

olivere:
I just released 3.0.35 which removes the `search_type=scan` from Scroll. Thanks for your contribution.

PR #292, opened 2016-05-25
Add Field Stats API

tusharm:
Hi, 

I've been working on Field Stats support in the client. Note that it doesn't currently handle indices options such as expand wildcards, ignore unavailable etc. but I hope to add these soon.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1278,7 +1278,11 @@ func (c *Client) Percolate() *PercolateService {
 // TODO Search Shards API
 // TODO Search Exists API
 // TODO Validate API
-// TODO Field Stats API
+
+// FieldStats returns statistical information about fields in indices
+func (c *Client) FieldStats(indices ...string) *FieldStatsService {
+	return NewFieldStatsService(c).Indices(indices...)
+}
 
 // Exists checks if a document exists.
 func (c *Client) Exists() *ExistsService {
--- a/field_stats.go
+++ b/field_stats.go
@@ -1,0 +1,180 @@
+package elastic
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/atlassian/elastic/uritemplates"
+	"net/http"
+)
+
+const (
+	ClusterLevel = "cluster"
+	IndicesLevel = "indices"
+)
+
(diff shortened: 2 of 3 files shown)


olivere:
Fantastic. Thanks for being a contributor.

I'm on a short trip over the weekend, so I'm probably not being able to merge until next week.

olivere:
Hmm... I think I merged too early and we need to have another look. There is a [REST API Specification for the Field Stats API](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/field_stats.json) which has more fields than your PR.

Also, did you notice that there's a [generator](https://github.com/olivere/elastic/tree/generate-api/generator) that takes the REST API spec and generates a Go wrapper for the API? The generated code looks a bit different to your manual implementation. We need to check the differences carefully.

Besides, there are subtle differences in the usage pattern of your implementation and that of other services. E.g. you expect the body of the Field Stats API to always be of type `FieldStatsRequest`. However, generated services typically allow to construct the body themselves and either set it via `BodyString(...)` or `BodyJson(...)`. That doesn't mean that `FieldStatsRequest` is a bad idea: It's not. It just means that it's just a helper to construct the body of the request--if the end user wants.

Another thing I found was that you validate the `level` parameter. The [REST API](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/field_stats.json) specifies a default value of `cluster`. However, I typically ignore this as ES always has a default on the server-side. If the server-side later decides that the default is wrong, the client would ignore that change. Hence I typically accept to not use the default value and let the server decide.

I will push an update to `release-branch.v3` later. It'd be very helpful if you could review my changes and see if it works for you. Okay?

olivere:
Can you please review the changes in the [`field-stats-api`](https://github.com/olivere/elastic/tree/field-stats-api/) branch?

olivere:
Actually, let's stop discussing on this PR. I merged my changes into `release-branch.v3` and we should simply do clean up via issues. The Field Stats API is a new addition so we cannot break existing clients, plus: it's not released to gopkg.in yet.

tusharm:
Hey, thanks for pointing those out. I'll take a look.

PR #308, opened 2016-06-21
support Multi termvectors API

suzuken:
supports https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-termvectors.html

### DIFF
--- a/client.go
+++ b/client.go
@@ -1239,7 +1239,14 @@ func (c *Client) ReindexTask() *ReindexService {
 }
 
 // TODO Term Vectors
-// TODO Multi termvectors API
+
+// MultiTermVectors returns information and statistics on terms in the fields
+// of multiple documents.
+func (c *Client) MultiTermVectors(index, typ string) *MultiTermvectorService {
+	builder := NewMultiTermvectorService(c)
+	builder = builder.Index(index).Type(typ)
+	return builder
+}
 
 // -- Search APIs --
 
--- a/mtermvector.go
+++ b/mtermvector.go
@@ -1,0 +1,443 @@
+package elastic
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/olivere/elastic/uritemplates"
+)
+
+// MultiTermvectorService returns information and statistics on terms in the
+// fields of a particular document. The document could be stored in the
+// index or artificially provided by the user.
+//
+// See https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-termvectors.html
+// for documentation.
+type MultiTermvectorService struct {
+	client          *Client
+	pretty          bool
+	index           string
+	typ             string
+	fieldStatistics *bool
+	fields          []string
+	ids             []string
+	offsets         *bool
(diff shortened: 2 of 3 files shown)


olivere:
That's looking really good, thank you for your support. :-)

Can you double check that all properties defined in the [REST API spec for `mtermvectors`](https://github.com/elastic/elasticsearch/blob/2.4/rest-api-spec/src/main/resources/rest-api-spec/api/mtermvectors.json) are available?

Also: Did you notice there's a generator that takes a [REST API spec file](https://github.com/elastic/elasticsearch/tree/2.4/rest-api-spec/src/main/resources/rest-api-spec/api) as input and emits Go code for a service as output? You can see it in the [generate-api branch](https://github.com/olivere/elastic/tree/generate-api/generator). It makes creating missing services a bit easier than doing everything manually.

suzuken:
Thank you for your reply. 

> Can you double check that all properties defined in the REST API spec for mtermvectors are available?

Oh, OK. I'll see it.

> Also: Did you notice there's a generator that takes a REST API spec file as input and emits Go code for a service as output? You can see it in the generate-api branch. It makes creating missing services a bit easier than doing everything manually.

It seems good. This PR implemented by my hands, I'll try it.

olivere:
It's okay. No need to change it if all properties are there. The generator is simply for creating a skeleton for you.

I'll integrate it with the next release. Thanks for being a contributor.

suzuken:
Thank you for your comments. I've added some properties related [elasticsearch/mtermvectors.json at 2.4 · elastic/elasticsearch](https://github.com/elastic/elasticsearch/blob/2.4/rest-api-spec/src/main/resources/rest-api-spec/api/mtermvectors.json) .

olivere:
So I've just added your PR. I've changed a few things; maybe you can double check if that makes sense.

Most changes are stylistic or fix a typo. However there's a significant change in verifying the request. In my understanding, it should be okay to call the `/_mtermvectors` endpoints in 3 possible ways (as described [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-termvectors.html) and in the [REST API spec](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/mtermvectors.json).

1. Index and type: `/{index}/{type}/_mtermvectors` (no docs in body required)
2. Index but no type: `/{index}/_mtermvectors` (pass docs with type in the body)
3. No index and no type: `/_mtermvectors` (pass docs in the body)

What's not possible is 4. No index but a type. I've changed the Validate func accordingly and implemented a test for that behavior.

I've also changed the call in `client.go` to not require index and type, i.e. change

```go
func (c *Client) MultiTermVectors(index, typ string) *MultiTermvectorService { ... }
```

... to ...

```go
func (c *Client) MultiTermVectors() *MultiTermvectorService { ... }
```

I will release a new version of Elastic later this evening. If something's wrong, we'll fix it in the next release.

suzuken:
looks good to me. thanks! :)
2016/06/24 20:18 "Oliver Eilhard" <notifications@github.com>:

PR #309, opened 2016-06-22
Fix json encode error

djgilcrease:
When doing a term agg, then attempting to encode the result back to json I was getting an error

Elastic was returning
```json
{
	"buckets": [{
		"key": "DJ-2",
		"doc_count": 24697
	}],
	"doc_count_error_upper_bound": 0,
	"sum_other_doc_count": 0
}
```

So when it decoded, it saw the key field and assumed it was a number and thus when I attempted to re-encode the Aggregations back to json I was getting `json: invalid number literal "DJ-2"`

This change fixes the error and produces

```json
{
	"Aggregations": {
		"buckets": [{
			"key": "DJ-2",
			"doc_count": 24697
		}],
		"doc_count_error_upper_bound": 0,
		"sum_other_doc_count": 0
	},
	"DocCountErrorUpperBound": 0,
	"SumOfOtherDocCount": 0,
	"Buckets": [{
		"Aggregations": {
			"doc_count": 24697,
			"key": "DJ-2"
		},
		"Key": "DJ-2",
		"KeyAsString": "DJ-2",
		"KeyNumber": 0,
		"DocCount": 24697
	}],
	"Meta": null
}
```

This change has been tested against

```json
{
	"name": "silver-elastic-01",
	"cluster_name": "silver-logging",
	"version": {
		"number": "2.0.0",
		"build_hash": "de54438d6af8f9340d50c5c786151783ce7d6be5",
		"build_timestamp": "2015-10-22T08:09:48Z",
		"build_snapshot": false,
		"lucene_version": "5.2.1"
	},
	"tagline": "You Know, for Search"
}
```

and

```json
{
	"name": "Windeagle",
	"cluster_name": "elasticsearch",
	"version": {
		"number": "2.3.3",
		"build_hash": "218bdf10790eef486ff2c41a3df5cfa32dadcfde",
		"build_timestamp": "2016-05-17T15:40:04Z",
		"build_snapshot": false,
		"lucene_version": "5.5.0"
	},
	"tagline": "You Know, for Search"
}
```

### DIFF
--- a/search_aggs.go
+++ b/search_aggs.go
@@ -1010,6 +1010,12 @@ func (a *AggregationBucketKeyItem) UnmarshalJSON(data []byte) error {
 	if v, ok := aggs["key"]; ok && v != nil {
 		json.Unmarshal(*v, &a.Key)
 		json.Unmarshal(*v, &a.KeyNumber)
+		# Verify KeyNumber is really a num
+	        if _, err := a.KeyNumber.Float64(); err != nil {
+	            a.KeyNumber = ""
+	        }
+	        # Always set KeyAsString
+	        json.Unmarshal(*v, &a.KeyAsString)
 	}
 	if v, ok := aggs["key_as_string"]; ok && v != nil {
 		json.Unmarshal(*v, &a.KeyAsString)
--- a/search_aggs.go
+++ b/search_aggs.go
@@ -1010,11 +1010,11 @@ func (a *AggregationBucketKeyItem) UnmarshalJSON(data []byte) error {
 	if v, ok := aggs["key"]; ok && v != nil {
 		json.Unmarshal(*v, &a.Key)
 		json.Unmarshal(*v, &a.KeyNumber)
-		# Verify KeyNumber is really a num
+		// Verify KeyNumber is really a num
 	        if _, err := a.KeyNumber.Float64(); err != nil {
 	            a.KeyNumber = ""
 	        }
-	        # Always set KeyAsString
+	        // Always set KeyAsString
 	        json.Unmarshal(*v, &a.KeyAsString)
 	}
 	if v, ok := aggs["key_as_string"]; ok && v != nil {


olivere:
I'm afraid the structures of Elastic are not prepared to be serialized back into JSON. E.g. there's a lot of `omitempty` fields missing all over the structs. 

In other words: The use case for Elastic is to read data from ES and consume it in your application. If you need to pass data on to other layers in your application, it is best to do that in your application.

djgilcrease:
That is all well and good but there is still a bug in your decode. You make the faulty assumption that if
```go
if v, ok := aggs["key"]; ok && v != nil
```
is true then the value of v must be a json.Number, which is NOT the case.

olivere:
No. I'm making no assumptions at all. I deserialize the result from ES into the various types that ES could return and, deliberately, throw away any errors. I make it up to you, the application developer, to take the values and potentially check if they are valid. You're the only one who knows if a `key` is e.g. a numeric type or a string, and if `KeyNumber` can be used or not.

djgilcrease:
With this change KeyNumber is not set unless the value is actually a json.Number. Since KeyNumber is NOT something elasticsearch ever returns and is something unique to this library it would make sense to me that it is properly validating it's own custom values.

olivere:
When do I clear the `KeyNumber` value then? When `KeyNumber.Float64()` fails, or when `KeyNumber.Int64()` fails, or both? It's undecidable from within Elastic. `KeyNumber` is simply a helper. Use it when you need to, otherwise ignore it.

What's wrong with checking `KeyNumber` in the application?

djgilcrease:
KeyNumber.Float64() will succeed even if the value is an int64.

My application is basically a thin proxy in front of elastic search, and returning json. I do not want to have to loop over the 10k records for every request and explicitly set KeyNumber as an empty string or 0 just so I can render the result I got back.

olivere:
So you're asking me to change Elastic so that you don't have to iterate through 10k records. But it's okay for every other user of Elastic to hit the performance penalty of [`strconv.ParseFloat`](https://golang.org/src/encoding/json/decode.go?s=6063:6105#L169) without ever needing it?

I guess this discussion is going nowhere. Please don't take it personally.

I'm afraid if you need a proxy for ES, Elastic is not the right choice. It is not suited as a proxy.

PR #314, opened 2016-06-30
Support _nodes/stats for elasticsearch 1.x series.

jdelgad:
Follow pattern in nodes_info.go.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1456,6 +1456,12 @@ func (c *Client) NodesInfo() *NodesInfoService {
 	return NewNodesInfoService(c)
 }
 
+// NodesStats retrieves one or more or all of the cluster nodes information.
+func (c *Client) NodesStats() *NodesStatsService {
+	return NewNodesStatsService(c)
+}
+
+
 // Reindex returns a service that will reindex documents from a source
 // index into a target index. See
 // http://www.elastic.co/guide/en/elasticsearch/guide/current/reindex.html
--- a/nodes_stats.go
+++ b/nodes_stats.go
@@ -1,0 +1,601 @@
+// Copyright 2012-2015 Oliver Eilhard. All rights reserved.
+// Use of this source code is governed by a MIT-license.
+// See http://olivere.mit-license.org/license.txt for details.
+
+package elastic
+
+import (
+	"fmt"
+	"log"
+	"net/url"
+	"strings"
+
+	"gopkg.in/olivere/elastic.v2/uritemplates"
+)
+
+type Level int64
+
+const (
+	NODE Level = iota
+	INDICES
+	SHARDS
+)
+
+var (
+	_        = fmt.Print
+	_        = log.Print
+	_        = strings.Index
+	_        = uritemplates.Expand
+	_        = url.Parse
+	levelMap = map[Level]string{NODE: "NODE", INDICES: "INDICES", SHARDS: "SHARDS"}
+)
+
+// NodesStatsService allows to retrieve one or more or all of the
+// cluster nodes information.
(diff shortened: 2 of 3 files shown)


jdelgad:
Pending acceptance I plan to issue a PR for v3 supporting elasticsearch 2.x.

olivere:
Thank you. I'll hopefully find some time over the weekend for review.

You could double-check that the API contains all properties defined in the [REST API spec](https://github.com/elastic/elasticsearch/tree/2.4/rest-api-spec) for that endpoint.

Also notice that there is a Go generator that builds a skeleton for API's based on the REST API spec. You can find it in the [`generate-api` branch](https://github.com/olivere/elastic/tree/generate-api/generator). No need to implement everything manually. :-)

jdelgad:
@olivere I've found a few discrepancies. I'll mention you again when all of the API spec has been met.

olivere:
Take your time. I'll be out of office for a few days anyway.

jdelgad:
@olivere Added unit test similar to that of what is already in the repo. Implemented the REST API and included it in the commit message based on https://github.com/elastic/elasticsearch/blob/1.7/rest-api-spec/api/nodes.stats.json

I did notice that the `groups` parameter is marked as a `boolean`, although it is a list as referenced in its documentation https://www.elastic.co/guide/en/elasticsearch/reference/1.7/cluster-nodes-stats.html

olivere:
I'm sorry. I cannot merge this. There are a few problems.

First, please remove the `iota` values for `Level`. I tried that a few years ago, and it's just not worth it in the client. Simply let the caller pass a `string`.

Second, the Nodes Stats API has many endpoints. You need to use the one that matches the parameters given by the caller. You always assume `/_nodes/{node_id}/stats/{metric}/{index_metric}` is okay to use. But that doesn't work when `{node_id}` or `{metric}` isn't specified by the caller. We need to take this into consideration.

Third, although I didn't check in detail: Many of the structures returned by Nodes Stats API are already returned by Nodes Info API as well. I guess I have to figure out if duplicating the structs is really the better approach.

I'm thankful for your support. But it's hard to keep on top of all the endpoints ES provides and still keep everything under control. I think we'll have to have another round on Nodes Stats API.

jdelgad:
Comment 1 I've addressed, but will wait to put it up for now.

For comment 2 I'm a bit confused. Looking at https://github.com/olivere/elastic/blob/release-branch.v2/nodes_info.go#L85 I'm doing the same here https://github.com/olivere/elastic/pull/314/files#diff-3910d97348111eb8bb2b1710f118a56dR156

When `{node_id}` or `{metric}` isn't specified, then `_all` is used in its place. Perhaps I'm not understanding. Can you give me a code example using my PR so I can understand what you are saying doesn't work?

Question 3, I have time to work on merging the two. Just let me know what you would like and I'll go forward. Once we have this solidified for ES 1.x then I'll continue working on the ES 2.x branch.

olivere:
Sorry for the delay.

Regarding comment 2: I didn't see that you initialized `metric` and `nodeId` with `"_all"` in the initializer. Sorry for that.

Re question 3, it looks like both endpoints return the same structures. What needs to be done is first to check if that's a correct assumption by looking at the Java source code. It takes some experience to find the correct reference in the [elasticsearch repository](https://github.com/elastic/elasticsearch) (use the current `2.x` branch for elastic.v3). Then we'd need to check if the properties in our structs are in sync with those in the Java source.

I'll eventually do that, too. No worries if you don't feel like you can do that. Unfortunately, I'm rather busy at the moment, so it'll take some time.

olivere:
Backport should be available in 2.0.54.

PR #316, opened 2016-07-02
Add index analyze API

huydx:
This PR implement index analyze API, and support v3 only. (target for release-branch.v3)
Please take a look @olivere.

Thank you in advance.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1372,6 +1372,15 @@ func (c *Client) IndexPutSettings(indices ...string) *IndicesPutSettingsService 
 	return NewIndicesPutSettingsService(c).Index(indices...)
 }
 
+// Index Analyzer analyzer text for debug purpose
+func (c *Client) IndexAnalyze(indices ...string) *IndicesAnalyzeService {
+	if len(indices) > 0 {
+		return NewIndicesAnalyzeService(c).Index(indices[0])
+	} else {
+		return NewIndicesAnalyzeService(c)
+	}
+}
+
 // Optimize asks Elasticsearch to optimize one or more indices.
 // Optimize is deprecated as of Elasticsearch 2.1 and replaced by Forcemerge.
 func (c *Client) Optimize(indices ...string) *OptimizeService {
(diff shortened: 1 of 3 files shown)


olivere:
Sorry, I'm busy the next two weeks. Might take some time.

olivere:
Just a few notices:

1. Can you please compare the list of parameters with those publically specified in the [REST API specification](https://github.com/elastic/elasticsearch/tree/2.4/rest-api-spec) of that endpoint?
2. You can use the [generate-api branch](https://github.com/olivere/elastic/tree/generate-api/generator) to generate a skeleton of your service given the REST API spec for that endpoint. You might need to update the REST API spec files there first. Compare that with your service implementation, which is presumably implemented manually.
3. I don't like the `Do` vs. `DoWithExplain`. Can we merge that into one and use a parameter to differentiate?

Thank you.

huydx:
@olivere Thanks for your comments, really helpful informations.
1.2. I will take a look at REST API specification
3. Currently analyze api with explain set as true return completely different json, so merge it as one may require user to cast interface{} to appropriate type when get response, which seems not good. Do you have any idea?

olivere:
It seems that the Java API will generate two embedded structures into the response. See [here](https://github.com/elastic/elasticsearch/blob/2.4/core/src/main/java/org/elasticsearch/action/admin/indices/analyze/AnalyzeResponse.java#L160). Can we do that?

huydx:
Nice idea, two embedded structures seems fair.

huydx:
@olivere fixed as your advice

shouichi:
@olivere Could you take a look at this?

olivere:
I'm sorry, I took some time off between christmas and new year. I'll be back next week, and hopefully get this merged soon. Thank you for your work on Elastic.

olivere:
I recently added the Index Analyze API based on your original PR. I changed some things, e.g. 5.1.1 deprecated passing the parameters in the request (it wants them in the body instead), and did some formatting and documentation.

I will port this to elastic.v5 soon, and it will be part of the next releases of v3 and v5.

Thanks for being so patient, and thanks for your work on this PR.

PR #318, opened 2016-07-06
Support _nodes/stats for elasticsearch 2.x series

jdelgad:
Implemented against REST API published at
https://github.com/elastic/elasticsearch/blob/2.4/rest-api-spec/src/main/resources/rest-api-spec/api/nodes.stats.json

### DIFF
--- a/client.go
+++ b/client.go
@@ -1511,6 +1511,11 @@ func (c *Client) NodesInfo() *NodesInfoService {
 	return NewNodesInfoService(c)
 }
 
+// NodesStats retrieves one or more or all of the nodes stats information.
+func (c *Client) NodesStats() *NodesStatsService {
+	return NewNodesStatsService(c)
+}
+
 // TasksCancel cancels tasks running on the specified nodes.
 func (c *Client) TasksCancel() *TasksCancelService {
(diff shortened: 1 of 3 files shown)


jdelgad:
There were a few JSON result changes from 1.x series, but the API to access is consistent across both of my PRs.

olivere:
I guess we're having the same issues here as in #314.

olivere:
Sorry for the delay. I had to implement this myself to get a feel for the API. Nodes Info and Nodes Stats API only share their response structure; the actual contents are very different.

Could you take a look at [1c7b18e](https://github.com/olivere/elastic/commit/1c7b18ef09ff584e43fcecfcd29632de37c6d8a7) and see if it meets your expectations? If it does, I will port it over to `elastic.v2` (#314) and publish in the next release.

Thank you.

jdelgad:
Just tried it out and besides a few different changes to the API it works great! Consistency in the library is the way to go, so if you are comfortable with your change we'll accommodate.

👍

olivere:
It is available now for `elastic.v3` in 3.0.50+. Thanks for being a contributor to the project (and being so patient :-)).

PR #322, opened 2016-07-08
termvectors: add Score in TermsInfo

suzuken:
Enable to get `score` value in `terms`.

See detail: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-termvectors.html


olivere:
The missing `score` field will be available in all versions of Elastic in the next release, published later today.

Thanks for contributing.

PR #330, opened 2016-07-26
Add _source filtering support to scroll service

jrots:
Was something I needed .. so I've added it. Could be that I needed to wrap it completely in a source context struct.. but I only needed the _source part.

### DIFF
--- a/scroll.go
+++ b/scroll.go
@@ -14,14 +14,15 @@ import (
 
 // ScrollService manages a cursor through documents in Elasticsearch.
 type ScrollService struct {
-	client    *Client
-	indices   []string
-	types     []string
-	keepAlive string
-	query     Query
-	size      *int
-	pretty    bool
-	scrollId  string
+	client             *Client
+	indices            []string
+	types              []string
+	keepAlive          string
+	query              Query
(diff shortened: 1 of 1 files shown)


jrots:
Ok probably ignore this as you've mentioned here to use Scan service only: https://github.com/olivere/elastic/pull/245 ..
just confusing naming as elasticsearch tells us that scan requests are deprecated, 
so you immediately look for "scroll" things ;)

olivere:
Yeah, I mixed things up on the way: Scroll vs Scan is confusing. But now they're out, I cannot remove them without introducing a breaking change. Breaking changes in Elastic are for major versions only. I hopefully find the time to use a new Iterator-like service for 5.0 (to replace both), and will backport it into v3 ...

Thanks for contributing.

olivere:
BTW: `ScrollService` has been refactored in 3.0.48. You should now be able to do what you were after. Comment if something's missing, and I'll re-open the issue.

Also see [this example in the Wiki](https://github.com/olivere/elastic/wiki/ScrollParallel) to illustrate scrolling in parallel.

Issue #333, opened 2016-07-28
problem with from and size


jxstanford:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[ X ] elastic.v3 (for Elasticsearch 2.x)

### Please describe the expected behavior
When From = 0 and limit = 3, I would expect 13records returned

### Please describe the actual behavior
I have 8 records in my test index, and all 8 are returned

### Any steps to reproduce the behavior?
```
body := "{\"query\":{\"match_all\":{}}}"
sr, err := s.db.Search(indices).Source(body).
	From(0).Size(3).
	ExpandWildcards("open").AllowNoIndices(false).Do()
```

Any guidance appreciated!

olivere:
When you're passing the body as a string, you opt out of the DSL. You either construct the body yourself or you choose the DSL to do that for you. You can't mix both.

So either:

```go
sr, err := s.db.Search(indices).Source(elastic.NewMatchAllQuery()).From(0).Size(3).Do()
```

or

```go
sr, err := s.db.Search(indices).Source(`{"query":{"match_all":{}},"from":0,"size".3"}`).Do()
```

PR #334, opened 2016-07-29
Adds CustomQuery type 

jxstanford:
This is related to #333.  A simple wrapper around string that implements the Source method so it can be treated as a Query.

### DIFF
--- a/search_queries_custom.go
+++ b/search_queries_custom.go
@@ -6,11 +6,6 @@ package elastic
 
 import "encoding/json"
 
-// CustomQuery is a user provided query
-//
-// For more details, see
-// https://www.elastic.co/guide/en/elasticsearch/reference/master/query-dsl-custom-query.html
-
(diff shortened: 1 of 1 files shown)


olivere:
That should do the trick, yes. There's no `CustomQuery` in ES (and therefore the link to the documentation is 404ing). But overall that's quite useful. I will integrate it into one of the next releases.

Thanks for contributing.

jxstanford:
Great.  Fixed (removed) the broken doc link.  Thanks for paving the way!

olivere:
I'm merging this and will publish a new version just now. However, I changed this to `RawStringQuery` because I think it's more precise (see [6e5931d8](https://github.com/olivere/elastic/commit/6e5931d899c31cf0eb69d9b67274cf34cf2cff31)).

olivere:
Watch out. You probably need to change your `RawStringQuery` code: See #362.

PR #336, opened 2016-08-08
use github repo.

jinleileiking:


olivere:
No. That's wrong. Use `gopkg.in/olivere/elastic.v3` if you want to use the stable version of Elastic for ES 2.x.

PR #342, opened 2016-08-20
Fix to use healthcheckInterval

ww24:
Why don't you use `c.healthcheckInterval`?
It isn't used anywhere.


olivere:
You're correct. Thanks for the fix. I will publish a new version soon.

PR #344, opened 2016-08-23
add sort function for scroll search  API.

lixl1019:
When I'm using scroll search , I found that it does not support sorting, so I added sorting for it。

### DIFF
--- a/scroll.go
+++ b/scroll.go
@@ -19,6 +19,7 @@ type ScrollService struct {
 	types     []string
 	keepAlive string
 	query     Query
+	sorts     []SortInfo
 	size      *int
 	pretty    bool
 	scrollId  string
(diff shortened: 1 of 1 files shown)


olivere:
Thanks for your work. I'm a bit slow recently with regard to responding to issues. Sorry for that.

I'm in the process of rewriting the Scroll service from scratch for ES version 5. And I think I'll backport it to ES v3 and probably even ES v2. So I'm a bit hesistant of changing things now.

However, if you have a chance to use the [Scan service](https://github.com/olivere/elastic/blob/release-branch.v3/scan.go) ([tests](https://github.com/olivere/elastic/blob/release-branch.v3/scan_test.go)), you'd get sorting for free, plus queries and some other things.

The situation around Scan and Scroll is confusing, I'm sorry. It's a legacy problem. But if you can, use Scan.

olivere:
BTW: `ScrollService` has been refactored in 3.0.48. It now has sorting, just as `Scan` has. See [this example in the Wiki](https://github.com/olivere/elastic/wiki/ScrollParallel) to illustrate scrolling in parallel.

PR #363, opened 2016-09-22
Compute BulkService size in bytes only when asked

LotharSee:
Do not compute/update the `sizeInBytes` in bytes all the time, do it only if we ask for `EstimatedSizeInBytes`. 
It makes `BulkService` much faster for the common use-case (when `EstimatedSizeInBytes` is not used), but also at least as fast when `EstimatedSizeInBytes` is used (with cached size computation).

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -36,7 +36,9 @@ type BulkService struct {
 	refresh  *bool
 	pretty   bool
 
-	sizeInBytes int64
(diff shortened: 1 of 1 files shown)


olivere:
Looking good. It will be available in 3.0.53+ and the current development version 5.0.0 (for Elasticsearch 5.x) soon.

Thanks for being a contributor!

Issue #353, opened 2016-09-08
Add `request_cache` to search


olivere:
ES 2.x added `request_cache` to REST API spec for Search API.

timeemit:
Hey @olivere -- Just tried to use this feature and am seeing 

```
SearchParseException[failed to parse search source [{"request_cache":false}]]; nested: SearchParseException[failed to parse search source. unknown search element [request_cache]]
```

The [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html) explains that
> the request_cache must be passed as query-string parameters. The rest of the search request should be passed within the body itself.

PR #365, opened 2016-09-27
Search & SearchSource: Move `request_cache` from request body to param

timeemit:
Addresses the following error following up on issue #353 

```
SearchParseException[failed to parse search source[{"request_cache":false}]]; nested:
SearchParseException[failed to parse search source. unknown search element [request_cache]]
```

The [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html) explains that
> the request_cache must be passed as query-string parameters. The rest
> of the search request should be passed within the body itself.

### DIFF
--- a/search.go
+++ b/search.go
@@ -314,6 +314,9 @@ func (s *SearchService) buildURL() (string, url.Values, error) {
 	if s.ignoreUnavailable != nil {
 		params.Set("ignore_unavailable", fmt.Sprintf("%v", *s.ignoreUnavailable))
 	}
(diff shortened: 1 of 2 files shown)


olivere:
I think the correct solution is to move `request_cache` out of `SearchSource` because, according to documentation, it's a request-level setting that must be passed in the query parameters.

So I'll add a `RequestCache` setter into `SearchService` and remove the `RequestCache` setter from `SearchSource`.

Thanks for pointing this out.

olivere:
Just released 3.0.54.

Issue #366, opened 2016-09-28
Cannot use Scan or Scroll with raw request body


akotlar:
### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[X ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v5 (for Elasticsearch 5.x)

### Please describe the expected behavior
We should be able to configure the body of the request from a valid JSON string, or at least a valid map[string]interface{}

### Please describe the actual behavior
There is currently no way to do this.

### Any steps to reproduce the behavior?
This is a fundamental API choice. To reproduce it, try to find a way to search from raw request body. There is no API method for this. SearchService has one, using a workaround that relies on a "source" attribute, which is distinct from its "searchSource" attribute, and is solely used to allow people to work around the SearchSource limitation.

I've spent enough time with the source code to understand that several core modules force the use of the DSL, such as Sort, SearchSource, etc. There is no easy way to rework things to truly support raw queries. From my perspective, this is a tremendous opportunity to improve the utility of this library. The DSL should be completely optional.

I've begun updating the library, and will invest the necessary time if @olivere supports this idea.

Feedback on the proposal is welcome!

olivere:
Without knowing, you are actually touching quite a few different topics in this issue. First of all, I'm not negative on your proposal. I'm just asking to think through the choices before blaming the library of not supporting something that should be "the most basic thing that actually has to work."

First of all, the vast majority of the code in Elastic is due to the DSL. This is where I started, and this is still its core functionality. There were similar requests asking for Elastic to work as a kind of proxy for the responses from Elasticsearch. You may not like it, but this is not its purpose.

Second, if you ask for a feature that accepts something serializable as the body of a request to send it to Elasticsearch: Why not use `net/http` directly? Seriously. It's [not that hard](https://gist.github.com/olivere/478060604604514c163a5bed5cd2339c). If you want to scroll and pass just the query (and set other stuff via the DSL): [You can do that too](https://gist.github.com/olivere/eb0cea18708fdeb9ce04783a04378974).

[...]

PR #367, opened 2016-09-29
Scan raw query

akotlar:
This partially addresses #366. I left one TODO, line 117, handling of the errors.

This pull request follows the convention set in SearchService, setting a "source" attribute that handles the body of the request. We can of course rename this to body, and the method to Body().

Test provided.

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -5,13 +5,13 @@
 package elastic
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"gopkg.in/olivere/elastic.v3/uritemplates"
 	"io"
 	"net/url"
 	"strings"
-
-	"gopkg.in/olivere/elastic.v3/uritemplates"
 )
 
 const (
@@ -33,6 +33,7 @@ type ScanService struct {
 	types        []string
 	keepAlive    string
 	searchSource *SearchSource
+	source       map[string]interface{}
 	pretty       bool
 	routing      string
 	preference   string
@@ -94,6 +95,26 @@ func (s *ScanService) SearchSource(searchSource *SearchSource) *ScanService {
 	if s.searchSource == nil {
 		s.searchSource = NewSearchSource().Query(NewMatchAllQuery())
 	}
+	return s
+}
+
+// Source allows the user to set the request body manually without using
+// any of the structs and interfaces in Elastic.
+func (s *ScanService) Source(sourceToSet interface{}) *ScanService {
+	switch sourceToSet.(type) {
+	case string:
+		err := json.Unmarshal([]byte(sourceToSet.(string)), &s.source)
+
+		if err != nil {
+			panic(err)
+		}
+	case map[string]interface{}:
+		s.source = sourceToSet.(map[string]interface{})
+	default:
+		panic(errors.New("Source requires string or map"))
+	}
+	// TODO: We could decide what to do with errors, could return err
+	// but then couldn't change
 	return s
 }
 
@@ -221,12 +242,29 @@ func (s *ScanService) Do() (*ScanCursor, error) {
 		path += "/" + strings.Join(typesPart, ",")
 	}
 
+	// Define the body of the request
+	var body map[string]interface{}
+
+	if s.source != nil {
+		body = s.source
+	} else {
+		src, err := s.searchSource.Source()
+
+		if err != nil {
+			return nil, err
+		}
+
+		// searchSource.Source() actually returns a map, yet its return type
+		// is specified as interface...
+		body = src.(map[string]interface{})
+	}
+
 	// Search
 	path += "/_search"
 
 	// Parameters
 	params := make(url.Values)
-	if !s.searchSource.hasSort() {
+	if _, ok := body["sort"]; !ok {
 		// TODO: ES 2.1 deprecates search_type=scan. See https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_21_search_changes.html#_literal_search_type_scan_literal_deprecated.
 		params.Set("search_type", "scan")
 	}
@@ -246,10 +284,7 @@ func (s *ScanService) Do() (*ScanCursor, error) {
 	}
 
 	// Get response
-	body, err := s.searchSource.Source()
-	if err != nil {
-		return nil, err
-	}
+	// Perform request
 	res, err := s.client.PerformRequest("POST", path, params, body)
 	if err != nil {
 		return nil, err
@@ -257,6 +292,7 @@ func (s *ScanService) Do() (*ScanCursor, error) {
 
 	// Return result
 	searchResult := new(SearchResult)
+
 	if err := s.client.decoder.Decode(res.Body, searchResult); err != nil {
 		return nil, err
 	}
--- a/scan_test.go
+++ b/scan_test.go
@@ -282,7 +282,187 @@ func TestScanWithSearchSource(t *testing.T) {
 	src := NewSearchSource().
 		Query(NewTermQuery("user", "olivere")).
 		FetchSourceContext(NewFetchSourceContext(true).Include("retweets"))
+
 	cursor, err := client.Scan(testIndexName).SearchSource(src).Size(1).Do()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if cursor.Results == nil {
+		t.Fatalf("expected results != nil; got nil")
+	}
+	if cursor.Results.Hits == nil {
+		t.Fatalf("expected results.Hits != nil; got nil")
+	}
+	if cursor.Results.Hits.TotalHits != 2 {
+		t.Fatalf("expected results.Hits.TotalHits = %d; got %d", 2, cursor.Results.Hits.TotalHits)
+	}
+
+	numDocs := 0
+	pages := 0
+
+	for {
+		searchResult, err := cursor.Next()
+		if err == EOS {
+			break
+		}
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		pages += 1
+
+		for _, hit := range searchResult.Hits.Hits {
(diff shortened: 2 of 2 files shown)


scan.go
```diff
@@ -33,6 +33,7 @@ type ScanService struct {
 	types        []string
 	keepAlive    string
 	searchSource *SearchSource
+	source       map[string]interface{}
```
olivere: The `source` should be more flexible. It should be an `interface{}` and we should try our best to make it the body of the request. It should accept at least `io.Reader` and `string` and fall back to serialize as JSON (as we'd do with `map[string]interface{}`).

scan.go
```diff
@@ -97,6 +98,26 @@ func (s *ScanService) SearchSource(searchSource *SearchSource) *ScanService {
 	return s
 }
 
+// Source allows the user to set the request body manually without using
+// any of the structs and interfaces in Elastic.
+func (s *ScanService) Source(sourceToSet interface{}) *ScanService {
```
olivere: I guess we could go with `Source` but I changed that to `BodyString(...)`  and `BodyJson(...)` in other services to be more precise and to relief confusion about the `Source` that sets the body and the `Source` that returns the JSON structure of a DSL object.

But I'm okay with `Source` here because `Scan` will be dead with v5.

scan.go
```diff
@@ -221,12 +242,29 @@ func (s *ScanService) Do() (*ScanCursor, error) {
...
+		body = src.(map[string]interface{})
+	}
+
 	// Search
 	path += "/_search"
 
 	// Parameters
 	params := make(url.Values)
-	if !s.searchSource.hasSort() {
+	if _, ok := body["sort"]; !ok {
```
olivere: And things like these are the core of my problem with your approach: You pick one thing (e.g. `sort`) and fix that. But you open up a can of worms. In the near future I'll get an issue that says something like: "I have set the body of the request via `Source` and now I only want to change this or that with the DSL. Why the heck can't I do this?" That's why I chose to make it one-way. Use source? You're on your own.

Nevermind. Back to the review...

We should simply remove the `search_type=scan` here. It's deprecated and users should simply [specify `"sort":["_doc"]`](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/breaking_21_search_changes.html#_literal_search_type_scan_literal_deprecated) when using `Source` (we could still do that automatically when users chose the DSL via `SearchSource`, but maybe ES already does that automatically--will have to check my sources). Also, you'd no longer be required to do the type conversion mentioned earlier.

scan.go
```diff
@@ -221,12 +242,29 @@ func (s *ScanService) Do() (*ScanCursor, error) {
...
+		body = s.source
+	} else {
+		src, err := s.searchSource.Source()
+
+		if err != nil {
+			return nil, err
+		}
+
+		// searchSource.Source() actually returns a map, yet its return type
+		// is specified as interface...
```
olivere: Yes. That's on purpose. It makes it easy to change it to `[]byte`or `*json.RawMessage` whenever I want. Then your type conversion will make it crash.

scan.go
```diff
@@ -5,13 +5,13 @@
 package elastic
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"gopkg.in/olivere/elastic.v3/uritemplates"
```
olivere: Please keep the import statements intact.

1. stdlib imports
2. Non-stdlib imports
3. Application imports

scan.go
```diff
@@ -97,6 +98,26 @@ func (s *ScanService) SearchSource(searchSource *SearchSource) *ScanService {
 	return s
 }
 
+// Source allows the user to set the request body manually without using
+// any of the structs and interfaces in Elastic.
+func (s *ScanService) Source(sourceToSet interface{}) *ScanService {
+	switch sourceToSet.(type) {
```
olivere: You shouldn't to the serialization part here. This is done later when the request is sent via `PerformRequest`: See [`Request.SetBody(...)`](https://github.com/olivere/elastic/blob/release-branch.v3/request.go#L38). Simply save `source` and use it in `Do(...)` if it's non-nil.

scan.go
akotlar: ok... Source() is a function, with a defined output, which is actually a map, but defined as a less general interface. The problem with saying "my type conversion will crash" is that, unless the API is unstable, or I'm not actually meant to use the output of Source(), I should be able to use the data that the function outputs. If we take the concept you suggest to the extreme, why not just define everything as an interface{} and make any usable output depend on an explicit type check?

scan.go
akotlar: It's not my approach, it's yours. You chose to make a check for something that relies on the DSL. The change I made above is the easiest way not to modify the behavior of ScanService when it is used without setting ScanService.Source()...

If I use Source I'm **not** on my own. There's no way to use Source on it's own because params depended on the DSL.

Agreed, remove the scan type.

scan.go
akotlar: Ok, as mentioned this is following the convention set in SearchSource. I agree with Body(); the function should take either a valid JSON string, or a valid JSON map.

scan.go
olivere: `SearchSource.Source()` does not return a `map[string]interface{}`. [It returns an `interface{}`](https://github.com/olivere/elastic/blob/release-branch.v3/search_source.go#L295). You depend on the internals of that function, which can change. All that `Source()` promises is that you can serialize it to JSON, which you can do with a variety of types. The [`Query`](https://github.com/olivere/elastic/blob/release-branch.v3/query.go) establishes the same contract BTW.

Please provide an example of a function signature in Go that specifies the following contract: This function either returns `io.Reader`, `string`, `[]byte`, `*json.RawMessage`, or `map[string]interface{}`.

scan.go
akotlar: Got it

scan.go
olivere: When I say "you're on your own", I mean that once you do that, you cannot use the DSL to change individual settings. Do we agree on that?

scan.go
olivere: Yes, correct.

scan.go
akotlar: It's true, I'm relying on the internals of the function, which creates and returns a map (the (type) of that interface). The fact that you define it as an interface says nothing to the user as to the purpose of the function. It returns data, I should be able to understand what data it returns, and use that however I want, or the data shouldn't be returned (imo).

If the intention is not to make the output directly usable, I suggest not making the return value public

scan.go
akotlar: Yes, I think we're mostly saying the same thing: If I'm using Source, there should be no forced internal defeat of the data I intended to set in source, and I also shouldn't expect that the DSL will modify anything I set via Source.

I think the change you suggested makes the most sense, just remove the hasSort check.

scan.go
olivere: I explicitly stated that in the documentation of interfaces, e.g. [`Query`](https://github.com/olivere/elastic/blob/release-branch.v3/query.go). I know no way of enforcing it with the compiler.

I agree that documentation could be better throughout the library.

You're correct in that I should have hidden many things; unfortunately, it's about 3 years too late for that.

akotlar: As an example, json.Unmarshal returns a very defined set of possible values in interface, bool, float64, str, nil, and []interface{} (an array of the aforementioned types or a map or an array). I can understand the use of interface{} there, because more than one type can be returned.

akotlar: No worries! I missed the documentation bit. If that's documented, it's clear that I misunderstood.

scan.go
olivere: Just as a reference: The `_.hasSort()` check was in there as a performance hack. It made scrolling much faster because it forced ES to not apply any sorting logic. ES is more clever now and will automatically do the right thing, hence the deprecation of the `search_type=scan` in 2.1.

olivere:
To make a long story short, I further investigated the issue and have added a `Body` setter to `ScanService` and `ScrollService`. Can you please review the [set-body-for-scroll](https://github.com/olivere/elastic/tree/set-body-for-scroll) branch?

Upon further review of the necessary changes, I noticed that we cannot simply remove the `search_type=scan` and keep compatibility with older clients. ES behaves differently whether you set `search_type=scan` or not. Notice how it returns a result on the first call when you remove it, and it returns an empty result set when you add it.

With `search_type=scan`:

```
$ curl -is -XPOST 'localhost:9200/elastic-test/_search?scroll=5m&search_type=scan&size=1' -d '{"query":{"match_all":{}}}'
HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
Content-Length: 205

{"_scroll_id":"...",...,"hits":{"total":3,"max_score":0.0,"hits":[]}}
```

Without `search_type=scan`:

```
$ curl -is -XPOST 'localhost:9200/elastic-test/_search?scroll=5m&size=1' -d '{"query":{"match_all":{}}}'
HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
Content-Length: 412

{"_scroll_id":"...",...,"hits":{"total":3,"max_score":1.0,"hits":[{"_index":"elastic-test","_type":"tweet","_id":"1","_score":1.0,"_timestamp":1475227710915,"_source":{"user":"olivere","message":"Welcome to Golang and Elasticsearch.","retweets":0,"created":"0001-01-01T00:00:00Z"}}]}}
```

Conclusion: Scan must die. Use Scroll. Which is really the opposite of what I've recommended for a long time... _sigh_

akotlar:
Sounds good. I took a quick look, looks like what we need. How can I help get this out faster?

Thanks for the demonstration of scan vs scroll.

scan.go
akotlar: That makes sense! Thanks

olivere:
3.0.55 has been published.

PR #373, opened 2016-10-09
make the shard information returned by searches available

kylebrandt:
I would like to be able to access shard information that is returned by search queries. 

My particular use case is that I would like to know when 0 shards were hit when `IgnoreUnavailable` is `true`.


olivere:
Yep, thanks for being a contributor. I hope I'll find the time to merge your PR over the weekend.

olivere:
Will soon be available, in 3.0.56.

PR #376, opened 2016-10-11
Add authorize header option to elastic client

dylannz:
Not sure if this is the best way of implementing, but would be nice to have something like this.


dylannz:
sorry, ignore this, wrong base

PR #377, opened 2016-10-13
elasticsearch docs in comments for 2.x are leading to 404 page, update to avoid broken references

thomasmodeneis:
Hi,
Simple change just to help people finding the right reference documentation.

Actual docs refer to broken 404 page:

```
404 Well, That about Wraps It Up for This Page
`I refuse to prove that I exist,’ says Page, `for proof denies faith, and without faith I am nothing.’

`But,’ says Man, `The Babel fish is a dead giveaway, isn’t it? It could not have evolved by chance. It proves you exist, and so therefore, by your own arguments, you don’t. QED.’

`Oh dear,’ says Page, `I hadn’t thought of that,’ and promptly disappears in a puff of logic.
```

Cheers.

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -24,7 +24,7 @@ import (
 // reuse BulkService to send many batches. You do not have to create a new
 // BulkService for each batch.
 //
-// See https://www.elastic.co/guide/en/elasticsearch/reference/2.x/docs-bulk.html
+// See https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html
 // for more details.
 type BulkService struct {
 	client *Client
--- a/bulk_update_request.go
+++ b/bulk_update_request.go
@@ -79,8 +79,8 @@ func (r *BulkUpdateRequest) Parent(parent string) *BulkUpdateRequest {
 }
 
 // Script specifies an update script.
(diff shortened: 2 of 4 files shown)


olivere:
Yeah, documentation is hard. However, using `current` instead of `2.x` is also wrong in that it fails when they decide to remove a feature in a newer version, or restructure their website. I had thought that `2.x` was a permalink to the "current documentation of 2.x", which it isn't as you found out.

I've asked the Elastic people on what's the right strategy for external documentation.

olivere:
I learned from [issue #20907](https://github.com/elastic/elasticsearch/issues/20907) that it's wrong to link to something like `2.x`. Also, there is no symlink to the documentation of the current minor version of Elasticsearch, i.e. there is no permalink to a topic in `2.<current-minor-version>`. 

So we have to use links that point to e.g. `2.3`, `2.4` etc. and update them every now and then.

**Wrong**:

```
https://www.elastic.co/guide/en/elasticsearch/reference/2.x/docs-bulk.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html
```

**Right**:

```
https://www.elastic.co/guide/en/elasticsearch/reference/2.4/docs-bulk.html
```

(and eventually change to 2.5, 2.6 etc.)

olivere:
In 5.0.24+ I tried (at least) to fix all documentation links.

PR #395, opened 2016-11-17
Avoid //_alias request in aliases service

EugeneEgorov:
This is fixed in v5 branch, just backported here.
//_alias doesn't work with AWS elasticsearch service with IAM authentification(it returns 403 for double slash)


olivere:
Thanks. It's in 3.0.57.

Issue #401, opened 2016-11-23
Enabling no-op with bulk updates


gm42:
### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[x] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v5 (for Elasticsearch 5.x)

### Please describe the expected behavior
I am using `BulkIndexRequest` and would like to explicitly set `DetectNoop` to `true`.

### Please describe the actual behavior

There does not seem to be such method available.

Similarly to https://github.com/elastic/elasticsearch/issues/10881, I assume the feature is available in Elasticsearch 2.x but there is no client option to enable it?

olivere:
Yep. That's missing. I'll add it in the next release. Thanks for reporting.

gm42:
@olivere many thanks. As far as I understood by default it's not enabled, hence I was looking around for it.

gm42:
I made a (temptative) PR, however I notice that Elasticsearch maybe supports this only with **update** requests, I am not sure about **index** requests as per my initial post.

gm42:
So, I dug a bit into this. The `detect_noop` is enabled by default for **update** bulk requests (making PR #403 somewhat of a whistle&bell). For **index** bulk requests, as I was asking originally, it's a different situation, see https://github.com/elastic/elasticsearch/issues/12969

Basically from the documentation update they did there (https://github.com/nik9000/elasticsearch/blob/master/docs/reference/docs/index_.asciidoc#noop-updates), it's clear that index API has no option for this :cry:

olivere:
DetectNoop is available in 3.0.59 and 5.0.12.

PR #403, opened 2016-11-24
Implement DetectNoop for BulkUpdateRequest

gm42:
Fixes https://github.com/olivere/elastic/issues/401

(needs some tests too)


olivere:
Available in 3.0.59 and 5.0.12.

PR #408, opened 2016-11-30
fix suggest context bug

coseyo:
fixed in other branch but without this one.

### DIFF
--- a/suggest_field.go
+++ b/suggest_field.go
@@ -6,6 +6,7 @@ package elastic
 
 import (
 	"encoding/json"
+	"errors"
 )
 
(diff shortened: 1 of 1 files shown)


olivere:
Thank you. I will have to look up if this has changed in the same way in ES 2.x as it did in ES 5.x (sometimes documentation != Java source). If it does, I will merge in the next release.

coseyo:
OK. I used the 2.4, and it is good for me.

olivere:
I fixed this in the recent versions of both v3 and v5, which I'll publish later today.

Thanks for your support.

PR #416, opened 2016-12-07
Added example api using this library

kamorahul:
I was working on this library from last one month and build this API using this library. I think it will be helpful for the people want to utilize it.


olivere:
I added an [Examples section](https://github.com/olivere/elastic/wiki/Examples) to the Wiki.

PR #427, opened 2016-12-29
Add SetRawSource to ScanService

aliphen:
SearchService has a Source(query interface{}) method, useful for
performing direct requests crafted somewhere else in the code.
Having this available in ScanService is similarly useful.

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -32,6 +32,7 @@ type ScanService struct {
 	indices      []string
 	types        []string
 	keepAlive    string
+	source       interface{}
 	searchSource *SearchSource
 	pretty       bool
 	routing      string
@@ -112,6 +113,11 @@ func (s *ScanService) SearchSource(searchSource *SearchSource) *ScanService {
 	if s.searchSource == nil {
 		s.searchSource = NewSearchSource().Query(NewMatchAllQuery())
 	}
+	return s
+}
+
+func (s *ScanService) SetRawSource(source interface{}) *ScanService {
+	s.source = source
 	return s
 }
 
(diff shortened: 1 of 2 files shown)


olivere:
Thanks for contributing.

I will have to see if that makes sense for `Scan`, because `Scan` and `Scroll` in v2 really need access to the body to do what they are supposed to do.

BTW: I've switched to using `Body` as the function name to do that in other services, e.g. [`ScrollService`](https://github.com/olivere/elastic/blob/release-branch.v5/scroll.go#L93-L100) in 5.0. Then it won't collide with e.g. this [`Source`](https://github.com/olivere/elastic/blob/release-branch.v5/query.go).

aliphen:
Thanks for the review ! I do think it makes sense, it just means that it's really up to the user to provide a valid body. The interface name change is a good idea IMO.

Cheers !

aliphen:
Hi,
just for planning purposes, could you give me a very rough estimate of when you expect to merge (or close) this ? No hurry, I just don't know how often you have time to work on this so I figured I'd ask.
Thanks for your work on the lib !

olivere:
Hi there. I've not merged this on the weekend because it seems you're trying to get this into v2 which is in maintenance mode really. Is it correct you still need support for ES 1.x? It's very very stable, no problem, just asking.

I can try to get this in later in the evening. It's not too much work after all.

olivere:
The Travis build for 2.0.56 is running. Once it succeeds, I'll tag the release and you should be able to `go get -u gopkg.in/olivere/elastic.v2`.

aliphen:
Thank you for the quick answer and the merge !

PR #428, opened 2017-01-04
Update doc.go with context parameter for Do()

timbutler:
Just a minor change, I missed that context had been added to newer versions and can see that examples etc have already been updated.


olivere:
Oh, yeah. In v5, you're supposed to pass a context for the `ctxhttp` we're using in making HTTP requests. Thanks.

PR #430, opened 2017-01-05
added wait_for_no_relocating_shards call

bmconklin:
Version 5.0+ has depricated the Cluster Health Check parameter `wait_for_relocating_shards` (which takes a number of shards to wait for) and has replaced it with `wait_for_no_relocating_shards` (which takes a true/false value). See documentation: https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html

In v5.0+, using the deprecated parameter returns the response:
> Error 400 (Bad Request): wait_for_relocating_shards has been removed, use wait_for_no_relocating_shards [true/false] instead [type=illegal_argument_exception]

This change documents that the old parameter has been removed and adds the new parameter. I chose to not fully remove the old parameter so that if someone should use it mistakenly after updating to 5.0+ they will still get the error message that alerts them that it has been removed, and they will not assume that it continued to work as intended.

### DIFF
--- a/cluster_health.go
+++ b/cluster_health.go
@@ -19,17 +19,18 @@ import (
 // See http://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html
 // for details.
 type ClusterHealthService struct {
-	client                  *Client
-	pretty                  bool
-	indices                 []string
-	level                   string
-	local                   *bool
-	masterTimeout           string
-	timeout                 string
-	waitForActiveShards     *int
-	waitForNodes            string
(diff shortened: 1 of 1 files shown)


olivere:
Thanks for your support. It's hard to keep on top of all the changes.

If it's removed in 5.x, we should really also remove the old `wait_for_relocating_shards` setting. Can you do that?

bmconklin:
Sure, I went ahead and just removed it completely.

olivere:
I have committed your change to v5. Thanks for finding this, Bryan.

PR #422, opened 2016-12-14
Feature: Pipeline Stats Bucket

mcos:
This PR adds the operations for a Pipeline Stats Bucket aggregation, which is documented at https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-pipeline-stats-bucket-aggregation.html.

I'm creating this PR into the `release-branch.v3` as I work with Elasticsearch `v2.3`. If this is not an appropriate PR for this branch, I'm happy to close it. The main impetus for this was having to avoid the necessity of creating several different aggregations for `AvgBucket`, `MinBucket`, `MaxBucket` and `SumBucket`.

If this is worth merging, I can follow up with subsequent PRs for the `ExtendedStatsBucket` and `PercentilesBucket` pipeline aggregations.

PR #432, opened 2017-01-10
Bugfix: AggregationPipelineStatsBucket.Count should be int64, not *float64

mcos:
I noticed a problem with #422, when I saw that `Count` in [AggregationStatsMetric](https://github.com/olivere/elastic/blob/c39b96138de69a25630c9159ccc0393cab734861/search_aggs.go#L659) is an `int64`. `AggregationPipelineStatsMetric` should also have an `int64` `Count` property, for consistency.


olivere:
Fixed in v3 and v5.

mcos:
Thanks @olivere

PR #435, opened 2017-01-18
Replace reverse with order in ScriptSort

kaporzhu:
ref: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html#_sort_order


olivere:
Fixed in v5.0.22.

ES 2.x still uses the "reverse" flag… see [here](https://github.com/elastic/elasticsearch/blob/2.4/core/src/main/java/org/elasticsearch/search/sort/ScoreSortBuilder.java#L53).

Issue #439, opened 2017-01-20
can't use function_score query with filter


mosasiru:
### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[x] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v5 (for Elasticsearch 5.x)

### Please describe the expected behavior

I want to use function_score query with just filter query like this.
https://www.elastic.co/guide/en/elasticsearch/guide/current/function-score-filters.html

```
GET /_search
{
  "query": {
    "function_score": {
      "filter": { 
        "term": { "city": "Barcelona" }
      },
      "functions": [ 
        {
          "filter": { "term": { "features": "wifi" }}, 
          "weight": 1
        },
        {
          "filter": { "term": { "features": "garden" }}, 
          "weight": 1
        },
        {
          "filter": { "term": { "features": "pool" }}, 
          "weight": 2 
        }
      ],
      "score_mode": "sum", 
    }
  }
}
```

### Please describe the actual behavior

But elastic.v3 `FunctionScoreQuery` seems filter query is not acceptable.

olivere:
Fixed in v3.0.64.

PR #442, opened 2017-01-23
add filter query to function_score

mosasiru:
I fixed  https://github.com/olivere/elastic/issues/439.

ES2.x Function Score query accept query and filter.
https://www.elastic.co/guide/en/elasticsearch/guide/current/function-score-filters.html


olivere:
Yes. Elastic v5 [already has the filter](https://github.com/olivere/elastic/blob/release-branch.v5/search_queries_fsq.go#L42-L47) while elastic.v3 was still missing it. Thanks. I hope to do an update next weekend, with this PR included.

mosasiru:
thank you!

olivere:
Fixed in v3.0.64.

PR #454, opened 2017-02-06
Add scripted_upsert to bulk_update_request

salajlan:
its the same as on update, just add it to bulk update request.

work on bulk as mention [elasticsearch issue #19413](https://github.com/elastic/elasticsearch/issues/19413)


olivere:
Thanks for your support. The `scripted_upsert` flag is available in 5.0.24+ which will be released later today.

PR #460, opened 2017-02-14
Highlighting boundary chars

AdrianLungu:
It seems that the package currently generates the highlight boundary chars as an array:

```
{
   "boundary_chars":[
      32,
      9,
      10
   ],
   "fields":{
      "text":{}
   },
   "type":"fvh"
}
```

While ElasticSearch 5.1.1 wants a string:

```
{
   "boundary_chars":" \\t\\n",
   "fields":{
      "text":{}
   },
   "type":"fvh"
}
```

Otherwise throwing an error:

> elastic: Error 400 (Bad Request): [highlight] boundary_chars doesn't support values of type: START_ARRAY [type=illegal_argument_exception]

### DIFF
--- a/highlight.go
+++ b/highlight.go
@@ -19,7 +19,7 @@ type Highlight struct {
 	encoder               *string
 	requireFieldMatch     *bool
 	boundaryMaxScan       *int
-	boundaryChars         []rune
+	boundaryChars         *string
 	highlighterType       *string
 	fragmenter            *string
 	highlightQuery        Query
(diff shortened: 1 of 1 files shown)


olivere:
@AdrianLungu Thanks for your work on elastic. Sorry, I didn't have the time to look into this over the weekend. It's a larger change that I want to do correctly, so I have to check Java sources and see if there are other changes in that section of ES as well. I will do as soon as I have a few hours of spare time for Elastic.

olivere:
I finally found some time to review the latest changes. This change will be in 5.0.27 which is being tested on Travis as I type. Give it an hour or so and you'll be able to see your change. Thanks again.

PR #464, opened 2017-02-21
Updates example to provide string to refresh parameter.

jrmycanady:
For reasons I can't find in code/docs, v5 requires a string for `func (s *ExistsService) Refresh` compared to the bool in v4. This simple updates the example in the v5 branch to match the code requirements.

I am also curious why it has changed while the RealTime parameter has not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ _, err = client.Index().
     Type("tweet").
     Id("1").
     BodyJson(tweet).
-    Refresh(true).
+    Refresh("true").
     Do(ctx)
 if err != nil {
     // Handle error


olivere:
@jrmycanady Thanks for contributing to elastic. The change is due to [this breaking change in 5.0](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_50_document_api_changes.html#_literal_refresh_literal_no_longer_supports_truthy_and_falsy_values). I will review your changes and typically update in a batch over the weekends.

jrmycanady:
@olivere Sounds good.

PR #469, opened 2017-02-28
Multi Search ignore_unavailable option

initialcontext:
Hi there! My company relies on your excellent `olivere/elastic` Go binding, but for several of our uses we require the Multi Search API to expose the `ignore_unavailable` option. This PR adds that option, and a matching builder API call and unit test, to `SearchRequest`.

I submitted the patch applied to `release-branch.v5` but would really appreciate a backport to `release-branch.v3` if possible, should this PR be accepted. I have tested this code against `gopkg.in/olivere/elastic.v3` and it seems to work as expected.

 Thanks again for the great library!

### DIFF
--- a/search_request.go
+++ b/search_request.go
@@ -13,14 +13,15 @@ import (
 // query details (see SearchSource).
 // It is used in combination with MultiSearch.
 type SearchRequest struct {
-	searchType   string // default in ES is "query_then_fetch"
-	indices      []string
-	types        []string
-	routing      *string
-	preference   *string
-	requestCache *bool
-	scroll       string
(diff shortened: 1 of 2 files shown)


olivere:
I've added indices options to `SearchRequest` for both v3 and v5. However, the Travis builds are in an infinite loop, maybe due to the AWS issues today. All tests succeed locally, but I will delay tagging the releases and push to gopkg.in until CI succeeds.

olivere:
Published as 3.0.66 and 5.0.28.

initialcontext:
Thanks again for your fast response, the library is great!

PR #478, opened 2017-03-11
Eli/search profile api

initialcontext:
Just adds support for the [Profile API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-profile.html). Same deal as last time, was hoping to get this back ported to `olivere/elastic.v3` if you accept for `elastic.v5`. Tested and works with `elastic.v3` in my env. Thanks again!

### DIFF
--- a/search.go
+++ b/search.go
@@ -91,6 +91,13 @@ func (s *SearchService) Timeout(timeout string) *SearchService {
 	return s
 }
 
+// Profile sets the Profile API flag on the search source.
+// A search executed by this service will return query profiling data too.
+func (s *SearchService) Profile(should bool) *SearchService {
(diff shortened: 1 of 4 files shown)


olivere:
Thanks for contributing to Elastic again.

PR looks solid, and I'm quite happy to support the Profile API. Give me some more time to check with the Elasticsearch Java sources to see if we're missing something. I like to be in line with the Java source code.

initialcontext:
That sounds great, thanks again!

olivere:
Just a few tweaks, but it's ready to roll.

initialcontext:
Wonderful, thanks again for the quick response too!

PR #491, opened 2017-03-25
fix bulk update request bug

wangtuo:


olivere:
Thanks. Will be part of the 5.0.31 release.

Issue #493, opened 2017-03-27
Suggester Score Support


umpc:
I have tried this on [v5 stable](http://gopkg.in/olivere/elastic.v5) and [master](https://github.com/olivere/elastic/tree/13cdb2f8ba30d72761552bcd9ab2554728463a5b).

### Which version of Elastic are you using?

[*] elastic.v5 (for Elasticsearch 5.x)

### Please describe the expected behavior

The ```Score``` field of the ```elastic.suggestionOption``` is set to the option score from the HTTP response.

### Please describe the actual behavior

The ```Score``` field of the ```elastic.suggestionOption``` is always ```0```.

### Any steps to reproduce the behavior?

My query using this API:
```
phraseSuggester := elastic.NewPhraseSuggester("place_spelling_suggestions").
	Field("city").
	Text("gatlinberg")
	
searchSource := elastic.NewSearchSource().
	Suggester(phraseSuggester).
	FetchSource(false).
	TrackScores(true).
	Timeout("100ms")

searchResult, err := ElasticConn.Search().
	Index("places").
	Type("place").
	SearchSource(searchSource).
	Do(context.Background())

if err != nil {
	log.Println(err)
	return
}
```

The query that the API is sending to ElasticSearch:

```
POST /places/place/_search HTTP/1.1
Host: 172.18.0.3:9200
User-Agent: elastic/5.0.32 (linux-amd64)
Transfer-Encoding: chunked
Accept: application/json
Content-Type: application/json
Accept-Encoding: gzip

90
{"_source":false,"suggest":{"place_spelling_suggestions":{"text":"gatlinberg","phrase":{"field":"city"}}},"timeout":"100ms","track_scores":true}
```

When sending an identical query a different way to quickly read the response, the scores are visible:
```
HTTP/1.1 200 OK
content-type: application/json; charset=UTF-8
content-length: 306

{
  "took": 4,
  "timed_out": false,
  "_shards": {
    "total": 5,
    "successful": 5,
    "failed": 0
  },
  "hits": {
    "total": 0,
    "max_score": 0.0,
    "hits": []
  },
  "suggest": {
    "place_spelling_suggestions": [
      {
        "text": "gatlinberg",
        "offset": 0,
        "length": 10,
        "options": [
          {
            "text": "gatlinburg",
            "score": 0.0056934706
          },
          {
            "text": "galenberg",
            "score": 0.0039608595
          }
        ]
      }
    ]
  }
}
```

[...]

PR #494, opened 2017-03-28
Fix suggester score parsing

umpc:
This fixes issue #493.

I am not fully aware of the necessary fields that the type ```elastic.SearchSuggestionOption``` should have or not have, so I did not change anything else, though this likely hints that the block should be refactored to whatever the field names have been renamed to in the latest version of ElasticSearch.

Thank you for your help on this @olivere.


umpc:
Turns out that only the phrase suggester outputs its score to the ```score``` field, while completion suggesters, and possibly others, use ```_score```. This causes a regression and should not be merged until a better solution is found.

PR #495, opened 2017-03-28
RetryFunc added

utrack:
This PR replaces MaxRetries constant with RetryFunc that dynamically decides if a retry should be performed, making it possible to use rate-limiting and various feedback loops set by client's owner.

No APIs are broken except the removal of DefaultMaxRetries constant. It is possible to bring it back, but it would make the code less transparent.

Similar change can be made to every version of the client.

### DIFF
--- a/client.go
+++ b/client.go
@@ -79,10 +79,10 @@ const (
 )
 
 var (
-	// DefaultMaxRetriesFunc is the function that decides if a retry should be
+	// DefaultRetryFunc is the function that decides if a retry should be
 	// performed for a single request after Elastic gives up and returns an error.
 	// Retry is disabled by default (i.e. function always returns false).
-	DefaultMaxRetriesFunc = func(int) bool {
+	DefaultRetryFunc = func(int) bool {
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


utrack:
It seems that `SetMaxRetries` sets the max total HTTP request number per one client request instead of number of retries (excluding original request).

I've preserved previous behaviour for now and updated the `SetMaxRetries()`'s comment, but it can be misleading.

olivere:
Thanks for your contributions.

For elastic v3 (for ES 2.x) and v5 (for ES 5.x) I've already added [retries and backoff](https://github.com/olivere/elastic/wiki/Retrier-and-Backoff), which seem to be a bit more enhanced than your solution. However, they are not backported to v2 (for ES 1.x).

Are you still on ES 1.x, or did you overlook the Wiki page?

utrack:
@olivere Yep, I'm still on ES 1.x.

olivere:
Can you check if the [Retrier](https://github.com/olivere/elastic/wiki/Retrier-and-Backoff) will also work for you? If so, I'd prefer backporting to have a consistent implementation across all versions. Or even better: Create a PR based on e.g. [this commit for elastic.v3](https://github.com/olivere/elastic/commit/448e5f4a85ca197923e7ee49ee6b042d3f8022e4).

utrack:
Aha! Didn't see docs about the Retrier. Thanks, that'll do; I'll backport it and create a new PR. 🙂

olivere:
If CI is okay, I will push 2.0.58.

Thanks for your time working on the PR.

PR #501, opened 2017-03-29
Add support to use PUT rest api

muyadan:


olivere:
Thanks for your contributing. How is this different to [Index](https://github.com/olivere/elastic/blob/release-branch.v3/index.go)?

olivere:
I've modeled all services according to the [REST API spec](https://github.com/elastic/elasticsearch/tree/master/rest-api-spec/src/main/resources/rest-api-spec/api). There is no `Put` service in there. From the source I guess you want to use the `Index` service instead, which is [spec'd in this document](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/index.json).

olivere:
From the link to the documentation you might also be looking for the [`Update` service](https://github.com/olivere/elastic/blob/release-branch.v3/update.go), which is specified [here](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/update.json).

Whatever you're looking for, please find the service in the REST API spec. Then we can see if we're missing something.

muyadan:
Hey. Sorry for the delay.

The index only use "PUT" for new field. However, this will support using "PUT" for existing field which provides the ability to replacing existing item with a new one. The "Update" method can not removing fields for an existing item.

muyadan:
From the code. https://github.com/olivere/elastic/blob/release-branch.v3/index.go
It uses "PUT" when an item id does not exist.

olivere:
Can you please post a link to the official documentation that describes this behaviour? Or a test case in the REST API spec?

muyadan:
Hey. I could not find an official documentation in
https://github.com/elastic/elasticsearch/tree/master/rest-api-spec/src/main/resources/rest-api-spec/api
But in the elastic document
https://www.elastic.co/guide/en/elasticsearch/reference/2.4/_modifying_your_data.html#_modifying_your_data
It describes replacing data using "PUT" with existing id.

olivere:
The Index API has two endpoints, i.e. `POST /{index}/{type}` (to create a document and let ES create the ID) and `PUT /{index}/{type}/{id}` to upsert a document and provide the ID externally. Both of these endpoints are implemented by the [`IndexService` here](https://github.com/olivere/elastic/blob/release-branch.v5/index.go#L166-L182).

What endpoint are you missing?

olivere:
Any news here?

olivere:
Closing for inactivity.

magar91:
Hi I am also trying to use the PUT service for updating a whole document I'm using it like this 

```go
response, err := client.Index().
		Index("event").
		Type("event").
		Id(event.ID).
		BodyJson(event).
		Refresh("true").
		Do(ctx)
```
but that won't update the document already created, is this correct? what am I missing?

olivere:
I don't know what's wrong with that snippet. I posted an example [in this gist](https://gist.github.com/olivere/9f975d173a201deb6a8066c9f460f759), maybe that helps.

magar91:
it did help me @olivere thanks a lot!

muyadan:
Hey Oliver,

Sorry to get back late.

Based on the index doc and code. It should satisfy my need as what i put in
the merge request.

But i havent test it yet. Will get back to you later if it did not work.

Thanks,
Yadan

PR #503, opened 2017-03-31
v2: Added ability to do context-aware PerformRequest

utrack:
Small patch that adds `PerformRequestC` func which is usual `PerformRequest` that passes ctx to the HTTP client.

No API breakage or breakage of pre-1.7 builds.


utrack:
Bump. Are you OK with this @olivere?

olivere:
@utrack v2 is rather old, so I'm not pushing too aggressively for that. There has also been [some work recently](https://github.com/olivere/elastic/pull/514) to switch from `golang.org/x/net/context` to `context`, so this has to be done for v2 as well. We introduce `context` in v2 as well, of course, but all other service interfaces have to be switched as well, not just `PerformRequest`.

I won't have time to look into it this week. Maybe next weekend.

utrack:
Actually, v2 has no context-related code at all atm :)
I've used net/context for backwards compatibility, but it can be swapped in favour of vanilla context, of course.

olivere:
Yes. The PR mentioned above does exactly that. The `"context"` package is now available for 2 Go releases, and the majority of libraries are switching to it. I want elastic to do that as well.

However, the services need to be capable of using the `PerformRequestC`, which they are not with your change. Your change actually only add `context.Context` to `PerformRequest`. If you're not sure what I mean, see [this highlighted fragment from v3](https://github.com/olivere/elastic/blob/release-branch.v3/exists.go#L149-L182). It has to be done for all services in v2.

utrack:
Oh, I see. I'll add support for other services then, thanks.

olivere:
@utrack Yes, that'd be very helpful. However, keep in mind that I won't be able to merge this week as I'm out of the office. Sorry.

utrack:
@olivere Bump

olivere:
@utrack I'm sorry, I was on vacation last week and v2 is not my primary focus, so I updated 5.x first (and work on 6.x already). Please bear with me... it's just a matter of time. Your PR is very welcome, but I want to test it first.

client.go
```diff
@@ -227,7 +227,9 @@ func NewClient(options ...ClientOptionFunc) (*Client, error) {
 
 	if c.snifferEnabled {
 		// Sniff the cluster initially
-		if err := c.sniff(c.snifferTimeoutStartup); err != nil {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+		defer cancel()
+		if err := c.sniff(ctx, c.snifferTimeoutStartup); err != nil {
```
olivere: Please remove this and keep the code near to what we have in elastic.v5.

This opens more questions, but does not solve any issues. E.g. why a timeout of 10 seconds? Why cancel the context via timeout _and_ pass a timeout?

client.go
```diff
@@ -868,7 +876,7 @@ func (c *Client) healthchecker() {
 // the node state, it marks connections as dead, sets them alive etc.
 // If healthchecks are disabled this is a no-op.
 // The timeout specifies how long to wait for a response from Elasticsearch.
-func (c *Client) healthcheck(timeout time.Duration, force bool) {
+func (c *Client) healthcheck(ctx context.Context, timeout time.Duration, force bool) {
```
olivere: Same here. Please keep the code as it is in elastic.v5.

olivere:
@utrack Please do not add code that pushes v2 into a different direction than v5. It might be useful, but it's hard enough to keep 3 versions of a codebase. It's even harder when they slightly differ.

utrack:
Sorry for that, we've got the bug in production just yesterday. Without last two commits when ES cluster goes down - healthcheck waits for response indefinitely, blocking every other hc goroutine, leaking them and breaking the client. @olivere

olivere:
@utrack Now, that's a bug I'd like to see fixed. But I think we should address that in a separate issue/PR. Once you got the time, can you send a reproducible step-by-step/code snippet? It should also impact v3 and v5 then.

utrack:
OK @olivere, I'll try to create a reproducible case in a separate PR then :)
Reverted last two commits.

olivere:
Thanks for your support on getting context into v2. Even one of my projects will benefit from this :-)

PR #509, opened 2017-04-12
Add "version" field to IndicesGetTemplateResponse

andrew-d:
As per the Elasticsearch [docs][0], you can provide a "version" field that is used for external management of templates.  This adds the field to the response message so it can be read.

[0]: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html#versioning-templates


olivere:
Thanks for contributing! Will be merged shortly. Missed that for 5.0.33, so will release 5.0.34.

Issue #425, opened 2016-12-25
Switch to "context" from stdlib when 1.8 comes out


olivere:
We want to switch from `"golang.org/x/net/context"` to `"context"` when 1.8 is ready.

eticzon:
Hi @olivere! Thanks for this awesome lib! I'd like to get involved, if that's OK. 😄  Would you mind if I start with this "admin"-y task first?

olivere:
I'd appreciate support if you're ready to spend some of your time. I agree changing `golang.org/x/net/context` to `context` should be pretty straightforward. So if you want, feel free to create a PR. Please do so for v5 first.

eticzon:
Thanks @olivere! Much appreciated! PR's up as well when you've got time to review.

olivere:
@eticzon I haven't looked into this recently. But can we also switch away from `golang.org/x/net/context/ctxhttp` now that we have `WithContext(...)` in stdlib?

eticzon:
Good point @olivere. I'll change the calls to `ctxhttp` then and replace them `WithContext(...)` alternatives.

olivere:
Thanks so much for working on this. It's been tagged as 5.0.36 now.

PR #514, opened 2017-04-21
Switch to stdlib context package (go1.8+)

eticzon:
Unfortunately the `ctxhttp` package is still under the `"x"` directory.  So we're still left with the `"golang.org/x/net/context/ctxhttp"` import path for `client.go` and `ping.go`. Hopefully this would be merged to stdlib in the near future.

This should address issue #425.

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -6,11 +6,10 @@ package elastic
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"net/url"
-
-	"golang.org/x/net/context"
 
 	"gopkg.in/olivere/elastic.v5/uritemplates"
 )
--- a/bulk_processor.go
+++ b/bulk_processor.go
@@ -5,11 +5,10 @@
 package elastic
 
 import (
+	"context"
 	"sync"
 	"sync/atomic"
 	"time"
-
-	"golang.org/x/net/context"
 )
 
 // BulkProcessorService allows to easily process bulk requests. It allows setting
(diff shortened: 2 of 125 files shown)


eticzon:
> Unfortunately the ctxhttp package is still under the "x" directory. So we're still left with the "golang.org/x/net/context/ctxhttp" import path for client.go and ping.go. Hopefully this would be merged to stdlib in the near future.

As @olivere has pointed out the `http.Client` is now `context`-aware and `http.Request` has the handy `WithContext(...)` method.

On both `client.go` and `ping.go` the use of `ctxhttp` has been removed.

olivere:
Awesome. Thank you very much.

I'm still undecided if we should do this breaking change in the middle of the 5.0 release cycle. What do you think?

olivere:
Hmm. I checked three of my applications and they all compile without any problem with your changes. So I think there might only be a problem for those poor people still using libraries like [`google.golang.org/grpc`](https://github.com/grpc/grpc-go) who (for whatever reason) [don't like to switch to `"context"`](https://github.com/grpc/grpc-go/issues/711).

eticzon:
Thanks for merging. 👍 

IMHO since `context` has been in the stdlib for several releases now it maybe best to start do this now (rather than later).

I guess if for those projects using [https://github.com/grpc](url), if and when they raise an issue, we could refer them back to this PR and they can decide on how best to tackle their dependency problem.

From that issue discussion, there seems to be a strong push to start doing the same migration from from `golang.org/x/net/context` to the stdlib version, so hopefully it won't be a problem for them soon.

PR #520, opened 2017-04-27
Add more data on BulkResponseItem

arquivei:
In order to know if the bulk request returned 'noop' as the result, this commit added one more parameter to the BulkResponseItem struct.

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -305,6 +305,7 @@ type BulkResponseItem struct {
 	Id      string        `json:"_id,omitempty"`
 	Version int64         `json:"_version,omitempty"`
 	Status  int           `json:"status,omitempty"`
+	Result  string        `json:"result,omitempty"`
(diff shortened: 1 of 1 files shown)


olivere:
Thanks for that.

Can you also add the `forced_refresh` flag (a `bool`)? It is returned with every [`DocWriteResponse`](https://github.com/elastic/elasticsearch/blob/5.x/core/src/main/java/org/elasticsearch/action/DocWriteResponse.java#L57-L58)--which [`BulkItemResponse` uses](https://github.com/elastic/elasticsearch/blob/5.x/core/src/main/java/org/elasticsearch/action/bulk/BulkItemResponse.java).

arquivei:
Hey there,

I've just updated the pull request including this new flag :+1:

olivere:
Thanks. I hope to merge on the weekend. This one is a no brainer. :-)

PR #522, opened 2017-05-02
Error field of GetResult of type string instead of ErrorDetails

malpiatko:
### WHAT
* Change the type of field `Error` to string instead of `ErrorDetails`.

### WHY
* `Error` field is used only in `MultiGet` which returns a `string` error rather than `ErrorDetails`. If an error occurred for one of the documents in the `MultiGet` request this causes the following unmarshalling error:

`json: cannot unmarshal string into Go value of type elastic.ErrorDetails`

this obscured the actual error.

### WHO
@yayalice

### DIFF
--- a/get.go
+++ b/get.go
@@ -250,7 +250,5 @@ type GetResult struct {
 	Source  *json.RawMessage       `json:"_source,omitempty"`
 	Found   bool                   `json:"found,omitempty"`
 	Fields  map[string]interface{} `json:"fields,omitempty"`
-	//Error     string                 `json:"error,omitempty"` // used only in MultiGet
-	// TODO double-check that MultiGet now returns details error information
-	Error *ErrorDetails `json:"error,omitempty"` // only used in MultiGet
+	Error   string                 `json:"error,omitempty"` // used only in MultiGet
 }
--- a/mget_test.go
+++ b/mget_test.go
@@ -66,7 +66,7 @@ func TestMultiGet(t *testing.T) {
 	}
 
 	item := res.Docs[0]
-	if item.Error != nil {
+	if len(item.Error) != 0 {
 		t.Errorf("expected no error on item 0; got %v", item.Error)
 	}
 	if item.Source == nil {
@@ -81,7 +81,7 @@ func TestMultiGet(t *testing.T) {
 	}
 
 	item = res.Docs[1]
-	if item.Error != nil {
+	if len(item.Error) != 0 {
 		t.Errorf("expected no error on item 1; got %v", item.Error)
 	}
 	if item.Source == nil {


malpiatko:
To clarify the above mentioned error happens when running the following code:
```
multigetRequest = es.MultiGet()

noSourceContext := elastic.NewFetchSourceContext(false)

for _, docID := range docIDs {
	request := elastic.NewMultiGetItem().
		Index(index).
		Type(docType).
		Id(docID).
		FetchSource(noSourceContext)

	multiGetRequest.Add(request)
}

response, err := multiGetRequest.Do(context.Background())
```

The faulty unmarshalling happens at [mget.go#L134](https://github.com/olivere/elastic/blob/release-branch.v5/mget.go#L134)

olivere:
Hmm... if I look [at the Java source](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/action/get/MultiGetResponse.java#L140) and the response from HTTP requests (see below), it seems that it returns `*ErrorDetails` not a `string`, but maybe I'm missing something.

```
$ curl -H 'Content-Type: application/json' -is 'http://localhost:9200/no-such-index/_mget?pretty' -d '{"ids":["1", "2"]}'
HTTP/1.1 200 OK
content-type: application/json; charset=UTF-8
content-length: 1352

{
  "docs" : [
    {
      "_index" : "no-such-index",
      "_type" : null,
      "_id" : "1",
      "error" : {
        "root_cause" : [
          {
            "type" : "index_not_found_exception",
            "reason" : "no such index",
            "resource.type" : "index_expression",
            "resource.id" : "no-such-index",
            "index_uuid" : "_na_",
            "index" : "no-such-index"
          }
        ],
        "type" : "index_not_found_exception",
        "reason" : "no such index",
        "resource.type" : "index_expression",
        "resource.id" : "no-such-index",
        "index_uuid" : "_na_",
        "index" : "no-such-index"
      }
    },
    {
      "_index" : "no-such-index",
      "_type" : null,
      "_id" : "2",
      "error" : {
        "root_cause" : [
          {
            "type" : "index_not_found_exception",
            "reason" : "no such index",
            "resource.type" : "index_expression",
            "resource.id" : "no-such-index",
            "index_uuid" : "_na_",
            "index" : "no-such-index"
          }
        ],
        "type" : "index_not_found_exception",
        "reason" : "no such index",
        "resource.type" : "index_expression",
        "resource.id" : "no-such-index",
        "index_uuid" : "_na_",
        "index" : "no-such-index"
      }
    }
  ]
}
```

The exception in ES is created [here](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/action/get/MultiGetResponse.java#L140). However, [`ElasticsearchException.generateFailureXContent(...)`](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/ElasticsearchException.java#L539-L576) has a `detailed` parameter which can be set to `false` to return a simple `string` instead (though `detailed` is `true` for MultiGet). So maybe there's a way to trigger a simplified error message. In that case I think Elastic should still use `*ErrorDetails` and initialize with the `string` returned from ES.

Can you construct a query that returns a string only?

malpiatko:
I am closing this as I think what happened was that I used the wrong library version. Will reopen if this was not the case

PR #527, opened 2017-05-14
Prefix query new format

alehano:
Use field `value` instead of `prefix` to get rid of deprecated warning.


olivere:
The PR was not quite correct. The query

```

```

... is still valid and not deprecated according to the docs.

No worries. I will create a fix myself. Thanks for pointing this out.

PR #532, opened 2017-05-19
breakers is plural in json response

kmulvey:


olivere:
Thanks for the hint. I'm taking a deep look at the whole struct on the next occasion. I'm off for holidays for a week so please bear with me.

olivere:
Will be available in 5.0.39. Thanks again.

PR #533, opened 2017-05-19
Add Snapshot Create API

wedneyyuri:
This PR adds support for Elasticsearch's [snapshot create API](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html).

Example program:
```go
package main

import (
	"context"
	"log"
	"os"
	"time"

	elastic "gopkg.in/olivere/elastic.v5"
)

func main() {
	client, err := elastic.NewClient()
	if err != nil {
		panic(err)
	}

	// create snapshot repo
	repo := elastic.
		NewSnapshotCreateRepositoryService(client).
		Repository("repo1").
		BodyJson(map[string]interface{}{
			"type": "fs",
			"settings": map[string]interface{}{
				"location": "/tmp/elasticsearch",
			},
		})

	repoResponse, err := repo.Do(context.TODO())
	if err != nil {
		panic(err)
	}

	if !repoResponse.Acknowledged {
		panic("could not ack repository create")
	}

	// create snapshot
	service := elastic.
		NewSnapshotCreateService(client).
		Repository("repo1").
		Snapshot(time.Now().Format("2006-01-02-15-04-05")).
		WaitForCompletion(true).
		BodyJson(map[string]interface{}{
			"indices":              "website",
			"ignore_unavailable":   false,
			"include_global_state": true,
		})

	response, err := service.Do(context.TODO())
	if err != nil {
		panic(err)
	}

	if response.Snapshot.State != "SUCCESS" {
		panic("elasticsearch snapshot state != SUCCESS: " + response.Snapshot.State)
	}
}
```

### DIFF
--- a/snapshot_create.go
+++ b/snapshot_create.go
@@ -11,7 +11,7 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/olivere/elastic/uritemplates"
+	"gopkg.in/olivere/elastic.v5/uritemplates"
 )
 
 // SnapshotCreateService is documented at https://www.elastic.co/guide/en/elasticsearch/reference/5.x/modules-snapshots.html.
@@ -150,6 +150,16 @@ func (s *SnapshotCreateService) Do(ctx context.Context) (*SnapshotCreateResponse
 	return ret, nil
 }
 
+// SnapshotShardFailure stores information about failures that occurred during shard snapshotting process.
+type SnapshotShardFailure struct {
+	Index     string `json:"index"`
+	IndexUUID string `json:"index_uuid"`
+	ShardID   int    `json:"shard_id"`
+	Reason    string `json:"reason"`
+	NodeID    string `json:"node_id"`
+	Status    string `json:"status"`
+}
+
 // SnapshotCreateResponse is the response of SnapshotCreateService.Do.
 type SnapshotCreateResponse struct {
 	// Accepted indicates whether the request was accepted by elasticsearch.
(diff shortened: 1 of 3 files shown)


snapshot_create.go
```diff
@@ -0,0 +1,179 @@
...
+package elastic
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"time"
+
+	"github.com/olivere/elastic/uritemplates"
```
olivere: This must be `"gopkg.in/olivere/elastic.v5/uritemplates"`. The generator need some polish, I'm sorry.

snapshot_create.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// SnapshotCreateResponse is the response of SnapshotCreateService.Do.
+type SnapshotCreateResponse struct {
+	// Accepted indicates whether the request was accepted by elasticsearch.
+	// It's available when waitForCompletion is false.
+	Accepted *bool `json:"accepted"`
+
+	// Snapshot is available when waitForCompletion is true.
+	Snapshot *struct {
```
olivere: The `*struct`, although valid, doesn't look that good. Does ES always send back the `snapshot`? We have to look it up in the Java source. If it does, we should make it a normal `struct`, not a pointer to a struct.

snapshot_create.go
```diff
@@ -0,0 +1,179 @@
...
+		Snapshot          string        `json:"snapshot"`
+		UUID              string        `json:"uuid"`
+		VersionID         int           `json:"version_id"`
```
olivere: Did you compare these fields with what ES really sends, or did you make an educated guess? I typically consult the Java source to see what types ES really sends and more often than not, I find fields that are undocumented but sent anyway.

snapshot_create.go
```diff
@@ -0,0 +1,179 @@
...
+		Version           string        `json:"version"`
+		Indices           []string      `json:"indices"`
+		State             string        `json:"state"`
+		StartTime         time.Time     `json:"start_time"`
+		StartTimeInMillis int64         `json:"start_time_in_millis"`
+		EndTime           time.Time     `json:"end_time"`
+		EndTimeInMillis   int64         `json:"end_time_in_millis"`
+		DurationInMillis  int           `json:"duration_in_millis"`
```
olivere: This e.g. should be `int64` if `StartTimeInMillis` is `int64` as well.

snapshot_create.go
```diff
@@ -0,0 +1,179 @@
...
+		Failures          []interface{} `json:"failures"`
+		Shards            struct {
```
olivere: There is a [`shardsInfo` in `errors.go`](https://github.com/olivere/elastic/blob/1368b91516f39100baf5e23a02ae9f090c0d90e2/errors.go#L128-L133) that you can use here.

olivere:
The PR is looking good. Can you please take a look at those comments, so I can merge in the next run. I'm off for vacation next week so it'll take a few more days, I'm afraid.

snapshot_create.go
wedneyyuri: :+1:

snapshot_create.go
wedneyyuri: On my previous commit I compared these field with what ES really sends (turning logging on), but today I made a new commit [consulting the java source code](https://github.com/elastic/elasticsearch/blob/45d4938fcc8f1a4cf988e119d7dd23bb2920b9e2/core/src/main/java/org/elasticsearch/snapshots/SnapshotInfo.java#L298).

wedneyyuri:
Thanks @olivere, I looked at your comments and pushed some changes.

olivere:
It'll roll with 5.0.39. Thanks.

wedneyyuri:
Thanks @olivere, can I backport it to v3?

olivere:
@wedneyyuri Sure. That seems to be really easy. All you need to do is change some import paths.

PR #543, opened 2017-06-14
[v*] Healthcheck goroutine leaks

utrack:
Hi! This PR fixes goroutine leaks caused by healthcheck not using ctx properly.
I've managed to reproduce it via tests.

### DIFF
--- a/client.go
+++ b/client.go
@@ -817,8 +817,12 @@ func (c *Client) sniff(timeout time.Duration) error {
 
 	// Start sniffing on all found URLs
 	ch := make(chan []*conn, len(urls))
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
 	for _, url := range urls {
-		go func(url string) { ch <- c.sniffNode(url) }(url)
+		go func(url string) { ch <- c.sniffNode(ctx, url) }(url)
 	}
 
 	// Wait for the results to come back, or the process times out.
@@ -829,7 +833,7 @@ func (c *Client) sniff(timeout time.Duration) error {
 				c.updateConns(conns)
 				return nil
 			}
-		case <-time.After(timeout):
+		case <-ctx.Done():
 			// We get here if no cluster responds in time
 			return errors.Wrap(ErrNoClient, "sniff timeout")
 		}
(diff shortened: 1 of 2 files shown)


olivere:
That looks interesting. Thanks for reporting/fixing it. BTW: Tests fail for Go 1.7 because `http.Server` in 1.7 doesn't have a `Shutdown` method.

utrack:
Hmm. Stumbled upon something re: golang/go#20668.

olivere:
@utrack That works as expected here: Context is canceled.

utrack:
@olivere on Go 1.8 it works OK, but fails on 1.7.

olivere:
@utrack Oh, I see.

utrack:
I've also prepared v2 backport, should I open up another PR?

utrack:
@olivere Fixed tests for all versions. It's not possible to monitor leaks directly on Go pre-1.8, so I'm using side effects to monitor stuck connections on pre-1.8.

olivere:
@utrack Okay, thanks for figuring this out. Can you do three more things for me to merge this into v5:

1. As far as I can see it, you only have a test to check leaks for `healthcheck`. Can you also add a test for the `sniffNode`?
2. Can you please rename the test for `healthcheck` from `TestRoutineLeaks` to `TestClientHealthcheckTimeout` (or something more expressive). The test for the sniffer should be something like `TestClientSnifferTimeout`.
3. Can you move the tests to the end of the other `TestClientHealthcheck...` (and `TestClientSniffer...`) tests.

Then we can put together PRs for v2 and v3, or I'll do this in a batch.

utrack:
Sure, done @olivere

olivere:
Merged and will be published soon. Thanks again!

olivere:
@utrack Do you mind to review #545 (and [this commit](https://github.com/olivere/elastic/commit/902b51293ec58a82a15294e8b33069f53ba44f63)). It's this PR backported into elastic.v3.

I will also backport to elastic.v2 later with #546, after merging #503 first.

utrack:
Reviewed everything :) Looks good, thank you for the client!

olivere:
@utrack Thanks. Have a nice weekend. cu.

wtfiwtz:
Looks like this breaks Go 1.6.2... I'm using elastic.v3... Upgrading Go to 1.8.3 resolved it...

```
 % wget https://storage.googleapis.com/golang/go1.8.3.linux-amd64.tar.gz
 % sudo tar -C /usr/local -xzf go1.8.3.linux-amd64.tar.gz
 % sudo mv /usr/bin/go /usr/bin/go1.6
 % sudo ln -s /usr/local/go/bin/go /usr/bin/go
 % go version
go version go1.8.3 linux/amd64
 % go build
 %
```

olivere:
@wtfiwtz I see. However, 1.6 is no longer supported. I try to support the current and previous versions of Go (while Go itself even [only supports the current version](https://golang.org/doc/devel/release.html#policy)). From my experience, updating Go is nowhere as critical as in other environments.

caseyw:
Note that this bit me as well. I've updated to 1.8 and everything compiled.

Thanks,

-c

PR #544, opened 2017-06-14
Boost Suggestions by Context

peteclark-ft:
Elasticsearch supports boosting by certain suggester categories - see [here](https://www.elastic.co/guide/en/elasticsearch/reference/5.2/suggester-context.html#_category_query). 

This PR adds support for the boost field, while maintaining backwards compatibility.

**N.B.** The order of contexts is not important to the query, and so I didn't maintain their order in the json output. I didn't want to change the way you did tests (by parsing the json and comparing them), or add any new dependencies, so I simply gave the output two acceptable options. Hope that's ok.

### DIFF
--- a/suggester_completion.go
+++ b/suggester_completion.go
@@ -123,7 +123,7 @@ func (q *CompletionSuggester) Source(includeName bool) (interface{}, error) {
 				ctxq[k] = v
 			}
 		}
-		suggester["context"] = ctxq
+		suggester["contexts"] = ctxq
 	}
 
 	// TODO(oe) Add completion-suggester specific parameters here
--- a/suggester_completion_test.go
+++ b/suggester_completion_test.go
@@ -45,9 +45,8 @@ func TestCompletionSuggesterSourceWithMultipleContexts(t *testing.T) {
 		t.Fatalf("marshaling to JSON failed: %v", err)
 	}
 	got := string(data)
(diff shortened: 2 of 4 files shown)


olivere:
Hey! Thanks for your supporting elastic. I really need to read the documentation on suggesters first (I'm not an expert in this field), but the PR's looking good as far as I see it.

Re the tests: The output of the Go JSON serializer should be deterministic, so there should be no need for two expectations. Was it nondeterministic in your tests?

peteclark-ft:
Hi @olivere - weirdly I ran my tests locally and they passed, but they didn't pass in travis; so I assumed maps weren't ordered, and gave both options.. But you're right, the map key order is lexicographically sorted, so it should always output the same thing.

I've amended the tests back!

olivere:
Hmm... okay. The `TestSuggesterCategoryQueryWithoutBoost` test fails sporadically as `SuggesterCategoryQuery.values` is a `map`. The keys in a `map` are randomized when used in a `range` loop, so the test may or may not fail. I will change the test accordingly, and make it work in both cases.

peteclark-ft:
@olivere - ahh of course; I knew I wasn't going mad! We could maintain an additional array for ordering?

olivere:
@peteclark-ft I think I'll just make the test check for both options, as you already did. ES shouldn't care about the order, so no need to preserve the order.

peteclark-ft:
@olivere Thanks for merging! Good luck in the future!

olivere:
Your PR is going to make it into ES 5.0.40 soon. Thanks for your contributions.

PR #550, opened 2017-07-06
Support reindex as asynchronous task

khatton-ft:
When the `WaitForCompletion` parameter is set to `true`, the response from ElasticSearch is in a different format, as it just contains a task id. This PR provides a `Start` method on the ReindexService that will return the task id instead of the `BulkIndexByScrollResponse`, which is not appropriate for asynchronous requests. (I believe this could also be appropriate for the other Elastic APIs that have a `WaitForCompletion` parameter, although my present use case does not extend to them).

Secondly, the `TasksListService` does not work as expected with single `taskId` and `nodeId` values, so, for the simple case of monitoring a single task, I have added a `TasksGetTaskService`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1625,6 +1625,11 @@ func (c *Client) TasksList() *TasksListService {
 	return NewTasksListService(c)
 }
 
+// TasksGetTask retrieves a task running on the cluster.
+func (c *Client) TasksGetTask() *TasksGetTaskService {
+	return NewTasksGetTaskService(c)
+}
+
 // TODO Pending cluster tasks
 // TODO Cluster Reroute
 // TODO Cluster Update Settings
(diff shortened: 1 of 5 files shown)


olivere:
Thanks for contributing to Elastic. The PR is looking good. Give me some time to make a few more tests. But I hope it'll make it into the next release.

olivere:
I changed a few things in your original PR. See [3116eca](https://github.com/olivere/elastic/commit/3116eca7151ed9e9d7aea8707ba3f0e13ac4300d) for a summary.

Most noticeable is that I renamed `Start` to `DoAsync`. I want users to remember that `Do...` executes the request.

olivere:
Forgot: This will make it into 5.0.42, which I'll release once CI is green.

khatton-ft:
:+1:

PR #560, opened 2017-07-17
feat(update): allow the client to retrieve the document _source in updates

connor4312:
`GetResult` was present in the UpdateResponse, but no way was provided to set `_source` to `true`, needed to get that result ([docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#_parameters_2)). The result also is not a full `GetResult`, containing only `found` and `_source` in my ES5 instance, so I added a separate struct for that.

### DIFF
--- a/update.go
+++ b/update.go
@@ -6,6 +6,7 @@ package elastic
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/url"
 	"strings"
@@ -34,6 +35,7 @@ type UpdateService struct {
 	scriptedUpsert      *bool
 	docAsUpsert         *bool
 	detectNoop          *bool
+	includeSource       *bool
 	doc                 interface{}
 	timeout             string
 	pretty              bool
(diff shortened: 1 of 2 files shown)


olivere:
Thanks for your PR.

This needs to be done via [`FetchSourceContext`](https://github.com/olivere/elastic/blob/release-branch.v5/fetch_source_context.go), which isn't added to the Update API. I'll prepare an update to 5.0.42.

olivere:
ES also [returns a full `GetResult`](https://github.com/elastic/elasticsearch/blob/4796557a305013a5c6529bfdb0e693ee44ffe7e2/core/src/main/java/org/elasticsearch/action/update/UpdateResponse.java#L40) (which might miss some fields), so there's no additional struct necessary.

connor4312:
Ah, great! Somehow I totally missed that when grepping for `_source` in the update. I might have been on the master branch instead of v5. Thanks 😄

PR #562, opened 2017-07-17
feat(errors): add IsConflict checker

connor4312:
Self-explanatory!


olivere:
Thanks for contributing to Elastic. This seems like a good choice, knowing that it's a typical ES response code. However, I think I'll refactor it a bit and extract a `IsStatusCode(code int)`, then use that for both `IsConflict` and `IsNotFound`.

PR #577, opened 2017-08-08
Add custom headers on HTTP requests

elaPa:
This is to provide a mechanism to set custom headers on all HTTP requests to elastic search.
There are two levels where this can be done:
- set custom headers on the client, in which case they would be populated on all http requests (including healthcheck, ping etc) - this could be useful eg. when you use a custom authorization mechanism
- set custom headers on each http request explicitly - this could be useful eg. for auditing

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -42,7 +42,7 @@ type BulkService struct {
 	sizeInBytes       int64
 	sizeInBytesCursor int
 
-	headers map[string]string
+	headers map[string][]string
 }
 
 // NewBulkService initializes a new BulkService.
@@ -120,8 +120,8 @@ func (s *BulkService) Pretty(pretty bool) *BulkService {
 }
 
 // Headers adds headers on the http request
-func (s *BulkService) Headers(headers map[string]string) *BulkService {
-	s.headers = headers
+func (s *BulkService) Header(key, value string) *BulkService {
+	s.headers = addHeader(s.headers, key, value)
 	return s
 }
 
--- a/bulk.go
+++ b/bulk.go
@@ -119,7 +119,7 @@ func (s *BulkService) Pretty(pretty bool) *BulkService {
 	return s
 }
 
-// Headers adds headers on the http request
+// Header adds key, value pair to the header on the http request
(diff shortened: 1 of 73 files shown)


olivere:

There are already two ways to e.g. pass additional headers: First, you could [create your own `http.Transport`](https://github.com/olivere/elastic/wiki/HttpTransport). Second, you could [create and set your own `http.Client`](https://github.com/olivere/elastic/blob/release-branch.v5/client.go#L429) which Elastic will use under the covers.

Can you explain why we should pollute the services with cross-cutting concerns like HTTP headers, and not use any of the above?

elaPa:
Thank you for the quick response.
Our requirement is to be able to set custom headers per request for the auditing purposes.
If I understand correctly, the solutions that you suggested would set headers on the client - so the same headers would be used for all requests. What we need is the ability to set different audit token (eg uuid) per request.

olivere:
You can do that with a transport as well. Just switch over the `http.Request.URL`.

elaPa:
Sorry, I haven't explained it properly. 
Yes, using transport would work, if we only wanted to append uuid (or another randomly generated value).
But what we need to put on the headers is some audit token, different for each request, passed in by the caller, which will then help us linking each request in Elasticsearch audit logs with the caller (ie to have an ability to audit who and why has accessed specific data).
I don't think it is possible with transport - but maybe I'm missing something?

olivere:
Oh well, you could maybe do that by passing request-level data in `context.Context` and let the transport inject it into the request.

I don't know your case exactly, of course. But if auditing is required, you should probably do it on a different layer, not in the interface to the data store, no? Or, if you really need to do this at such a low level, maybe a proxy would do the job.

Anyway, I'll keep this open to think about it. But at the moment I'm quite negative about this edge-case sprinkling stuff all over the services.

elaPa:
Ah! I didn't realized that the context is set on the http request itself! My bad!
Yes, that definitely solves my problem ;) 
Thank you for your help and good pieces of advice, I'm closing the pull request then.

dustin-decker:
@elaPa could you share how you add headers to outgoing requests from elastic.Client using context?

PR #586, opened 2017-08-18
Bind client to specific cluster name

utrack:
This PR makes client check the node's ClusterName before using it to sniff nodes and/or marking it as healthy.

Currently it's possible that one node might go down and get up as a node from different cluster, respond to the sniffer faster than everyone else and effectively take over the client, making it query different cluster. Same can happen on startup if one of seed nodes had been moved to the different cluster.

### DIFF
--- a/client.go
+++ b/client.go
@@ -543,6 +543,9 @@ func SetHealthcheck(enabled bool) ClientOptionFunc {
 
 // SetClusterName restricts client to specific cluster name (nodes from different
 // clusters are treated as dead).
+//
+// This setting works properly in all situations
+// only with healthcheck and sniffing enabled.
 func SetClusterName(name string) ClientOptionFunc {
 	return func(c *Client) error {
 		c.clusterName = name
--- a/client.go
+++ b/client.go
@@ -137,6 +137,7 @@ type Client struct {
 	requiredPlugins           []string        // list of required plugins
 	gzipEnabled               bool            // gzip compression enabled or disabled (default)
 	retrier                   Retrier         // strategy for retries
+	clusterName               string          // cluster name if set
 }
 
 // NewClient creates a new client to work with Elasticsearch.
@@ -540,6 +541,16 @@ func SetHealthcheck(enabled bool) ClientOptionFunc {
 	}
 }
 
+// SetClusterName restricts client to specific cluster name (nodes from different
+// clusters are treated as dead).
+func SetClusterName(name string) ClientOptionFunc {
+	return func(c *Client) error {
+		c.clusterName = name
+		return nil
+	}
+
+}
+
 // SetHealthcheckTimeoutStartup sets the timeout for the initial health check.
 // The default timeout is 5 seconds (see DefaultHealthcheckTimeoutStartup).
 // Notice that timeouts for subsequent health checks can be modified with
@@ -861,13 +872,18 @@ func (c *Client) sniff(timeout time.Duration) error {
 	}
 
 	// Start sniffing on all found URLs
-	ch := make(chan []*conn, len(urls))
+	ch := make(chan []*conn)
 
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
 	for _, url := range urls {
-		go func(url string) { ch <- c.sniffNode(ctx, url) }(url)
+		go func(url string) {
+			select {
+			case ch <- c.sniffNode(ctx, url):
+			default:
+			}
+		}(url)
 	}
 
 	// Wait for the results to come back, or the process times out.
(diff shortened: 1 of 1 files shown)


utrack:
I'll need your help writing tests for this - I haven't found any mention of cluster name used in Jenkins setup.

olivere:
Does the [SniffingCallback](https://github.com/olivere/elastic/wiki/Sniffing#sniffing-callback) already solve this problem?

utrack:
Oh, didn't notice it. Neat!
It won't solve the problem completely though, since previously healthy node can became node from foreign cluster, bypassing the new-node process in sniffer; plus, node roster would be ruined if we'll sniff a node from foreign cluster - yielding only foreign cluster's nodes.

olivere:
> since previously healthy node can became node from foreign cluster

You mean existing nodes switch clusters? That'd indeed be very bad.

> node roster would be ruined if we'll sniff a node from foreign cluster

If `SnifferCallback` returns true, it shouldn't become a node in Elastic and shouldn't receive any requests. How `SnifferCallback` determines that is up to the callback.

I still struggle to see why nodes should/would switch clusters? I mean, it _could_ happen, but only if I've not done my homework on the infrastructure level (i.e. separate nodes of clusters). Can you give an example of why that would be beneficial?

utrack:
Well, we've hit this on prod when switching to new major version of Elastic. Since it had to be zero-downtime op we were using same physical nodes (nA) to build up a new cluster (cB) that is identical to current setup (cA) power-wise, gradually switch load to the new cluster cB while moving old machines from cA to new cluster as load moves to cB and then removing stand-in nodes nA to do the same later with remaining clusters running old version of Elastic.

While nodes nA were down for ~2min before forming new cluster after transition was completed - client still picked them up somehow and sniffed nodes of completely foreign cluster.

Besides, it's possible that someone somewhere would want to juggle nodes between clusters if there's high load expected for one of them.

This PR doesn't make the cluster binding mandatory, but it would create an additional safeguard for those who want it.

olivere:
Hmm... a node can only join a cluster if it shares its cluster name ([docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/important-settings.html#cluster.name)). So I'm still a bit surprise how you could end up with a node joining two different clusters other than the clusters shared the same name.

Anyway, I guess the requirement of isolating the nodes suitable for sniffing could be satisfied by something like this:

```
package main

import (
	"context"
	"log"

	elastic "gopkg.in/olivere/elastic.v5"
)

func main() {
		elastic.SetSnifferCallback(EnforceVersion("5.5.2")),
	)
	if err != nil {
		log.Fatalf("Connect failed: %v", err)
	}
	log.Print("Connected")

	_, err = client.IndexExists("twitter").Do(context.Background())
	if err != nil {
		log.Fatalf("IndexExists failed: %v", err)
	}
	log.Print("Done")
}

func EnforceVersion(version string) elastic.SnifferCallback {
	return func(node *elastic.NodesInfoNode) bool {
		// log.Printf("%#v", node)
		return node.Version == version
	}
}
```

Notice that the cluster name is not available in `elastic.NodesInfoNode`. As a workaround, you could annotate the nodes accordingly and use the `Attributes` property to figure out what to do.

utrack:
Hi! Sorry for the long response.

This solution wouldn't help - bad IndexExists response would require us to recreate client on failure w/ downtime, since roster is already poisoned and there's no way back.
EnforceVersion would wipe all nodes from the roster if foreign node is being sniffed.

I'm not exactly sure of exact sequence of events, but I suspect that node left the cluster and joined another cluster between the sniffing rounds, and then sniffer sniffed foreign nodes from it.
The node didn't join two different clusters, it just left one cluster and joined another one fast.

olivere:
@utrack I still don't see how a node might decide to leave cluster A and then join a totally different cluster B--with a different cluster name and, as you state, a different major version.

If you want to enforce Elastic to only use specific nodes/node configurations you might turn off sniffing altogether and e.g. label the nodes with annotations like specified in the [Wiki](https://github.com/olivere/elastic/wiki/Sniffing#sniffing-callback).

If that isn't feasible and your original solution solves the problem you should probably use your version until you completed the migration to the new cluster. Quite frankly I don't see a way to integrate it into the client because I think it solves a problem that shouldn't exist in the first place.

utrack:
It did not "decide" to move, it was reprovisioned fast enough that client wasn't able to catch up with the changes.
Official Java client explicitly requires cluster name binding as well, see https://www.elastic.co/guide/en/elasticsearch/client/java-api/5.5/transport-client.html

olivere:
@utrack You are refering to the `TransportClient`. The `TransportClient` works by being part of the cluster, is native Java, uses a binary protocol to talk to cluster nodes (not the REST API), and requires the cluster name for connecting to the cluster. AFAIK it's being retired and should be replaced by the official REST client for Java.

Elastic is a HTTP client, built in a similar fashion as the official clients for Ruby, Python, and now Java.

I'm trying to keep Elastic in line with the official clients. Until now, I haven't seen any configuration setting or code snippet that enforces cluster name affinity (see e.g. [`client.go`](https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-transport/lib/elasticsearch/transport/client.rb) or [`sniffer.rb`](https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-transport/lib/elasticsearch/transport/transport/sniffer.rb)). Can you point me to a code snippet or thread that discusses cluster name affinity?

PR #587, opened 2017-08-23
v2: Let FunctionScoreQuery allow both Query and Filter

edsrzf:
Previously, FunctionScoreQuery only allowed either Query or Filter to be set,
setting the other to nil. There's no reason to do this, though, as
ElasticSearch allows both to be set.

See the equivalent Java code here:
https://github.com/elastic/elasticsearch/blob/v1.7.6/src/main/java/org/elasticsearch/index/query/functionscore/FunctionScoreQueryBuilder.java#L85-L88


olivere:
Yep. Although v2 is unsupported, I will fix it in the next update. Thanks for reporting.

olivere:
Thanks.

PR #593, opened 2017-09-06
Use v5 sniffNode logic in v3

eyeamera:
It appears that the logic for sniffing nodes between version 3 and 5 has diverged. I'm not entirely certain if this intentional or not. 

Currently, enabling sniffing in v3 and setting the client's scheme to `https` while querying against a cluster setup with [Search Guard SSL](https://github.com/floragunncom/search-guard-ssl) will produce an empty node list due to the fact that Search Guard will not return a `https_address` key for each node.

Here's a snippet of what a node returns for `GET /_nodes/http` with Search Guard SSL setup:

```
{
  "cluster_name" : "elasticsearch-cluster",
  "nodes" : {
    "m1Mp_vxXRsudaMXvNeKtQg" : {
      "transport_address" : "10.0.0.111:9300",
      "host" : "10.0.0.111",
      "ip" : "10.0.0.111",
      "http_address" : "10.0.0.111:9200",
      "http" : {
        "bound_address" : [ "0.0.0.0:9200" ],
        "publish_address" : "10.0.0.111:9200",
      }
      ...
    }
    ...
  }
}
```

My only concern here is that I'm unaware of how a ES v2.x node behaves with Shield installed to enable SSL... Do you have any insight there?

Happy to take any feedback and adjust accordingly.

### DIFF
--- a/client.go
+++ b/client.go
@@ -847,21 +847,11 @@ func (c *Client) sniffNode(ctx context.Context, url string) []*conn {
 	var info NodesInfoResponse
 	if err := json.NewDecoder(res.Body).Decode(&info); err == nil {
 		if len(info.Nodes) > 0 {
-			switch c.scheme {
-			case "https":
-				for nodeID, node := range info.Nodes {
-					url := c.extractHostname("https", node.HTTPSAddress)
-					if url != "" {
-						if c.snifferCallback(node) {
-							nodes = append(nodes, newConn(nodeID, url))
-						}
-					}
-				}
-			default:
-				for nodeID, node := range info.Nodes {
-					url := c.extractHostname("http", node.HTTPAddress)
-					if url != "" {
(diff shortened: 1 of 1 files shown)


olivere:
Thanks for your support. I wasn't aware of Search Guard, so I haven't tested that.

The logic for finding the correct endpoint URL is something that ES has changed several times, IIRC even between minor versions. That's why the logic differs between e.g. v3 and v5.

Let me dive into that topic when I have some more time. I'm currently very busy at my day job. Sorry.

eyeamera:
Thanks, look forward to seeing what you find. 

Just for reference, I'm running ES v2.4.3 w/ SearchGuard v2.4.3.20

olivere:
This change has landed in 3.0.71. Thanks for being a contributor to Elastic and open source in general.

Issue #497, opened 2017-03-28
Refactor Suggesters


olivere:
Suggesters are not in line with Elasticsearch 5.x. The different suggesters return different options, and e.g. phrase suggesters use a `score` field for the score while completion suggesters use `_score`. We probably need to have different `Options` structs for the different suggesters.

See also #493

sinnawat:
It's also had a problem with phrase suggesters => "highlighed" when I use ```
"highlight": {
          "post_tag": "<em>",
          "pre_tag": "</em>"
        }
```

**Comment by olivere** [comment]
@sinnawat I will look into this. Maybe you can write a test that illustrates the problem?

**Comment by sinnawat** [comment]
Quick write test

```
package main

import (
	"context"
	"encoding/json"
	"fmt"
	"testing"

	elastic "gopkg.in/olivere/elastic.v5"
)

func TestSearchSuggest(t *testing.T) {
	// Connect
	client, err := elastic.NewClient(
		elastic.SetSniff(false),
		elastic.SetURL("http://localhost:9201"))
	if err != nil {
		fmt.Println(err)
	}

	// createIndex
	mapping := `{
			"settings":{
				"number_of_shards":1,
				"number_of_replicas":0
			},
			"mappings":{
				"tweet":{
					"properties":{
						"tags":{
							"type":"string"
						}
					}
				}
			}
		}`

	if isExists, _ := client.IndexExists("twitter").Do(context.Background()); !isExists {
		_, err := client.CreateIndex("twitter").BodyString(mapping).Do(context.Background())
		if err != nil {
			// Handle error
			panic(err)
		}
	}

	// Insert tweet
	for i := 0; i < 10; i++ {
		_, _ = client.Index().
			Index("twitter").
			Type("tweet").
			BodyString(`{"tags" : "cat"}`).
			Do(context.Background())
	}

	// Flush
	_, _ = client.Flush().Index("twitter").Do(context.Background())

	// Phrase Suggest Query
	field := "tags"
	suggesterQuery := elastic.NewPhraseSuggester("phrase-suggest").
		Text("cad").
		Field(field).
		Size(4).
		Confidence(0.7).
		MaxErrors(1).
		RealWordErrorLikelihood(1).
		CandidateGenerator(elastic.NewDirectCandidateGenerator(field).
			SuggestMode("always").
			MinWordLength(2)).
		Highlight("<em>", "</em>")

	// Search
	suggestResult, _ := client.Search().
		Index("twitter").
[...]

PR #597, opened 2017-09-11
solve incorrect response field for search suggestion

misterciput:
Can not get score and frequency of suggestion when using search suggestion.
Caused by parsing wrong json field. We see that in elasticsearch raw search suggestion option response will be like this:

`"options": [
    {
           "text": "shoes",
           "score": 0.8,
           "freq": 34084
     }
 ]
`
and the code parse "_score" and did not parse "freq"
This PR fix this by change the field we parse into correct field


olivere:
Thank you. This will of course be fixed in the next version. I will look the structures up in the Java code. Maybe there's some other fields missing / renamed.

olivere:
I did have a look at this and added/modified the response in accordance with [Suggest.Option](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/search/suggest/Suggest.java#L621). I couldn't find a `freq` field there. I know the [documentation page](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-suggesters.html) lists it, but it doesn't seem to be in the code any more. Did you really get that with a query in 5.6.3?

Anyway, I will push 5.0.49 soon.

Thanks for being a contributor to Elastic and open source in general. 👍

misterciput:
Sure, I really get the `freq` field when I do terms suggestion query like this

```
{
  "suggest" : {
	"suggestion": {
		"text": "shues",
		"term": {
			"analyzer": "whitespace",
			"field": "ad_title",
			"min_doc_freq": 1,
			"size": 1,
			"suggest_mode": "popular"
		}
	}
  }
}
```

I think the code is not in [Suggest.Option](https://github.com/elastic/elasticsearch/blob/5.5/core/src/main/java/org/elasticsearch/search/suggest/Suggest.java#L621) but it still exist in [TermSuggestion.Option](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/search/suggest/term/TermSuggestion.java#L178), maybe you can check about this. It should be related to https://github.com/olivere/elastic/issues/497 that every suggestion (terms, phrase, and completion) has different response. I still find a best way to refactor each type of suggestion into a better one.

It is a great honor to contribute in Elastic

olivere:
@misterciput Ah, I see. Yes, Suggesters need to be refactored to do this cleanly. In the meantime, I should simply add this field and aggregate all all possible returns values.

misterciput:
@olivere Yes that should be useful for this time. Still didn't know why elastic have different response in each suggestion like in using `_score` in completion suggester but use `score` in phrase and terms suggester.

olivere:
@misterciput You mean Elastic has that, or Elasticsearch?

misterciput:
Sorry I mean Elasticsearch

olivere:
@misterciput Yeah. Elastic, this library, had that name even before Elastic, the company, was established :-)

It's quite hard to manage such a huge API. Things always drift apart, so I fully understand why that happens from time to time. But they usually fix it in new releases.

misterciput:
@olivere amazing :-)

Yeah, But it is some kind of fun to have an amazing API like this, a lot of better than golang API from elasticsearch itself.

It will be great if I can see the updated `freq` field in the next release, I have tested it by this commit and no something weird happen.

olivere:
Yeah. I will push 5.0.52 in a minute.

misterciput:
@olivere great, thanks a lot

PR #620, opened 2017-10-13
 Switch Elastic v3 from golang.org/x/net/context to stdlib context

### LINKED ISSUE #547: v3: Switch to stdlib context
The elastic.v3 package still has references to `golang.org/x/net/context` (e.g. [client.go](https://github.com/olivere/elastic/blob/release-branch.v3/client.go)). We should switch them to `context` from stdlib, just as we've already done with v2 and v5.

liketic:
Closes #547

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -6,11 +6,10 @@ package elastic
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"net/url"
-
-	"context"
 
 	"gopkg.in/olivere/elastic.v3/uritemplates"
 )
--- a/bulk.go
+++ b/bulk.go
@@ -10,7 +10,7 @@ import (
 	"fmt"
 	"net/url"
 
-	"golang.org/x/net/context"
+	"context"
 
 	"gopkg.in/olivere/elastic.v3/uritemplates"
 )
--- a/client.go
+++ b/client.go
@@ -1199,6 +1199,9 @@ func (c *Client) PerformRequestC(ctx context.Context, method, path string, param
 		c.dumpRequest((*http.Request)(req))
 
 		// Get response
+		if ctx == nil {
+			ctx = context.Background()
+		}
 		res, err := c.c.Do((*http.Request)(req).WithContext(ctx))
 		if err == context.Canceled || err == context.DeadlineExceeded {
 			// Proceed, but don't mark the node as dead
--- a/client.go
+++ b/client.go
@@ -6,6 +6,7 @@ package elastic
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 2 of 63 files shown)


bulk.go
```diff
@@ -10,7 +10,7 @@ import (
 	"fmt"
 	"net/url"
 
-	"golang.org/x/net/context"
+	"context"
```
olivere: Can you please group the imports into stdlib / external / internal imports. Tools like `goimports` should help you with that:

Instead of ...

```
import (
	"bytes"
	"errors"
	"fmt"
	"net/url"

	"context"

	"gopkg.in/olivere/elastic.v3/uritemplates"
)
```

... it should be ...

```
import (
	"bytes"
	"context"
	"errors"
	"fmt"
	"net/url"

	"gopkg.in/olivere/elastic.v3/uritemplates"
)
```

client.go
```diff
@@ -1208,7 +1207,7 @@ func (c *Client) PerformRequestC(ctx context.Context, method, path string, param
 		if ctx == nil {
```
olivere: v5 [doesn't have that `ctx == nil`](https://github.com/olivere/elastic/blob/release-branch.v5/client.go#L1253). We need to review if we can use the same `PerformRequest` as v5 in v3.

bulk.go
liketic: Thanks for reviewing. Seems like `goimports ` won't help for those kind imports. Actually, I did run `goimports` before I commit. Now I group them manually. Could you please review again?

client.go
```diff
@@ -1123,9 +1122,6 @@ func (c *Client) PerformRequest(method, path string, params url.Values, body int
 // Optionally, a list of HTTP error codes to ignore can be passed.
 // This is necessary for services that expect e.g. HTTP status 404 as a
 // valid outcome (Exists, IndicesExists, IndicesTypeExists).
-//
-// If ctx is not nil, it uses the ctxhttp to do the request,
-// enabling both request cancelation as well as timeout.
 func (c *Client) PerformRequestC(ctx context.Context, method, path string, params url.Values, body interface{}, ignoreErrors ...int) (*Response, error) {
```
olivere: Did you double-check that no code calls `PerformRequestC` with `ctx` set to `nil`?

olivere: Seems like you didn't?! https://travis-ci.org/olivere/elastic/jobs/287568213#L751

olivere: I think a possible solution would be to check for `ctx == nil` in `PerformRequestC` and initialize it with `context.Background()` then.

bulk.go
olivere: With `-local`, it should [group the imports](https://stackoverflow.com/questions/38704769/golang-import-grouping-by-package), no?

olivere:
LGTM. I will look again over the weekend and prepare a new v3 release if it's still looking good.

Thanks for being a contributor and spending time on open source.

liketic:
Thanks. 😅

Issue #510, opened 2017-04-13
Wrap errors so that details are more visible


olivere:
Use e.g. [github.com/pkg/errors](https://github.com/pkg/errors) to wrap certain errors and make them more visible.

The goal is to make the underlying reason for certain errors more transparent, e.g. `elastic.ErrNoClient`.

eticzon:
@olivere if you don't mind, I'd like to get cracking on this one next. What do you think should the strategy here be? Should we do it in phase or one big bang? 

Also we might want to consider drafting a list of "underlying reasons" or "friendly messages" for certain error classes. I'm happy to draft one up for your review.

olivere:
@eticzon Thanks for your helping hand.

It'd be awesome if we could start with those `elastic.ErrNoClient` errors. That's probably the no. 1 issue that people have issues with. If we could provide a useful error message for each of the underlying problem for each one of those case, that'd help a lot (just search for "No ElasticSearch Node Available" in the issues).

Right now, elastic is swalling most of the underlying errors. It could probably be as simple as wrapping the underlying error via `github.com/pkg/errors`.

eticzon:
Thanks for the quick reply. 👍  Ok I'll start with those `elastic.ErrNoClient` ones first.

olivere:
I also added the `help appreciated` label so you can quickly decide what'd be interesting for you (or anyone willing to spend some time to support this little library of ours).

eticzon:
+1! Thanks for that.

eticzon:
I've opened #518 to get the ball rolling. 

Let me know if this shallow `errors.Wrap` is good for now and if you think we need to add more tests to check that we're correctly adding detail to every single error scenario.

olivere:
I like the error messages. Not too chatty and to the point. 👍 

However, the wrapping could become a problem in that existing clients can no longer compare `err == elastic.ErrNoClient`. I mean, they can, but it won't be the same as before. This test will e.g. fail now:

```
func TestClientHealthcheckStartupTimeout(t *testing.T) {
	_, err := elastic.NewClient(
		elastic.SetURL("http://localhost:9299"),
		elastic.SetHealthcheckTimeoutStartup(5*time.Second),
	)
	if err != elastic.ErrNoClient {
		t.Fatal(err)
	}
}
```

... because it will directly compare the error with the `elastic.ErrNoClient`. As it's wrapped now, it will not be equal to `elastic.ErrNoClient` while before it was!

[...]

PR #634, opened 2017-11-13
Don't ignore error while doing startup check

### LINKED ISSUE #625: Client logs non-obvious error on failure to verify https endpoint
[x] elastic.v5 (for Elasticsearch 5.x)

Receive an error message related to the issue

Receive an error message: `"no Elasticsearch node available"`

Have elastic behind a reverse proxy. Run the elastic client in an environment without ca-certificates

tanner-bruce:
Error is checked for if it is nil in Client#startupHealthcheck, but not for if the error actually contains an error.

Fixes #625 and will likely give a better message for basic auth failures as well

This code:
```
package main

import (
	"log"
	"os"

	"gopkg.in/olivere/elastic.v5"
)

func main() {
	host := os.Getenv("ELS_HOST")
	user := os.Getenv("ELS_USER")
	pass := os.Getenv("ELS_PASS")

	_, err := elastic.NewClient(
		elastic.SetURL(host),
		elastic.SetSniff(false),
		elastic.SetBasicAuth(user, pass),
	)
	if err != nil {
		log.Fatalf("could no create elastic client to %q: %v", host, err.Error())
	}
}
```

and this Dockerfile
```
FROM bitnami/minideb
#RUN install_packages ca-certificates
ADD elastic ./
CMD ["./elastic"]
```

showcase the difference this PR makes.

e.g

Instead of receiving the message 
```
health check timeout: no Elasticsearch node available
```
on client creation, we will now receive [in the case where the underlying system has no ca-certificates]
```
Head https://domain.net: x509: failed to load system roots and no roots provided
```

which will be nice for UX when debugging issues.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1104,6 +1104,8 @@ func (c *Client) startupHealthcheck(timeout time.Duration) error {
 			res, err := cl.Do(req)
 			if err == nil && res != nil && res.StatusCode >= 200 && res.StatusCode < 300 {
 				return nil
+			} else if err != nil {
+				return err
 			}
 		}
 		time.Sleep(1 * time.Second)


tanner-bruce:
One thing to note, the general approach of this is worth a discussion, as well as how it affects testing, due to how it affects multiple urls being passed to the startup healthcheck. A possible solution would be to aggregate any errors in the `for`, and return that. 

At that point we'd likely need a separate Error type, a 'soft' error of sorts that essentially says 'one of your urls was broken, but the others work'

Once that is figured out I can fix up the tests/approach as needed.

olivere:
@tanner-bruce  Better errors codes are very welcome, of course. In #510 we introduced [github.com/pkg/errors](https://github.com/pkg/errors) in the code base with `errors.Wrap` / `errors.Wrapf` to add context to an error. I think we should do that here as well.

Regarding your last comment: I agree that errors could be improved even more. But your code will enhance error messages, so let's just start with that.

olivere:
@tanner-bruce This PR is not the correct solution to pass error information back to the caller. A health check checks if any node in the list is _available_. Your PR makes health check to return with an error if any node in the list is _unavailable_.

I'm working on a different solution to this.

olivere:
[This commit](https://github.com/olivere/elastic/commit/41b733956ba2a0ec296b3e19e28dce2cf016d70d) should now print a more helpful error message like `health check timeout: Head http://localhost:9299: dial tcp 127.0.0.1:9299: getsockopt: connection refused: no Elasticsearch node available`. It prints the last error found while trying to connect to any of the nodes. It is available in v5 and v6.

tanner-bruce:
Hi @olivere, I mentioned mine wasn't correct and was looking for feedback on the best way to move forward. I actually already wrote nearly the same code as your commit, I just wasn't sure how you wanted to deal with the case of >1 ES endpoint where one of them fails to connect.

PR #642, opened 2017-11-17
add terminate_after to params

hcliff:
Add support for terminate_after, as specified in 

https://www.elastic.co/guide/en/elasticsearch/reference/5.4/search-count.html

### DIFF
--- a/count.go
+++ b/count.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


olivere:
Yep. Thank you. That's looking good. Will try to fix it in the next release.

olivere:
It is in 5.0.58+ and the current release-branch.v6, which will become 6.0.0 eventually.

ArcticSnowman:
@olivere  This param seems to be present in a number of other calls like '_search/'

Issue #574, opened 2017-08-03
No support for boundary_scanner in Highlight


nan0tube:
### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[x] elastic.v5 (for Elasticsearch 5.x)

### Please describe the expected behavior

ES5 has 3 types of [`boundary_scanner`](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-highlighting.html#boundary-scanners): `chars`, `sentence`, and `word`. This sets the boundary scanner, and expands its functionality beyond the currently supported `boundary_chars`. 

### Please describe the actual behavior

At present, there is no way to set `boundary_scanner`.

olivere:
Thanks for your report. I will dig into this.

olivere:
Sorry that it took so long. But #645 fixes this. It is available in 5.0.57+.

PR #645, opened 2017-11-20
Add in boundary_scanner to highlighter

jimberlage:
This adds in the boundary_scanner option to the elastic.Highlighter
object.  I based my changes off of release-branch.v5 for selfish
reasons; I need this feature for an Elasticsearch 5.6.4 cluster.

Java API:
https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/SearchContextHighlight.java#L277
Elasticsearch docs:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-highlighting.html#highlighting-settings

### DIFF
--- a/highlight.go
+++ b/highlight.go
@@ -20,6 +20,7 @@ type Highlight struct {
 	requireFieldMatch     *bool
 	boundaryMaxScan       *int
 	boundaryChars         *string
+	boundaryScannerType   *string
 	highlighterType       *string
 	fragmenter            *string
 	highlightQuery        Query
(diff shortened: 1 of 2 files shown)


olivere:
Thanks for being a contributor.

There are some other settings missing as well (if I see that correctly). So let me keep this long-standing issue open (#574) until I checked and reviewed all the settings now being used in v5 and v6.

olivere:
This will be included in 5.0.57 and the current release-branch.v6, which will eventually become 6.0.0.

PR #649, opened 2017-11-21
Add TrackScores to SearchService

FdeFabricio:
SearchService was missing the property trackScores

### DIFF
--- a/search.go
+++ b/search.go
@@ -31,6 +31,7 @@ type SearchService struct {
 	ignoreUnavailable *bool
 	allowNoIndices    *bool
 	expandWildcards   string
+	trackScores 	  bool
 }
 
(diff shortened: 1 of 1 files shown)


olivere:
Thanks for being a contributor.

You will get that setting, but other settings might have changed as well. So I will keep this PR open and use it as a reminder to check and refresh all the different settings available now in the [Java API](https://github.com/elastic/elasticsearch/blob/6.0/core/src/main/java/org/elasticsearch/search/builder/SearchSourceBuilder.java).

FdeFabricio:
Thanks. Please let me know once it is included in a release.

olivere:
This will be included in 5.0.57 and the currentl release-branch.v6 (which will become 6.0.0 eventually).

Issue #646, opened 2017-11-20
Customizable retry for BulkProcessorService


rwynn:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[X] elastic.v5 (for Elasticsearch 5.x)
[ ] elastic.v6 (for Elasticsearch 6.x)

### Please describe the expected behavior
Methods are available on the BulkProcessorService to override the default initialTimeout, maxTimeout, or set the retrier/policy that is used.  Alternatively, the BulkProcessorService uses the associated client's retrier which is customizable. 

### Please describe the actual behavior
Unless I am missing something, the BulkProcessorService does not expose a way to set the initialTimeout or maxTimeout for retries.  There is also no way to opt into a retrier or policy different than ExponentialBackoff.  This makes it impossible to opt out or customize retry for bulk.  Finally, the BulkProcessorService has access to the client's retrier but does not use it.  

### Any steps to reproduce the behavior?
N/A

olivere:
Yes. I think #610 is rather similar. I think both BulkProcessor as well as Bulk API and Scroll API need to be able to configure their own custom retrier.

There are [different Retrier implementations](https://github.com/olivere/elastic/wiki/Retrier-and-Backoff) though, with configurable initial timeout and a maximum timeout as well. Plus, you can simply create your own retrier implementation. However, Elastic uses the same Retrier for all API calls. But retries may e.g. be dangerous for scrolling as #610 correctly points out.

rwynn:
Yes, I am aware of the different retrier implementations and that I can create my own.  I'm just not able to set one or even override the timeouts on BulkProcessor due to the fields being private (and no public methods to set them - there is Before, After, Name, Workers, etc but not InitialTimeout or MaxTimeout).  I think you may be agreeing with me, but I'm not sure with your 2nd paragrah.

By the way, Elastic is an amazing library.  You are doing awesome work here and I love the way Elastic is designed.

[...]

PR #650, opened 2017-11-21
customizable backoff for issue #646

rwynn:
The behavior should be the same as before this commit, but this enables one to set a custom backoff policy.  A Stopbackoff e.g. would prevent retry.


olivere:
@rwynn That's looking good. I'll put this on the list of final fixes for the 6.0 release. Thanks for supporting the work on elastic.

PR #652, opened 2017-11-23
fix problems in search.go:Each when hits have 'nil' _Source field

asac:
_source can be nil in hits in my elastic it seems. This PR proposes to be more resilient to such entries and just skip them.

### DIFF
--- a/search.go
+++ b/search.go
@@ -427,7 +427,8 @@ func (r *SearchResult) TotalHits() int64 {
 
 // Each is a utility function to iterate over all hits. It saves you from
 // checking for nil values. Notice that Each will ignore errors in
-// serializing JSON.
+// serializing JSON and hits with empty/nil _source will get an empty
+// value
 func (r *SearchResult) Each(typ reflect.Type) []interface{} {
 	if r.Hits == nil || r.Hits.Hits == nil || len(r.Hits.Hits) == 0 {
 		return nil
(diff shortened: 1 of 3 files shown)


olivere:
Hmm... I'm wondering if it may be better to return a `nil` hit as `nil` in the slice returned. Otherwise you may get less elements in the slice than there are hits, and that may lead to other problems.

asac:
let me try that...

asac:
how does this one feel?

search.go
```diff
@@ -427,14 +427,19 @@ func (r *SearchResult) TotalHits() int64 {
...
+// value
 func (r *SearchResult) Each(typ reflect.Type) []interface{} {
 	if r.Hits == nil || r.Hits.Hits == nil || len(r.Hits.Hits) == 0 {
 		return nil
 	}
 	var slice []interface{}
 	for _, hit := range r.Hits.Hits {
 		v := reflect.New(typ).Elem()
+		if hit.Source == nil {
+			slice = append(slice, v.Interface())
```
olivere: Let me see. Without testing: This would create an empty element of the given type. I think we need to add a unit test for this, so that we not only know that it works as expected but also guide people what happens when `_source == nil`. Something below [this](https://github.com/olivere/elastic/blob/release-branch.v6/search_test.go#L137-L197), which simply calls `Each` on a `SearchResult` with gaps / nil values. You need to test that for pointer and value types, i.e. `tweet{}` and `*tweet{}`.

asac: ok added test too... let me know!

olivere:
Looking good. Thanks!

olivere:
This will be available in 5.0.57 and current release-branch.v6 (which will be released as 6.0.0 eventually).

PR #662, opened 2017-12-14
update remove_on_conflict param

kmulvey:
https://www.elastic.co/guide/en/elasticsearch/reference/6.1/docs-bulk.html#bulk-update

### DIFF
--- a/bulk_index_request.go
+++ b/bulk_index_request.go
@@ -183,7 +183,7 @@ func (r *BulkIndexRequest) Source() ([]string, error) {
 		indexCommand["_version_type"] = r.versionType
 	}
 	if r.retryOnConflict != nil {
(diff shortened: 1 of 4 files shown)


olivere:
Hmm... it's [`retry_on_conflict` for ES 6.1](https://www.elastic.co/guide/en/elasticsearch/reference/6.1/docs-bulk.html#bulk-update) and [`_retry_on_conflict` for ES 6.0](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/docs-bulk.html#bulk-update). That's a problem from a compatibility perspective. I will have to review if any of the two works for both versions. We also need to fix the tests as well.

olivere:
I settled with `retry_on_conflict` for elastic version 6.1.0 and later, which will be tagged later today.

PR #680, opened 2018-01-17
Ran gofmt across library

jdelgad:


olivere:
Did so as well after some changes. Thanks.

PR #689, opened 2018-01-19
Add TimeZone to DateRangeAggregation

higebu:
See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-daterange-aggregation.html#time-zones


olivere:
Yep. That seems to be missing. Thanks for reporting. Will add it in the next release.

olivere:
Thanks.

PR #696, opened 2018-01-26
Add support for Composite Aggregation

snowzach:
This PR adds support for the new CompositeAggregation.

### DIFF
--- a/search_aggs_bucket_composite.go
+++ b/search_aggs_bucket_composite.go
@@ -7,23 +7,25 @@ package elastic
 // CompositeAggregation is a multi-bucket values source based aggregation
 // that can be used to calculate unique composite values from source documents
 //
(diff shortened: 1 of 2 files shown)


olivere:
Thanks for contributing to elastic 👍   I will review and merge this in the next release.

olivere:
Turns out your solution wasn't quite complete. While you were implementing the request side, the response side was missing, i.e. how to extract the composite aggregation response from the e.g. search response.

I also changed the structure of the code a bit and aligned with the Java source, which [you can find here](https://github.com/elastic/elasticsearch/tree/master/server/src/main/java/org/elasticsearch/search/aggregations/bucket/composite).

If all works as expected, this will be part of the next release. Feel free to review the code.

PR #701, opened 2018-01-31
add cat api implement.

soarpenguin:


olivere:
The `/_cat` endpoints are not regarded as part of the REST API and therefore not supported. They are meant to be read by humans, and are not parseable as JSON. You can get most of these numbers, stats, and information from using the REST API, e.g. Cluster Stats or Cluster Health.

olivere:
Sorry. This wasn't meant to turn you down as a contributor, of course. Thanks so much for investing your time into open source!

PR #710, opened 2018-02-16
Add Relation to RangeQuery

plopik:
https://www.elastic.co/guide/en/elasticsearch/reference/6.2/range.html Thank you


olivere:
That's useful, and has existed since 5.2. Didn't know that. Thanks. Will be in the next release, but I will have to backport to elastic.v5 as well.

olivere:
Available in 6.1.8 and 5.0.65.

olivere:
Thanks!

PR #713, opened 2018-02-20
Add ability to override request body for the 'next' scroll calls

ArcticSnowman:
Add ability to override request body for the 'next' scroll calls

Also adds the 'scroll' param KeepAlive to the request

### DIFF
--- a/scroll.go
+++ b/scroll.go
@@ -458,12 +458,19 @@ func (s *ScrollService) buildNextURL() (string, url.Values, error) {
 // body returns the request to fetch the next batch of results.
 func (s *ScrollService) bodyNext() (interface{}, error) {
 	s.mu.RLock()
-	body := struct {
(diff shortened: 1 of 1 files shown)


olivere:
What is the use case for this?

ArcticSnowman:
Allows me to use the v5 SDK with the ES 1.7...    as ES < 2.x expect just the scroll id as the body..
Also 1.7/2.4/5.6 accept the scroll id in the request, not just in the request body.

olivere:
I'm sorry but I won't give up compatibility on a service-to-service basis. Sniffing with v2 and v5 are also different, so where should we draw the line?

If you want to keep a single dependency to elastic, you have plenty of options. E.g. keep your own fork of elastic or extract the Scroll service into your own code and modify as required for your cluster(s).

PR #714, opened 2018-02-20
Add terms to add terms to query

ArcticSnowman:
This allow a user to loop through an []string array and add the values to a filter


olivere:
You could achieve the same with a loop before calling `NewTermsQuery`:

```
c := make([]interface{}, len(in))
for i, s := range in {
  c[i] = s
}
q := elastic.NewTermsQuery("field", c...)
```

ArcticSnowman:
I have a '[]string' but I missed that appending '...'  to the variable treats the array as the variadic instead of passing the array itself..

PR #722, opened 2018-03-09
SearchShards API added + tests

Zyqsempai:
Added SearchShards API functionality + minimum tests.
https://www.elastic.co/guide/en/elasticsearch/reference/6.0/search-shards.html

### DIFF
--- a/search_shards.go
+++ b/search_shards.go
@@ -13,8 +13,7 @@ import (
 	"github.com/olivere/elastic/uritemplates"
 )
 
-// SearchShardsService computes a score explanation for a query and
-// a specific document.
+// SearchShardsService returns the indices and shards that a search request would be executed against.
 // See https://www.elastic.co/guide/en/elasticsearch/reference/6.0/search-shards.html
 type SearchShardsService struct {
 	client            *Client
--- a/search_shards.go
+++ b/search_shards.go
@@ -7,21 +7,25 @@ package elastic
 import (
 	"context"
 	"fmt"
-	"github.com/olivere/elastic/uritemplates"
 	"net/url"
 	"strings"
+
+	"github.com/olivere/elastic/uritemplates"
 )
 
 // SearchShardsService computes a score explanation for a query and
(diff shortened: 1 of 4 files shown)


search_shards.go
```diff
@@ -0,0 +1,153 @@
+// Copyright 2012-present Oliver Eilhard. All rights reserved.
+// Use of this source code is governed by a MIT-license.
+// See http://olivere.mit-license.org/license.txt for details.
+
+package elastic
+
+import (
+	"context"
+	"fmt"
+	"github.com/olivere/elastic/uritemplates"
```
olivere: Can you put that into a separate block, just like in the other services, please?

search_shards.go
```diff
@@ -0,0 +1,153 @@
...
+// NewSearchShardsService creates a new SearchShardsService.
+func NewSearchShardsService(client *Client) *SearchShardsService {
+	return &SearchShardsService{
+		client: client,
+	}
+}
+
+// Index sets the names of the indices to restrict the results.
+func (s *SearchShardsService) Index(index ...string) *SearchShardsService {
+	if s.index == nil {
```
olivere: I think the `nil` check is not necessary. It was a misconception I had when I learned Go years ago :-)

olivere:
Hey, thanks for that. Just two minor issues, but I think we can merge it after your fixes.

olivere:
Oh, can you please also add the other parameters (inherited from most services) like `ignore_unavailable` etc. See the [REST API Spec for Search Shards](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/search_shards.json).

Zyqsempai:
Hi Olivere, all done, and also a have in plans to fill up repo with missing APIs in near feature, so if you don't mind, expect few PR on this week.
By the way, thank you for the great library.

olivere:
Sure. Very welcome.

Issue #730, opened 2018-03-14
Bucket Sort Aggregation not available for elastic.v6


mmfrb:
### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[x] elastic.v6 (for Elasticsearch 6.x)

### Please describe the expected behavior
I'd like to use the brand new [Bucket Sort Aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/6.2/search-aggregations-pipeline-bucket-sort-aggregation.html)  through the elastic client.

### Please describe the actual behavior
Apparently this is not implemented yet nor will be released in the next release (at least I could not find it in the files of the branch).

This is more of a question of whether this is already being implemented. I'd be more than happy to implement it myself, and if that's the case, some instructions on how to implement a new pipeline aggregation would be much appreciated.

Thanks in advance

olivere:
Yes, it looks like that's still missing.

If you have the time and commitment to become a contributor, here's a few tips. Aggregations have two sides: The query-side, which is what you find in e.g. `search_aggs_pipeline_avg_bucket.go` (and its test file), and the response-side, which extracts the aggregated data from the response (you find that in `search_aggs.go` and its test file). We've recently added a few missing aggregation types, so that should give you a blueprint of what it takes to add a new aggregation type (see e.g. [this PR](https://github.com/olivere/elastic/pull/725)).

Otherwise, don't worry. I'll add it to the backlog and will add it in one of the next releases.

mmfrb:
I have the pull request ready. Could you please give me permission to push my branch into the repo?

olivere:
@mmfrb That's not how it works. You need to fork this repository first, then work on your fork, and finally send me a Pull Request to review. Once it's looking good, I will then merge it into this repository.

See [here](https://help.github.com/articles/creating-a-pull-request-from-a-fork/) for details. This is also quite a [good reference](https://gist.github.com/Chaser324/ce0505fbed06b947d962).

olivere:
Merged in #732, thank you. It will be available in the next release of v6.

PR #732, opened 2018-03-15
Bucket sort aggregation

mmfrb:
#730 

This PR adds the Bucket Sort pipeline aggregation to elastic client.

Few important things to notice:
1) As pipeline aggregations [do not support subaggregations ](https://www.elastic.co/guide/en/elasticsearch/reference/6.2/search-aggregations-pipeline.html#search-aggregations-pipeline), there's no subAggregations field in the new struct.
2) Metadata, although supported by Elasticsearch for this aggregation, is not used, since this aggregation doesn't add data to the response. It only reorders parent buckets.
3) No response parser was made for this aggregation since, like said above, this aggregation does not add data to the elasticsearch response, so there's no new data to parse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ There are also [some recipes](https://github.com/olivere/elastic/tree/release-br
   - [x] Cumulative Sum
   - [x] Bucket Script
   - [x] Bucket Selector
-  - [ ] Bucket Sort
+  - [x] Bucket Sort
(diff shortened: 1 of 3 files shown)


olivere:
Thanks for supporting Elastic as a contributor. ❤️ I will review ASAP and let you know when it will be released.

PR #733, opened 2018-03-15
New Xpack functionality implementation start:

Zyqsempai:
https://www.elastic.co/guide/en/elasticsearch/reference/6.2/xpack-api.html
This PR contains WATCHER API implementation.
https://www.elastic.co/guide/en/elasticsearch/reference/6.2/watcher-api.html


Zyqsempai:
So, here is the Wather API implementation, for start.
Points to discuss:
1) Test - before elastic 6.3 release, we have no way to test them, for now there are just dummy tests.
2) Request body - we have few options, simple String/Json Body without validation, or we can use Structures to control bodies.
3) I found some mismatch in ES xpack documentation and REST specification, there are possibility that specification is little bit out of date.
What is your thoughts on this.

olivere:
I will look into the code over the weekend, thank you.

Regarding your questions:

1. Are you not sure we can't test? We should probably install [the platinum package via Docker](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html)? You should be able to edit the Docker image (use `elasticsearch-platinum:6.2.2`) in both `run-es.sh` and `.travis.yml`. With the platinum image, we should be able to use a 30-day free trial of all of X-Pack.

2. I typically look at the Java code for this, not the online documentation. Code doesn't lie. However, we don't have any way to look at the code before 6.3 is released. Setting up the request body, we should follow the rest of the services and use the builder pattern that everyone is used to when using Elastic. As for the response, we should also deserialize into a struct, at least for the majority of fields.

3. Yes, we can only be sure when we can look into the source code. Before that, we can only do our best and wait for bug reports hitting us ;-)

Notice also that [not all of X-Pack will become free](https://www.elastic.co/products/x-pack/open): [See the subscriptions](https://www.elastic.co/subscriptions).

xpack_watcher_ack_watch.go
```diff
@@ -0,0 +1,149 @@
...
+	var body interface{}
+	if s.bodyJson != nil {
+		body = s.bodyJson
+	} else {
+		body = s.bodyString
+	}
+
+	// Get HTTP response
+	res, err := s.client.PerformRequest(ctx, PerformRequestOptions{
+		Method: "",
```
olivere: The method should be `PUT`, right?

xpack_watcher_delete_watch_test.go
```diff
@@ -0,0 +1,24 @@
...
+// See http://olivere.mit-license.org/license.txt for details.
+
+package elastic
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+func TestWatcherdeleteWatch(t *testing.T) {
```
olivere: This should be `TestWatcherDeleteWatch` with a capital `D`.

olivere:
Now I see what you mean with the 2nd question--setting up the request body. We can, of course, start with asking people to pass the request body with `BodyJson` and/or `BodyString`. This makes sense if the request body isn't structure, like e.g. when creating an index. If the request body is structured (and we know how it really looks like), we should probably create a struct/builder-to-set-up-a-struct and allow it to be passed to the X-Pack service. This is what I did with [`SearchSource`](https://github.com/olivere/elastic/blob/release-branch.v6/search_source.go) in the [SearchService](https://github.com/olivere/elastic/blob/release-branch.v6/search.go): You can [either pass it into the service directly](https://github.com/olivere/elastic/blob/release-branch.v6/search.go#L45-L52), or  [the service passes it into `SearchSource`](https://github.com/olivere/elastic/blob/release-branch.v6/search.go#L89).

Zyqsempai:
oh, sorry for typos, fixed.
Tomorrow i will check the elasticsearch-platinum:6.2.2 image, and try to use it for tests.
Thank you for your advices.

Zyqsempai:
@olivere Hi there, fortunately, i found some time on the weekend to add some tests with new docker image, also a detailed response structures for APIs with complex responses.

xpack_test.go
```diff
@@ -0,0 +1,146 @@
+package elastic
+
+import (
+	"context"
+	"testing"
+)
+
+func TestWatchWorkFlow(t *testing.T) {
```
olivere: Can we use an `Xpack` or `XPack` prefix for the tests, please? That makes testing the X-Pack features easier as we can run `go test -run=XPack` and be sure that we hit all X-Pack-related tests.

xpack_watcher_ack_watch.go
```diff
@@ -145,5 +145,14 @@ func (s *XpackWatcherAckWatchService) Do(ctx context.Context) (*XpackWatcherAckW
...
+	Status AckWatchStatus `json:"status"`
+}
+
+type AckWatchStatus struct {
+	State            map[string]interface{}            `json:"state"`
+	LastChecked      string                            `json:"last_checked"`
+	LastMetCondition string                            `json:"last_met_condition"`
+	Actions          map[string]map[string]interface{} `json:"actions"`
+	ExecutionState   string                            `json:"execution_state"`
+	Version          int                               `json:"version"`
```
olivere: `Version` is hard: We have to check here in the Java code whether it is `int` or `int64`.

xpack_watcher_get_watch.go
```diff
@@ -102,8 +102,21 @@ func (s *XpackWatcherGetWatchService) Do(ctx context.Context) (*XpackWatcherGetW
...
+	Found  bool        `json:"found"`
+	Id     string      `json:"_id"`
+	Status WatchStatus `json:"status"`
+	Watch  Watch       `json:"watch"`
+}
+
+type WatchStatus struct {
+	State   map[string]interface{}            `json:"state"`
+	Actions map[string]map[string]interface{} `json:"actions"`
+	Version int                               `json:"version"`
```
olivere: Same comment wrt `Version`.

olivere:
Thanks. I just left a few comments. We also need to fix the Travis test.

Zyqsempai:
OK, i will fix everything today.

Zyqsempai:
About version: Found something like that in .NET code, so i suppose it's, INT
https://github.com/elastic/elasticsearch-watcher-net/blob/master/src/Nest.Watcher/Domain/WatchStatus.cs#L13

olivere:
I've tried to simplify testing with both the OSS as well as the Platinum version in my `xpack` branch.

We now have two ES containers running. Connect to `http://localhost:9200` for the OSS version (without X-Pack Security or anything like that) and use `http://elastic:elastic@localhost:9210` for the Platinum version.

You should also be able to spin up both containers locally by running `docker-compose up -d`.

Unfortunately, this causes conflicts with your PR. But I hope the advantages will outweigh the disadvantages.

Zyqsempai:
Thanks, that works, all tests fixed.

setup_test.go
```diff
@@ -276,6 +276,7 @@ type logger interface {
 
 func setupTestClient(t logger, options ...ClientOptionFunc) (client *Client) {
 	var err error
+	options = append(options, SetBasicAuth("elastic", "elastic"))
```
olivere: I think we need to remove that line again, now that we have both versions via Docker Compose. I wonder why that didn't cause any problems in the tests.

Zyqsempai: Ok, i will remove this today, but this line didn't caused the problems, because it's just additional headers for request, and plain OSS Elastic just ignored them.

Zyqsempai:
Done.

olivere:
Note to self: The x-pack source is available now: https://github.com/elastic/elasticsearch/tree/master/x-pack

Zyqsempai:
Cool, i am back from vacation, so i will be glad to continue working on x-pack API.

olivere:
Heya. I need to get back to working on Elastic and merge the PRs as well. Regarding this PR, I think we could try to get it into Elastic when 6.3 is released. I think that’s not too far ahead.

Zyqsempai:
Agree, i will try to prepare few more PR till that moment.

olivere:
See also: https://www.elastic.co/blog/opening-x-pack-phase-1-complete

olivere:
@Zyqsempai  Thanks so much. I merged this now. :-)

PR #737, opened 2018-03-16
Add PrecisionString to NewGeoHashGridAggregation()

### LINKED ISSUE #736: NewGeoHashGridAggregation allow also Precision as string like '1km' etc.

[ x ] elastic.v6 (for Elasticsearch 6.x)

`NewGeoHashGridAggregation().Precision(int|string)`
maybe newFunction...

`NewGeoHashGridAggregation().Precision(int)`

https://www.elastic.co/guide/en/elasticsearch/reference/6.2/search-aggregations-bucket-geohashgrid-aggregation.html

Dexus:
Fix #736


olivere:
Thanks for spending time to make Elastic better.

I just skimmed over the Java sources, and they only define [precision as an `int`](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoGridAggregationBuilder.java#L135-L138). May I ask where you found a reference of being able to use a string? Is there an integration test that using a string actually works?

If it does, I would favor to use an `interface{}` for `precision` so that we could pass both an `int` and a `string` with the same setter. What do you think?

Dexus:
Hi @olivere,

I found it at https://www.elastic.co/guide/en/elasticsearch/reference/6.2/search-aggregations-bucket-geohashgrid-aggregation.html 

> precision:     Optional. The string length of the geohashes used to define
                cells/buckets in the results. Defaults to 5.
                The precision can either be defined in terms of the integer
                precision levels mentioned above. Values outside of [1,12] will
                be rejected.
                **Alternatively, the precision level can be approximated from a
                distance measure like "1km", "10m". The precision level is
                calculate such that cells will not exceed the specified
                size (diagonal) of the required precision.** When this would lead
                to precision levels higher than the supported 12 levels,
                (e.g. for distances <5.6cm) the value is rejected.

Dexus:
Again hi @olivere,

Sure, an `interface{}` would be a good idea too, just thought to make sure that the type is a string, a new function would be the better choice. Otherwise you have to check there for an error...  and then there would no longer be a connection of functions, as is now the case.

olivere:
Regarding the usage of a string for `precision`. Can you check that ES actually accepts a string? In the Java sources, I cannot find a reference for parsing `precision` from a string. But I might just have skipped it.

If we change `Precision(precision int)` to `Precision(precision interface{})`, where do we have to check for errors? Of course, if you pass something that ES doesn't understand, it responds with an error. But that might also be the case with you passing some `precision` value (e.g. `15`) or a precision string of `100tons`. Elastic doesn't save you from doing the wrong thing ;-)

olivere:
We should of course use the documentation to state that you could pass a precision value of 1..12 or a string like `1km` (if that works) to `Precision`. If you don't follow the instructions, that's your problem really.

Dexus:
So I'm using this patch in my Go application right now, so it works without problems and returns the corresponding answers like with comparable int values.

Okay, then I'll go right back to it and use an `interface{}`. Then that should remain uniform.

Greetings, Josef

Dexus:
Hi @olivere,

so, now I have adapted the PR and query all int and int variants as well as the string. The string is then built correctly and supplied back by ES with correct values.

Tests are included but not for each Type only one of each (u)Int16 string etc.

Regards, Josef

search_aggs_bucket_geohash_grid.go
```diff
@@ -23,8 +27,18 @@ func (a *GeoHashGridAggregation) Field(field string) *GeoHashGridAggregation {
 	return a
 }
 
-func (a *GeoHashGridAggregation) Precision(precision int) *GeoHashGridAggregation {
-	a.precision = precision
+// Precision accepts the level as int value between 1 and 12 or Distance Units like "2km", "5mi" as described at
+// https://www.elastic.co/guide/en/elasticsearch/reference/6.2/common-options.html#distance-units and
+// https://www.elastic.co/guide/en/elasticsearch/reference/6.2/search-aggregations-bucket-geohashgrid-aggregation.html
+func (a *GeoHashGridAggregation) Precision(precision interface{}) *GeoHashGridAggregation {
+	switch reflect.TypeOf(precision).Kind() {
```
olivere: Is there any reason you don't just simply set the value?

```
a.precision = precision
```

search_aggs_bucket_geohash_grid_test.go
```diff
@@ -23,6 +23,96 @@ func TestGeoHashGridAggregation(t *testing.T) {
...
+		t.Fatalf("Marshalling to JSON failed: %v", err)
+	}
+	got := string(data)
+	expected := `{"geohash_grid":{"field":"location","precision":"2km"}}`
+	if got != expected {
+		t.Errorf("expected\n%s\n,got:\n%s", expected, got)
+	}
+}
+
+func TestGeoHashGridAggregation_PrecisionAsUint(t *testing.T) {
```
olivere: I think we could leave it as one example with a `string` and another with an `int`.

Dexus:
@olivere fixed

Dexus:
No problem, this is fixed now too.

olivere:
I just pushed 6.1.11. Thanks for your time.

Dexus:
You're welcome. Thanks for your hints and tips :+1:

Dexus:
@olivere maybe the prepare v6 relase need also be done ;)

olivere:
@Dexus I forgot something? Version [6.1.11 is tagged](https://github.com/olivere/elastic/tree/v6.1.11) and has [these changes](https://github.com/olivere/elastic/compare/v6.1.10...v6.1.11).

Dexus:
in the client as of the last you allways updated the Version

Dexus:
https://github.com/olivere/elastic/blob/release-branch.v6/client.go#L29

olivere:
Ahh, noes... _pushing 6.1.12. Thank you.

PR #745, opened 2018-03-27
Added roles field to node info

snowzach:
This will add the Roles field to node info


olivere:
Thanks. I will fix some of the other fields in that response. It should [align with the Java code](https://github.com/elastic/elasticsearch/blob/6.x/server/src/main/java/org/elasticsearch/action/admin/cluster/node/info/NodesInfoResponse.java#L58-L124).

PR #746, opened 2018-04-03
Add remove index alias action

mbesh:
As described in the Elasticsearch 5.0 and 6.0 documentation, there is an alias action that will remove an index as part of an Index Alias operation. This pull request implements that action.

An example is provided in the documentation just above the section labeled "Filtered Aliases": https://www.elastic.co/guide/en/elasticsearch/reference/6.0/indices-aliases.html#filtered

Relevant Java code: 
Definition of remove index action:
https://github.com/elastic/elasticsearch/blob/8e8fdc4f0efc6ddac866009baa22552668d94a12/server/src/main/java/org/elasticsearch/action/admin/indices/alias/IndicesAliasesRequest.java#L136

Validate function only checks index name on a remove index action:
https://github.com/elastic/elasticsearch/blob/8e8fdc4f0efc6ddac866009baa22552668d94a12/server/src/main/java/org/elasticsearch/action/admin/indices/alias/IndicesAliasesRequest.java#L251

### DIFF
--- a/indices_put_alias.go
+++ b/indices_put_alias.go
@@ -189,6 +189,39 @@ func (a *AliasRemoveAction) Source() (interface{}, error) {
 	return src, nil
 }
 
+// AliasRemoveIndexAction is an action to remove an index during an alias
+// operation.
+type AliasRemoveIndexAction struct {
+	index string // index name
+}
+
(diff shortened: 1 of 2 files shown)


olivere:
It's taken me some time, but your PR will be part of the next release... Thank you for supporting Elastic.

Issue #750, opened 2018-04-15
missing quote_field_suffix in simple_query_string


slawo:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

[X] elastic.v6 (for Elasticsearch 6.x)

### Please describe the expected behavior
The following code should be valid:

```
main() {
	ID := "1234"
	q := elastic.NewSimpleQueryStringQuery(`"` + ID + `"`)
	q.Field("items.id")
	q.Field("discounts.id")
	q.QuoteFieldSuffix(".exact")
	//...
}
```

### Please describe the actual behavior

`SimpleQueryStringQuery` doesn't implement `QuoteFieldSuffix()`

### Any steps to reproduce the behavior?

The code above.

Refer to https://www.elastic.co/guide/en/elasticsearch/reference/master/mixing-exact-search-with-stemming.html for another example of a valid query.

olivere:
Yep. That makes sense. Thanks for letting me know.

PR #751, opened 2018-04-16
add quote_field_suffix to simple_query_string

slawo:
fix for #750

### DIFF
--- a/search_queries_simple_query_string.go
+++ b/search_queries_simple_query_string.go
@@ -19,6 +19,7 @@ import (
 type SimpleQueryStringQuery struct {
 	queryText              string
 	analyzer               string
+	quoteFieldSuffix       string
 	operator               string
 	fields                 []string
 	fieldBoosts            map[string]*float64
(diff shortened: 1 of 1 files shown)


slawo:
It has been tested in an internal project

olivere:
Looking good. I will add it in the next release. I typically use these "bugs" as a chance to look into the Java code for other missing/removed properties. So it'll take a few days more.

slawo:
If you want I can complete all the other missing properties from es6.2

olivere:
Thanks for your support. It's basically synchronizing the Go code with [`QueryStringQueryBuilder.java`](https://github.com/elastic/elasticsearch/blob/v6.2.3/server/src/main/java/org/elasticsearch/index/query/QueryStringQueryBuilder.java) and [`SimpleQueryStringBuilder.java`](https://github.com/elastic/elasticsearch/blob/v6.2.3/server/src/main/java/org/elasticsearch/index/query/SimpleQueryStringBuilder.java).

But, to be honest, I think I'd like to do this myself. That way it's slower, but I keep myself in touch with the changes in core.

slawo:
OK, no wories

olivere:
Thanks again.

PR #752, opened 2018-04-16
remove droped properties in simple_query_string

slawo:
remove:
- lowercase_expanded_terms
- locale
as per: https://github.com/elastic/elasticsearch/pull/10086


olivere:
I marked them as Deprecated in 6.x. Will need to be removed with 7.x I guess.

PR #757, opened 2018-04-20
Cat api

justinbeal:
Problem: the [cat API's](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat.html) are not currently supported. This implements these API's for all models but snapshots*. Subsequent PR's should flesh out the other options for each individual model as well as format type, column filtering and other interesting things. 

* `_cat/snapshots/$REPOSITORY_NAME?v` currently throws an exception if the repository does not exist, and the repository is required. Setting up a local cluster with a repository to test the command requires a non-trivial amount of work, and while I don't care to test the output of `cat` I would like to test that the command itself runs like I do for the other models.

### DIFF
--- a/cat.go
+++ b/cat.go
@@ -1,0 +1,88 @@
+// Copyright 2012-present Oliver Eilhard. All rights reserved.
+// Use of this source code is governed by a MIT-license.
+// See http://olivere.mit-license.org/license.txt for details.
+
+package elastic
+
+import (
+	"context"
+	"net/url"
+)
+
+// CatIndicesService executes cat commands against the cluster.
+//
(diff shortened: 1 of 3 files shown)


olivere:
Thanks for your support. May I ask how you use the output? Wouldn’t it be better suited to implement the Cat APIs, but in a way that makes the results accessible without parsing text?

olivere:
I've closed this PR as I don't think returning strings would be helpful for anyone. You could do that with a simple HTTP call.

Instead, I've started to add some Cat APIs that return data deserialized from JSON, which are more usable in a programming environment.

PR #760, opened 2018-04-26
Added roles field to node info

cesarjimenez:
This will add the Roles field to node info.

This is exactly the same change made in #745, but in the Elasticsearch 5.X branch.


olivere:
Thanks for supporting Elastic. I hope to release a new Elastic version over the weekend, including this backport.

olivere:
Thank you.

PR #762, opened 2018-04-27
Use DoAsync in other queries WIP

gburanov:
WHY:

 I want to run async queries and get the task id for other queries as well (not only reindexer)


olivere:
Thanks for contributing to Elastic. I will review and see what we can do...

olivere:
I added `DoAsync` to both `UpdateByQuery` and `DeleteByQuery`. However, I decided to implement it with my own version. While extracting common code is generally a good idea, it has backfired with Elastic quite often as ES changed things in a way that breaks the common basis. So I tend to copy & paste in order to be prepared for future changes.

gburanov:
thank you!

PR #768, opened 2018-05-04
nested: Add support for nested queries to search hits

wuurrd:
As stated in https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-inner-hits.html

### DIFF
--- a/search.go
+++ b/search.go
@@ -467,6 +467,12 @@ type SearchHits struct {
 	Hits      []*SearchHit `json:"hits"`      // the actual hits returned
 }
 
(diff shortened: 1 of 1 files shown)


olivere:
@wuurrd Yep, that's looking good. Will add it to the next release. Can you please rename `NestedHit` to `NestedIdentity` to be [in line with the Java source](https://github.com/elastic/elasticsearch/blob/916bf9d26d6292dcb9b229123c58a86f0e741ffd/server/src/main/java/org/elasticsearch/search/SearchHit.java#L847-L961).

PR #776, opened 2018-05-13
Add GetRequest function to retrieve the requests from the queue

rchicoli:
I would like to add some external checks for the requests. As long the `requests` var is not exported, I've wrote a small func for returning them.

Well, if this gets merged, I will make sure to update the other versions as well.

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -163,6 +163,11 @@ func (s *BulkService) estimateSizeInBytes(r BulkableRequest) int64 {
 	return int64(size)
 }
 
+// GetRequests returns the bulkable requests from the queue
+func (s *BulkService) GetRequests() []BulkableRequest {
+	return s.requests
+}
+
 // NumberOfActions returns the number of bulkable requests that need to
 // be sent to Elasticsearch on the next batch.
 func (s *BulkService) NumberOfActions() int {
--- a/bulk_test.go
+++ b/bulk_test.go
@@ -11,6 +11,7 @@ import (
 	"math/rand"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


olivere:
I can see all kinds of things going wrong if we make `requests` visible to the outside. E.g. you could add a request from the outside without the Bulk Service knowing anything about it. That makes future changes like e.g. adding a cache or something impossible.

I cannot foresee a use case for this. You are responsible for adding requests to Bulk API, so you could control and monitor what's being added in the first place. Can you post a use case where this is necessary at all?

rchicoli:
In other wrods, I would like to write a custom bulk processor similar to what you did. So I am trying to achieve this line https://github.com/olivere/elastic/blob/release-branch.v6/bulk_processor.go#L516. The `GetRequests` would solve my problem.

olivere:
Wouldn't you be able receive the requests along with a potential error [here](https://github.com/olivere/elastic/blob/release-branch.v6/bulk_processor.go#L531-L533)?

I mean: I'm not arguing against anything. I'm just trying to steer us into a direction that will keep the library stable and robust in the long run. Quick fixes have the potential to make things even worse.

rchicoli:
I understand your concern, no worries. If I was using the bult-in `BulkProcessor`, it would be enough.
But I would prefer to go with my client wrapper. Not forgetting to mention, this is only a `readonly` function.

The only problem I could see here, if someone misuse this function and run it in a different thread then the `Do` func, this would cause this function to retrieve false values. Apart from that, it should be just fine.

Alternative: I could add the function `afterFn` and `beforeFn` to the bulkService (so the bulk processor could inherit them), but this looks for me like an overkill. Don't you think so?

rchicoli:
Yeh, maybe it is a good idea, though the `BulkProcessor` is well written imho.
The question is now, what to do with this PR?
Well, I think I could build some workaround in my project in order to reach the same goal, even though I was trying to avoid that.

olivere:
When starting `BulkProcessor` I really wanted a rather simple solution that matches the [`BulkProcessor` from Java](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/action/bulk/BulkProcessor.java). It seems to me that that doesn't match your use case, which is okay. I was quite aware of its limitations, and I wrote about [it in the wiki](https://github.com/olivere/elastic/wiki/BulkProcessor). E.g. with regards to error handling I wrote: "There is no promise from bulk processor other than "we try our best to put your data into Elasticsearch".

So I think the best way to move forward is to copy bulk processor and modify it to match your use case.

PR #777, opened 2018-05-13
Change bulk reset func to exportable

rchicoli:
the reason of this change is: all requests are persistent.

if `s.client.PerformRequestWithOptions` (https://github.com/olivere/elastic/blob/release-branch.v6/bulk.go#L249) returns an error, the request won't be reseted, but in some special cases, I would like to decide that.

So by exporting the Reset function, it gives the user the flexibity to reset the queue and possibily write those requests to the file system, if wished.

If this change gets merged, I will make sure to update all other versions.

Great project btw.

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -52,7 +52,8 @@ func NewBulkService(client *Client) *BulkService {
 	return builder
 }
 
-func (s *BulkService) reset() {
+// Reset cleans up the request queue
+func (s *BulkService) Reset() {
 	s.requests = make([]BulkableRequest, 0)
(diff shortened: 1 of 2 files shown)


olivere:
Opening up the `Reset` function is okay. I will enable this in the next release.

You could then decide whether to reset the Bulk API or not via [Retrier](https://github.com/olivere/elastic/wiki/Retrier-and-Backoff).

rchicoli:
Great, so I will try to push this change for the others client versions. Cheers.

PR #800, opened 2018-06-14
nested hit offset support

pattanapongJWS:
add struct Nested for offeset and field

### DIFF
--- a/search.go
+++ b/search.go
@@ -411,12 +411,6 @@ func (s *SearchService) Do(ctx context.Context) (*SearchResult, error) {
 	return ret, nil
 }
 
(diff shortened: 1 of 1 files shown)


olivere:
I'm sorry, I almost forgot this.

I see this is not quite complete as a `NestedIdentity` (as it's called in the Java source) can itself contain a child [`NestedIdentity` recursively](https://github.com/elastic/elasticsearch/blob/v6.4.0/server/src/main/java/org/elasticsearch/search/SearchHit.java#L911-L924).

I will address this in 6.2.4, promised.

olivere:
... and 5.0.75 as well, of course.

Issue #766, opened 2018-04-30
Reset Bulk Service when body cannot be serialized


rwynn:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[X ] elastic.v5 (for Elasticsearch 5.x)
[X ] elastic.v6 (for Elasticsearch 6.x)

### Please describe the expected behavior
The bulk service should be reset when it encounters an errors serializing its body. E.g. when the input contains a value of NaN, +/- Infinity, or a time.Time with a year outside of [0, 9999].

https://github.com/olivere/elastic/blob/release-branch.v6/bulk.go#L202

Otherwise, the service (when used in conjunction with a bulk processor) gets stuck in an endless error loop since it is only reset after a request is completed.  The request will never happen if the input is invalid.

I'm not sure if it should be the responsibility of the bulk service or the bulk processor to break out of this loop.

### Please describe the actual behavior

When the bulk service is used with a bulk processor, once a request is encountered that will not serialize, the bulk processor will not be able to reset the bulk service and continually get errors back.  

### Any steps to reproduce the behavior?

A document with a float64 that passes `math.IsNaN()` or `math.IsInf(var, 0)` or a time.Time with a year outside of `[0,9999]` will trigger this error loop since those values trigger a JSON serialization error.

larrycinnabar:
I've had problems because of this particular reason.

In my case, I needed very quick solution, so I've forked library and made an exported version `.Reset()` of `.reset()` method. So I could reset bulk service in client code manually when needed

olivere:
Making `Reset` public is a viable workaround. It puts the caller into control. I have to test if resetting on error automatically will come with side-effects.

olivere:
@rwynn Does the problem with an infinite loop even exist if you [set a specific backoff strategy with Bulk Processor](https://github.com/olivere/elastic/blob/release-branch.v6/bulk_processor.go#L125-L129)?

rwynn:
@olivere I would need to verify on the backoff stategy but I think the infinite loop would still exist.  The bulk processor reuses the configured number of worker/service pairs created when the processor is started.  The only path to reset the requests slice for those services is forward: to be able to form request, transmit, and parse response.

[...]

PR #802, opened 2018-06-16
Prevent deadlock in the processor which results when a bulk line item…

rwynn:
… cannot be serialized

See issue #766 for more details.  This commit ensures that the Source()
of a bulk index request can be constructed successfully before it is
added to the bulk service.

When the bulk processor adds a request that cannot be serialized it can
enter an infinite loop where it repeatedly tries to flush the request but cannot.
This can happen when any source document contains JSON marshalling errors (e.g. +/-
Infinity or invalid time.Time values).  This problem becomes especially evident
when the bulk process is configured to flush asynchronously.

### DIFF
--- a/bulk_processor.go
+++ b/bulk_processor.go
@@ -451,9 +451,11 @@ func (w *bulkWorker) work(ctx context.Context) {
 		case req, open := <-w.p.requestsC:
 			if open {
 				// Received a new request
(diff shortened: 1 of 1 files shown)


olivere:
I'm on vacation and be back in July. I will review asap when I'm back in town.

PR #803, opened 2018-06-16
Requeue bulk items with response code 429 - Too Many Requests

rwynn:
When Elasticsearch gets overloaded it may return response code 429
for line items in the bulk response that it could not process.  This
commit detects that response code and requeues the line item so that it
is not dropped.

For reference see the code in the elastic beats elasticsearch
output plugin.

https://github.com/elastic/beats/blob/master/libbeat/outputs/elasticsearch/client.go#L500.


olivere:
I'm on vacation and be back in July. I will review asap when I'm back in town.

PR #818, opened 2018-07-08
add keyed option to date histogram aggregation

sgnrslv:
The keyed option is present since version 5.4 (see https://www.elastic.co/guide/en/elasticsearch/reference/5.4/search-aggregations-bucket-datehistogram-aggregation.html).


olivere:
Will be part of the next release. Sorry for the delay.

PR #819, opened 2018-07-08
add keyed option to date histogram aggregation

sgnrslv:
this pull is similar to https://github.com/olivere/elastic/pull/818 for v6


olivere:
Thanks. Will be part of 6.2.4.

PR #827, opened 2018-07-18
Allow RetrierFunc to be an adapter

veqryn:
This PR allows the use of RetrierFunc as an adapter:
```go
client, err := NewClient(
	SetRetrier(RetrierFunc(func(context.Context, int, *http.Request, *http.Response, error) (time.Duration, bool, error){
		return 0, false, nil
	})),
)
```

Here is an example from the stdlib: https://github.com/golang/go/blob/00ebfcde7f3b6e63447e2a3962975e0a8c3729dd/src/net/http/server.go#L1964-L1973

---
This change is [Reviewable](https://reviewable.io/reviews/olivere/elastic/827)

### DIFF
--- a/retrier.go
+++ b/retrier.go
@@ -10,8 +10,15 @@ import (
 	"time"
 )
 
-// RetrierFunc specifies the signature of a Retry function.
+// RetrierFunc specifies the signature of a Retry function, and is an adapter
+// to allow the use of ordinary Retry functions. If f is a function with the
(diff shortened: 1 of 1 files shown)


olivere:
Yeah, that looks good. Will be part of the next version.

olivere:
Thank you.

veqryn:
thx!

PR #836, opened 2018-07-25
check for nested context errors

veqryn:
I am using this library against AWS ElasticSearch Service, and found it was constantly logging out that the connection/nodes were dead, whenever the context was cancelled.
I looked and saw you were already handling that, so it took me a while to figure out why it wasn't being caught.

Turns out that the url.Error is wrapping a url.Error that is wrapping the context error, because of transport wrapping a second call to the http client:
https://github.com/olivere/elastic/blob/release-branch.v6/recipes/aws-connect/main.go#L32

I thought about avoiding the second call to the http client, however it is actually a good thing to make that second call after modifying the request, since the http client includes safeguards around not forwarding the authorization headers to untrusted domains.

So a simple solution is just to check for a url.Error wrapping the original url.Error.

---
This change is [Reviewable](https://reviewable.io/reviews/olivere/elastic/836)

### DIFF
--- a/client.go
+++ b/client.go
@@ -1275,16 +1275,9 @@ func (c *Client) PerformRequest(ctx context.Context, opt PerformRequestOptions) 
 
 		// Get response
 		res, err := c.c.Do((*http.Request)(req).WithContext(ctx))
-		if err == context.Canceled || err == context.DeadlineExceeded {
+		if isContextErr(err) {
 			// Proceed, but don't mark the node as dead
 			return nil, err
-		}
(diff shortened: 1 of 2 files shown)


olivere:
Thanks for supporting Elastic. That's looking good. 👍 

Would it make sense to export `IsContextErr(...)`?

veqryn:
I definitely could. I guess I should move it to the errors.go file then too.

olivere:
Merged, thanks. Will be in 6.1.26.

Issue #863, opened 2018-08-15
TaskId() requires a nodeId in the request


GarrettKelley:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[x ] elastic.v6 (for Elasticsearch 6.x)

### Please describe the expected behavior
As of 6.3, the POST request looks like this: `POST _tasks/node_id:task_id/_cancel`. The request requires the nodeId followed by taskId. Can be seen in this documentation: https://www.elastic.co/guide/en/elasticsearch/reference/current/tasks.html#task-cancellation

### Please describe the actual behavior
In your code the request looks like `POST _tasks/task_id/_cancel`, so it is not cancelling tasks properly.

### Any steps to reproduce the behavior?

Here is the change I made in our local copy to get cancellation to work properly.
[tasks_cancel.zip](https://github.com/olivere/elastic/files/2290325/tasks_cancel.zip)

olivere:
As outlined in #864: The Elastic Java source was a bit convoluted around `taskId` when I implemented this. Sometimes they used a `long`, sometimes they used a `string`. Maybe they have cleaned up a bit. I have to check, but it might take a while. Sorry.

olivere:
The REST API specifies a single parameter called [`task_id`](https://github.com/elastic/elasticsearch/blob/6.x/rest-api-spec/src/main/resources/rest-api-spec/api/tasks.get.json#L6), defined as a `node_id:task_number` for the `parent_task_id` parameter in the [Task Cancel API](https://github.com/elastic/elasticsearch/blob/6.x/rest-api-spec/src/main/resources/rest-api-spec/api/tasks.cancel.json#L25). The `task_id` is also sometimes [returned individually](https://github.com/elastic/elasticsearch/blob/6.x/server/src/main/java/org/elasticsearch/tasks/TaskInfo.java#L190-L191) and sometimes [returned as a `string`](https://github.com/elastic/elasticsearch/blob/6.x/server/src/main/java/org/elasticsearch/tasks/TaskInfo.java#L207).

[...]

PR #864, opened 2018-08-15
Add node id to task cancel request

GarrettKelley:
Task cancel request requires a node id as of 6.3.

References issue https://github.com/olivere/elastic/issues/863


olivere:
The Elastic Java source was a bit convoluted around `taskId` when I implemented this. Sometimes they used `long`, sometimes they used `String`. Maybe they have cleaned up a bit. I have to check, but it might take a while. Sorry.

Issue #862, opened 2018-08-15
Ability to append headers to requests


GarrettKelley:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[ x ] elastic.v6 (for Elasticsearch 6.x)

### Please describe the expected behavior
As of ES6, you can add a the X-Opaque-ID header to requests in order to be able to trace/track tasks as they execute. This functionality is found in the ES documentation here:
https://www.elastic.co/guide/en/elasticsearch/reference/6.2/tasks.html#_identifying_running_tasks
The current version of this API doesn't allow headers to be appended.  I modified the source code to resolve this, but it appears I can't make a branch to make a PR without first creating an issue.

### Please describe the actual behavior
That the library provides a way to attach headers to the requests that are submitted to ES.

### Any steps to reproduce the behavior?
If you show the list of running tasks, the headers field is always empty.  The impacted files are attached.  I only modified the ScrollService and TaskList (as opposed to all services) because that was the service I needed to use for my personal case.

Here are the modified files:
[elasticHeaders.zip](https://github.com/olivere/elastic/files/2290216/elasticHeaders.zip)

olivere:
Add in 51d010201ba7bcb2dec3d84f225fec8617de11d5 and 51d010201ba7bcb2dec3d84f225fec8617de11d5.

PR #865, opened 2018-08-15
add headers to request

GarrettKelley:
Added ability to append headers to rest calls.
I only added the field on scroll and tasks_list because those were my use case, but this could be useful everywhere.

References issue https://github.com/olivere/elastic/issues/862

### DIFF
--- a/client.go
+++ b/client.go
@@ -1182,6 +1182,7 @@ type PerformRequestOptions struct {
 	ContentType  string
 	IgnoreErrors []int
 	Retrier      Retrier
+	Headers      map[string]string
 }
 
 // PerformRequest does a HTTP request to Elasticsearch.
@@ -1259,6 +1260,12 @@ func (c *Client) PerformRequest(ctx context.Context, opt PerformRequestOptions) 
 		}
 		if opt.ContentType != "" {
 			req.Header.Set("Content-Type", opt.ContentType)
+		}
+
+		if len(opt.Headers) > 0 {
+			for headerName, headerValue := range opt.Headers {
+				req.Header.Set(headerName, headerValue)
+			}
 		}
 
 		// Set body
--- a/client.go
+++ b/client.go
@@ -1182,7 +1182,7 @@ type PerformRequestOptions struct {
 	ContentType  string
 	IgnoreErrors []int
 	Retrier      Retrier
-	Headers      map[string]string
+	Headers      http.Header
 }
 
 // PerformRequest does a HTTP request to Elasticsearch.
@@ -1263,8 +1263,10 @@ func (c *Client) PerformRequest(ctx context.Context, opt PerformRequestOptions) 
 		}
 
 		if len(opt.Headers) > 0 {
-			for headerName, headerValue := range opt.Headers {
-				req.Header.Set(headerName, headerValue)
+			for key, value := range opt.Headers {
+				for _, v := range value {
+					req.Header.Add(key, v)
+				}
 			}
 		}
 
--- a/scroll.go
+++ b/scroll.go
@@ -36,6 +36,7 @@ type ScrollService struct {
 	ignoreUnavailable *bool
 	allowNoIndices    *bool
 	expandWildcards   string
+	headers           map[string]string
 
 	mu       sync.RWMutex
 	scrollId string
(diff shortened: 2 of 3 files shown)


tasks_list.go
```diff
@@ -232,6 +243,7 @@ type TaskInfo struct {
 	RunningTimeInNanos int64       `json:"running_time_in_nanos"`
 	Cancellable        bool        `json:"cancellable"`
 	ParentTaskId       string      `json:"parent_task_id"` // like "YxJnVYjwSBm_AUbzddTajQ:12356"
+	Headers            interface{} `json:"headers"`
```
olivere: We should try to make it a specific type instead of `interface{}`.

scroll.go
```diff
@@ -51,6 +52,15 @@ func NewScrollService(client *Client) *ScrollService {
 	return builder
 }
 
+// AddHeader sets headers on the request
+func (s *ScrollService) AddHeader(headerName string, headerValue string) *ScrollService {
```
olivere: I have some comments here:

1. Can we rename this from `AddHeader` to `Header` please? All services are builders, and it looks really bad to have a builder pattern with:

```
service.
  AddHeader("...", "...").
  SetTimeout(...).
  SetXXX(...).
  SetYYY(...).
  Do(...)
```

instead of 

```
service.
  Header("...", "...").
  Timeout(...).
  XXX(...).
  YYY(...).
  Do(...)
```

2. Can we use variable names like `key` and `value` instead of `headerName` and `headerValue` (like [`http.Header#Add`](https://golang.org/pkg/net/http/#Header) does). It's clear that e.g. the `value` is the header value for a method called `Header`.

scroll.go
```diff
@@ -36,6 +36,7 @@ type ScrollService struct {
 	ignoreUnavailable *bool
 	allowNoIndices    *bool
 	expandWildcards   string
+	headers           map[string]string
```
olivere: Can we make this a [`http.Header`](https://golang.org/pkg/net/http/#Header) instead.

tasks_list.go
```diff
@@ -97,6 +98,15 @@ func (s *TasksListService) GroupBy(groupBy string) *TasksListService {
 	return s
 }
 
+// AddHeader sets headers on the request
+func (s *TasksListService) AddHeader(headerName string, headerValue string) *TasksListService {
```
olivere: See previous comment in `ScrollService`.

client.go
```diff
@@ -1261,6 +1262,12 @@ func (c *Client) PerformRequest(ctx context.Context, opt PerformRequestOptions)
 			req.Header.Set("Content-Type", opt.ContentType)
 		}
 
+		if len(opt.Headers) > 0 {
+			for headerName, headerValue := range opt.Headers {
```
olivere: Can we use `key` / `value` here as well. See comment below.

olivere: Or even: `k, v`.

tasks_list.go
```diff
@@ -220,18 +232,19 @@ type DiscoveryNode struct {
...
+	Action             string                 `json:"action"`
+	Status             interface{}            `json:"status"`      // has separate implementations of Task.Status in Java for reindexing, replication, and "RawTaskStatus"
+	Description        interface{}            `json:"description"` // same as Status
+	StartTime          string                 `json:"start_time"`
+	StartTimeInMillis  int64                  `json:"start_time_in_millis"`
+	RunningTime        string                 `json:"running_time"`
+	RunningTimeInNanos int64                  `json:"running_time_in_nanos"`
+	Cancellable        bool                   `json:"cancellable"`
+	ParentTaskId       string                 `json:"parent_task_id"` // like "YxJnVYjwSBm_AUbzddTajQ:12356"
+	Headers            map[string]interface{} `json:"headers"`
```
olivere: This could be a `http.Headers` as well, if [`TaskInfo.java`](https://github.com/elastic/elasticsearch/blob/6.x/server/src/main/java/org/elasticsearch/tasks/TaskInfo.java#L209-L213) is up-to-date.

If you don't have the time to correct this, don't worry. I will merge this probably over the weekend.

olivere:
Awesome. Nearly done. Don't worry about the last comment: If you don't have the time, I will change this and tests for all the things (probably) over the weekend.

PR #867, opened 2018-08-17
Remove operation_threading parameter from Forcemerge

xose:
The `operation_threading` parameter was removed on Elasticsearch 1.2. Setting it causes an error to be returned.

https://github.com/elastic/elasticsearch/pull/6044


olivere:
Thanks for supporting Elastic. It's hard to keep on top of all the changes, even going back to 1.x.

Issue #868, opened 2018-08-19
Support `index_patterns` in template response?


phillbaker:
Thanks for all the hard work on this library. I think this is a result of some of the breaking changes in v5 -> v6 Elasticsearch, but wanted to open an issue before starting a PR.

### Which version of Elastic are you using?

[ ] elastic.v2 (for Elasticsearch 1.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[x] elastic.v6 (for Elasticsearch 6.x)

### Please describe the expected behavior

Retrieving an index template returns the `index_patterns` it was created with.

### Please describe the actual behavior

elastic.v6 defines a `Template` member of the `IndicesGetTemplateResponse` struct but not an `IndexPatterns` member. In Elasticsearch v6, `index_patterns` is always returned instead of `template`.

https://github.com/olivere/elastic/blob/eb69fbabfd6151d6c0f6d3d850d4f61edcff3763/indices_get_template.go#L124

### Any steps to reproduce the behavior?

```
res, err := client.IndexGetTemplate(id).Do(context.TODO())
```

Assuming that it makes sense to replace `Template` with `IndexPatterns`, I'd be happy to contribute a PR.

olivere:
Sorry for the delay.

Yes, you're right. There even is a comment in the Java code regarding that `"template"`. I think we can keep that in the response for v5 but safely remove it in v6. In v6, as you said, it needs to be replaced by `index_patterns`. Now that we're at this, we should check the whole response structure, which is basically what [this code](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/action/admin/indices/template/get/GetIndexTemplatesResponse.java#L72-L80) and [this code](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/metadata/IndexTemplateMetaData.java#L393-L439) generates.

I'd love a PR for that if you have the time.

olivere:
Here's [the comment](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/metadata/IndexTemplateMetaData.java#L504-L507) I talked about in the last comment.

PR #874, opened 2018-08-26
Update GetTemplateResponse for v6.

phillbaker:
Closes https://github.com/olivere/elastic/issues/868.

Updates the `IndicesGetTemplateResponse` so that the `IndicesGetTemplateResponse` struct defines an `IndexPatterns` member not a `Template` member. In Elasticsearch v6, `index_patterns` is always returned instead of `template`. Adds a test to cover this struct.

As suggested in the issue, checked the response structure in the java source, and the rest of the response seems unchanged.


olivere:
Thanks. I will merge this into the next release.

Issue #897, opened 2018-09-07
Unexpected SearchShardsService.Do() error when shards are relocating


lr-paul:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

[ ] elastic.v6 (for Elasticsearch 6.x)
[ X ] elastic.v5 (for Elasticsearch 5.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v2 (for Elasticsearch 1.x)

### Please describe the expected behavior
`SearchShardsService.Do()` does not return an error when the cluster is otherwise alive and healthy but shards are relocating.

### Please describe the actual behavior
This error is returned by `SearchShardsService.Do()`: `SearchShards error: json: cannot unmarshal string into Go struct field ShardsInfo.relocating_node of type bool`

### Any steps to reproduce the behavior?
Call `SearchShardsService.Do()` while shards are relocating.

olivere:
Thanks for your support. Will be part of 6.2.6 and 5.0.76.

PR #898, opened 2018-09-07
[Issue #897] [v5] Fix type for parsed ShardsInfo json object

lr-paul:
see https://github.com/olivere/elastic/issues/897


olivere:
Oh, I see. Thanks for fixing that. Will be part of the next release.

olivere:
Must probably also be fixed in v6 as well.

olivere:
Yep., it's [here](https://github.com/olivere/elastic/blob/release-branch.v6/search_shards.go#L183) as well.

olivere:
Thanks for your support. Will be part of 6.2.6. Added a few substructures of the SearchShards API response in v6 as well.

PR #904, opened 2018-09-12
Add missing "missing" options to some aggregations

carldunham:
A simple copy/paste job to add an option for averaging aggregators to use for missing data elements.


olivere:
Thanks for your support. I added the `missing` case to all single-metric aggs as necessary. Furthermore, 9fafea4 adds the weighted average aggregation. It will be part of 6.2.6.

PR #929, opened 2018-10-08
Add SearchService.MaxResponseSize() to guard against OOM situations

erikgrinaker:
`elastic` currently decodes the entire ES response uncritically, which can cause out-of-memory errors with very large responses. This adds a `MaxResponseSize()` method to `SearchService` which causes it to error out with `ErrResponseSize` if the ES response exceeds the provided limit.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1230,14 +1230,15 @@ func (c *Client) mustActiveConn() error {
 
 // PerformRequestOptions must be passed into PerformRequest.
 type PerformRequestOptions struct {
-	Method       string
-	Path         string
-	Params       url.Values
-	Body         interface{}
-	ContentType  string
-	IgnoreErrors []int
-	Retrier      Retrier
-	Headers      http.Header
+	Method          string
+	Path            string
+	Params          url.Values
+	Body            interface{}
+	ContentType     string
+	IgnoreErrors    []int
+	Retrier         Retrier
+	Headers         http.Header
+	MaxResponseSize int64
 }
 
 // PerformRequest does a HTTP request to Elasticsearch.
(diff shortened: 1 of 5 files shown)


erikgrinaker:
Not sure what's up with the test failure here. It doesn't seem to be related to my changes, since it fails locally with the parent revision as well (156bc8f4ccd31cbf7567a0ed3a64335d6767e4b9), even though it apparently passed on Travis.

olivere:
Don't worry, [this test](https://travis-ci.org/olivere/elastic/jobs/438601295#L1837) seems flaky in recent versions of Elasticsearch. It allows you to pass an arbitrary HTTP header value that should be returned in the response, but sometimes isn't.

The PR looks okay. May I ask with what kind of requests you get these errors. We have fairly large responses in production as well but AFAIK we've yet to see an OOM panic. Is your environment resource constrained?

erikgrinaker:
We're a SaaS startup providing a hosted data backend with a custom query language on top. We support JSON documents of up to 32 MB each, and our query engine currently uses pages/scrolls of 200 documents when fetching hits from ES (a decent compromise between memory usage and performance as most documents are a few KB). This means that in the worse case a query can return a 6.4 GB response, which also has to be decoded, requiring even more memory. A user can easily submit multiple of these queries, and if one of our API instances crashes due to OOM then the load balancer will resubmit the GET request to other instances, leading to cascading failures.

We'll clearly have to use a range of approaches to prevent OOM situations like these, but having a fail-safe when reading very large ES responses is quite important to us.

olivere:
Interesting. Thanks for the feedback.

Just merged the PR. Will be available in 6.2.11, probably released over the weekend.

erikgrinaker:
Great, thanks!

PR #941, opened 2018-10-15
Add ScrollService.MaxResponseSize()

erikgrinaker:
I'm afraid I forgot to include this in PR #929, sorry for the noise.


olivere:
Yep. Will include that in the next release.

PR #943, opened 2018-10-18
Add Snapshot deletion

danlsgiga:
The Elasticsearch snapshot deletion API is a single endpoint that accepts a `DELETE` verb to delete a previously created snapshot.

This PR add 2 new files following the same structure used by `snapshot_create.go` but simplified for the delete feature.

1. snapshot_delete.go
2. snapshot_delete_test.go

I have it running successfully in an internal project, so pushing the contribution back upstream.

Ref: https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,7 +358,8 @@ The cat APIs are not implemented as of now. We think they are better suited for 
 
 - Snapshot and Restore
   - [x] Repositories
-  - [x] Snapshot
+  - [x] Snapshot create
+  - [x] Snapshot delete
   - [ ] Restore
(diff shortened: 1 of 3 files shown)


olivere:
Thanks. I'll look into it.

olivere:
I merged this into v6 in a3ac0902b2c71a860fd4e668eb1a5cab52579b10. I had to change a few bits to make the services look-a-like, but the feature should be there in the next release.

Thanks for your support.

PR #953, opened 2018-10-31
Add _nested to SearchHit

yangjqun:


olivere:
Thank you.

The nested identity can be recursive according to the [Java code](https://github.com/elastic/elasticsearch/blob/6.x/server/src/main/java/org/elasticsearch/search/SearchHit.java#L928-L951). Can you please add the missing code?

Issue #958, opened 2018-11-02
NewMatchQuery and NewCommonTermsQuery acts different when using same CutoffFrequency


david30907d:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

- [x] elastic.v6 (for Elasticsearch 6.x)
### Please describe the expected behavior

i've read these two doc and persumed that these function would have the same results when setting cutofffrequency
* <https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-common-terms-query.html>
* <>https://www.elastic.co/guide/en/elasticsearch/reference/6.2/query-dsl-match-query.html

### Please describe the actual behavior

maybe they should behave in the same way?

### Any steps to reproduce the behavior?

you can check whether these two query return the same results

* `elastic.NewMatchQuery(field, query).Analyzer("some_analyzer").CutoffFrequency(0.0001)`
* `elastic.NewCommonTermsQuery(field, query).Analyzer("some_analyzer").CutoffFrequency(0.0001)`
<https://github.com/olivere/elastic/blob/release-branch.v6/search_queries_common_terms.go#L102>
<https://github.com/olivere/elastic/blob/release-branch.v6/search_queries_match.go#L179>

olivere:
The PR is merged, thank you!

From the client perspective, I think it shouldn't care if two different queries now return the same results. If there are two APIs/DSLs that do the same, the client should have them, unless they've been deprecated.

david30907d:
@olivere  thanks, it's my pleasure to have a commit in this repo.

Sorry but i still have an unrelated and newbie question:

1. `minimum_should_match` in `match query` means high frequency minimum_should_match or low frequency?
2. `operator` in `match query` means high frequency operator or low frequency

That is, this Match Qeury

```
GET /_search
{
    "query": {
        "match": {
            "body": {
                "query": "nelly the elephant as a cartoon",
                "cutoff_frequency": 0.001,
                "minimum_should_match": 2,
                "operator": "or"
            }
        }
    }
}
```

Is it  roughly equivalent to the DSL below or not ?

```
GET /_search
{
    "query": {
        "bool": {
            "must": {
                "bool": {
                    "should": [
                    { "term": { "body": "nelly"}},
                    { "term": { "body": "elephant"}},
                    { "term": { "body": "cartoon"}}
                    ],
                    "minimum_should_match": 2
                }
            },
            "should": [
[...]

PR #959, opened 2018-11-02
<Fix>: dereferencing the pointer cutoffFrequency, not get the address of cutoffFrequency

david30907d:
i've opened a [issue](https://github.com/olivere/elastic/issues/958)

the weird thing is that even i've dereferenced the pointer cutoffFrequency, these commonTermsQuery and MatchQuery still don't return the same results...

Also check this [blog](https://www.elastic.co/blog/stop-stopping-stop-words-a-look-at-common-terms-query) if you're available and I quote `Note: Common Terms has also been incorporated into the Match query and can be enabled by setting `cutoff_frequency` to a value like 0.001` 

Making me feel that Common Terms Query and Match query can return same results in some way

### DIFF
--- a/search_queries_match.go
+++ b/search_queries_match.go
@@ -176,7 +176,7 @@ func (q *MatchQuery) Source() (interface{}, error) {
 		query["zero_terms_query"] = q.zeroTermsQuery
 	}
 	if q.cutoffFrequency != nil {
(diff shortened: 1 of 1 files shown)


olivere:
That looks like a bug. Thank you.

PR #972, opened 2018-11-22
Added the Global object in the Xpack Get Role response

Enteris:
Fix: the Global object was missing from the Get Role response. Added a test to add it to the body as it doesn't appear by default if not put during role creation.


olivere:
Can we maybe further specify the `XPackSecurityRole` entity instead of using an `interface{}` here? It looks like [you can see the structure in the Java code](https://github.com/elastic/elasticsearch/blob/6.x/client/rest-high-level/src/main/java/org/elasticsearch/client/security/user/privileges/Role.java#L174-L176).

olivere:
The `metadata` field can also be exposed as a `map[string]interface{}`, right?

Enteris:
I updated the metadata field as it is indeed a map[string]interface{}
For the other part, I think you mean the Global object. I checked a bit in the code (as it is not really documented) and it really seems to be an Object. The object itself is somewhat complex and I'm really unsure of its stability going forward so I would leave it as an empty interface if that's alright with you. 
We could always update the object mapping when it is better defined by Elastic (and better supported too, it only seems to support manage privilege class for now)

olivere:
I did some final changes to the code according to the Java sources. I will leave it like that and will include it in 6.2.14. If you find something not quite right, just file a new issue. Cheers.

PR #973, opened 2018-11-23
search: added the feature to use a stored template at search time.

ehsangolshani:
sometimes there is a need to a stored template when we use search method, especially when our query is very large, it's a good practice to store the query in Elasticsearch first, then use the search method in client with template_id and some parameters to pass it.
link to elasticsearch feature: [Search Template](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-template.html)

### DIFF
--- a/search.go
+++ b/search.go
@@ -320,6 +320,18 @@ func (s *SearchService) MaxResponseSize(maxResponseSize int64) *SearchService {
 	return s
 }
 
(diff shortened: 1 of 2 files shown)


olivere:
I looked this up in the Java code and official clients. The Search Template API is a separate API from the Search API. I created issue #977 to track this.

PR #974, opened 2018-11-23
Example of v5 with Elascticsearch, Kibana

pcherednichenko:
Example of v5 with Elascticsearch, Kibana in docker-compose:
https://github.com/pcherednichenko/goek/

Using:
- Golang 1.11
- Elasticsearch v6.5.1
- Kibana v6.5.1
- docker-compose

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,8 @@ creating a client, creating an index, adding a document, executing a search etc.
 
 An example is available [here](https://olivere.github.io/elastic/).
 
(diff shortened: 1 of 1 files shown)


olivere:
You're using elastic.v5 with Elasticsearch/Kibana 6.x. That's not a good example actually.

There's a [Wiki page for examples, actually](https://github.com/olivere/elastic/wiki/Examples). It's probably better to put it there than in the README.

PR #983, opened 2018-11-30
Add support for listing snapshots for a repository

wesleyk:


olivere:
Thanks for supporting open source and elastic.

The available services are defined by the REST API specifications listed [here](https://github.com/elastic/elasticsearch/tree/6.x/rest-api-spec/src/main/resources/rest-api-spec/api). 

What you are trying to implement here is specified in [`snapshots.get.json`](https://github.com/elastic/elasticsearch/blob/6.x/rest-api-spec/src/main/resources/rest-api-spec/api/snapshot.get.json), which we don't have yet (yay!). We'll have to use the REST API spec and all its fields on the request side.

Unfortunately the response data is not specified, so we need to dig into [the Java source](https://github.com/elastic/elasticsearch/tree/6.x/server/src/main/java/org/elasticsearch/action/admin/cluster/snapshots/get) for that. Here's what [the server returns as a response](https://github.com/elastic/elasticsearch/blob/6.x/server/src/main/java/org/elasticsearch/action/admin/cluster/snapshots/get/GetSnapshotsResponse.java#L93-L102).

wesleyk:
Thanks for the quick reply! In terms of next steps, if I'm understanding, you'd like:

1) To include the rest of the fields onto the request as specified by the REST API spec.
2) Confirm on the java source side that the response is as expected (I've thus far only done this with local testing on my own instance).

Is that correct?

olivere:
Correct.

olivere:
Awesome. Maybe I can put that in 6.2.14 which I plan to release over the weekend. Thanks!

wesleyk:
Thank you for the quick turnaround! Let me know if there's any other changes you'd like to see.

wesleyk:
I added one more test case where snapshots are specified.

PR #995, opened 2018-12-18
Adds support for is_write_index to alias actions

cludden:


olivere:
Looking good. Thanks for contributing. Will be part of 6.2.15.

PR #997, opened 2018-12-24
Fix example code to enable trace log

mas9612:
`*Client.SetTracer` method is no longer exists.
https://github.com/olivere/elastic/blob/46a8b35d25c08bebb767e8de83d081709af95dc3/example_test.go#L40-L41

So fix example code to use `SetTraceLog` method instead of it.


olivere:
Thank you.

PR #998, opened 2018-12-24
Adding SpanNear, SpanMulti and SpanTerm Queries

otherview:
Adding SpanNear, SpanMulti and SpanTerm Queries.

Also adding simple tests on query generation.

### DIFF
--- a/search_queries_span_multi.go
+++ b/search_queries_span_multi.go
@@ -1,0 +1,46 @@
+package elastic
+
+// SpanMultiQuery allows you to wrap a multi term query (one of wildcard, fuzzy, prefix, range or regexp query)
+// as a span query, so it can be nested.
+type SpanMultiQuery struct {
+	query Query
+}
+
+// NewSpanMultiQuery creates and initializes a new SpanMultiQuery.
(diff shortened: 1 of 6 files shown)


olivere:
Thanks for adding these. Unfortunately, they need some more work to be included. There's little things missing (like the Copyright notice in the header of the files), and some bigger things. For the latter, it's almost always to align with the source code of the Elasticsearch server. E.g. `SpanMulti` is [`SpanMultiTermQuery` in Java](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/index/query/SpanMultiTermQueryBuilder.java). We also need to check if the span queries in Java support things that are not included in your PR. Sometimes, they implement things not even mentioned in the official docs.

Can you spend some more time on this PR, or do you want me to complete it? The latter might take some time.

Notice you can use your implementation just fine as it implements the [`elastic.Query` interface](https://github.com/olivere/elastic/blob/release-branch.v6/query.go), and then later switch to the "official" queries.

otherview:
No problem! 

I'll try to sneak some time there to make those changes. :)

olivere:
I will not include Span queries in v6.

PR #1003, opened 2018-12-31
Is write index in aliases api

elafarge:
Elasticsearch recently introduced write indexes for multi-index aliases. [This PR](https://github.com/olivere/elastic/pull/995/files) already adds support for creating writeable aliases that point to multiple indexes. 

However, support for figuring out wether or not an index is the write index of an alias is missing. This commit aims at fixing that.

Note that this PR has been created atop https://github.com/olivere/elastic/pull/995/files since - in order to test that the aliases API returns the `IsWriteIndexOption` as expected - we need to create an alias with the `IsWriteIndex` flag set to `true`.


olivere:
Thank you for contributing to elastic!

I'll still have to review the code, but it certainly seems like many people desperately need this feature. So it will very probably be in the next release.

Reference to future self:
* [Write index in docs](https://www.elastic.co/guide/en/elasticsearch/reference/6.5/indices-aliases.html#aliases-write-index)
* [Code in 6.x branch](https://github.com/elastic/elasticsearch/blob/6.x/server/src/main/java/org/elasticsearch/action/admin/indices/alias/IndicesAliasesRequest.java)

elafarge:
Thanks a lot @olivere :raised_hands:

PR #1004, opened 2018-12-31
Add support for extended stats bucket for pipeline aggregations https…

lechnertech:
Add support for extended stats bucket for pipeline aggregations https://www.elastic.co/guide/en/elasticsearch/reference/6.2/search-aggregations-pipeline-stats-bucket-aggregation.html


olivere:
That's looking good, thank you. It even looks complete when [comparing with the Java source [here](https://github.com/elastic/elasticsearch/blob/6.x/server/src/main/java/org/elasticsearch/search/aggregations/pipeline/bucketmetrics/stats/extended/ExtendedStatsBucketPipelineAggregator.java)). I will double-check before releasing 6.2.15, but it will probably in the next release.

PR #1012, opened 2019-01-14
Return the SearchResult, even in the case of EOF error, to allow for root cause analysis

Jpnock:
Upon performing a scrolled search with a scroll size greater than index.max_result_window, Elasticsearch will not process the request and return zero hits, thus leading to a blanket io.EOF error being returned by this client.

As the SearchResult is not returned with the io.EOF error, it is not problematically possible to inspect the root cause of the error, which resulted in us taking a PCAP in order to diagnose the issue.

By returning the SearchResult, along with the error, it allows for inspection of the Shards.Failures list to see the true reason why the query failed.

### DIFF
--- a/scroll.go
+++ b/scroll.go
@@ -347,7 +347,7 @@ func (s *ScrollService) first(ctx context.Context) (*SearchResult, error) {
 	s.scrollId = ret.ScrollId
 	s.mu.Unlock()
 	if ret.Hits == nil || len(ret.Hits.Hits) == 0 {
-		return nil, io.EOF
+		return ret, io.EOF
 	}
 	return ret, nil
 }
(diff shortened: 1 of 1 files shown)


olivere:
Thanks for reporting. I will have a look for the next release (and will backport to v5 as well).

Returning the response seems to be okay. I wonder why ES silently returns, as this should've been an error, no? Anyway, I'll look into this.

kevinconaway:
To add to @sunhay's point, the responses contain the __shards_  metadata which indicates how many shards were queried and if any of them failed:


```
{
  "took": 1,
  "timed_out": false,
  "_shards": {
    "total": 20,
    "successful": 20,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": 0,
    "max_score": null,
    "hits": []
  }
}
```

If some of the shards failed (as we suspect is happening), the current behavior of the client is to return `nil, io.EOF` if the hit count is 0.  We would like to be able to inspect the response returned to see if there were truly no hits returned or if the query is incomplete due to failed shards.

ravitejag:
Hi @olivere may I know when is v6.2.16 being released. Thanks

olivere:
@ravitejag It's currently rather challenging at my day time job, but I hope to get it rolled out over the weekend.

ravitejag:
Cool. Thanks for this awesome lib. 👏

olivere:
Update will happen soon, and this PR is included in v5 and v6.

PR #1014, opened 2019-01-16
fix typo in HighlightQuery API call

elireisman:
Hi! Just noticed a typo in the HighlightQuery API call on in `highlight.go` across several release versions. I suppose folks might already be dependent on the alternate spelling but just in case, I thought I'd submit a PR. Thanks, and once again great library 🙇

### DIFF
--- a/highlight.go
+++ b/highlight.go
@@ -120,7 +120,7 @@ func (hl *Highlight) Fragmenter(fragmenter string) *Highlight {
 	return hl
 }
 
-func (hl *Highlight) HighlighQuery(highlightQuery Query) *Highlight {
+func (hl *Highlight) HighlightQuery(highlightQuery Query) *Highlight {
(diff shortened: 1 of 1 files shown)


olivere:
Oh, too bad. It'll be a breaking change, but one that seems to be okay with me. Recompiling will uncover this, and it's a simple change. Thanks.

olivere:
Ah well, you looked at the `master` branch, which is defunct for about 5 years. You need to look at the release branches for the different versions: `release-branch.v6` is the current version and `release-branch.v5` is still maintained.

elireisman:
Ah crap sorry I added my commit to `master` but that was because I wasn't sure where a multi-version PR goes, should have asked you first sorry! But yeah it definitely was missing that `t` on `highlight` in the 5 and 6 release branches too

PR #1032, opened 2019-02-02
Use %v to format nil argument in Fatalf

ceh:
Trivial suggestion to fix arguments to t.Fatalf in tests.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -388,7 +388,7 @@ func ExampleAggregations() {
 		histogram, found := userBucket.DateHistogram("history")
 		if found {
 			for _, year := range histogram.Buckets {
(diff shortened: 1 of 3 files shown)


ceh:
Fixes the following go vet reports:
```
./search_aggs_test.go:709: T.Fatalf format %q has arg dateHistoRes.Buckets[0].KeyAsString of wrong type *string
./search_aggs_test.go:721: T.Fatalf format %q has arg dateHistoRes.Buckets[1].KeyAsString of wrong type *string
./search_test.go:355: T.Fatalf format %q has arg hit.Source of wrong type *encoding/json.RawMessage
```

PR #1040, opened 2019-02-11
Allow get the number of result by action

Elbandi:
This patch allow get the buld response status by action. Same result as len(r.Updated()), etc. but more performance (mostly in big bulk request)

### DIFF
--- a/bulk.go
+++ b/bulk.go
@@ -334,9 +334,19 @@ type BulkResponseItem struct {
 	GetResult     *GetResult    `json:"get,omitempty"`
 }
 
+// IndexedLen returns num results of "index" actions.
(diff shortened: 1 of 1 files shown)


olivere:
What's wrong with `len(Indexed())` etc.?

olivere:
As for performance: Can you provide a benchmark?

Elbandi:
i havent benchmark :( but just looked into the source: ByAction build a new items array.

if the bulk has >100k items, thats not only lots of cpu (unused memory copy) usage, but also high memory usage ~12Mb (pure sizeof(BulkResponseItem) == 128) too.

Issue #823, opened 2018-07-11
Disable sniffing by default in 7.x: Yes or no?


olivere:
By far the most frequently asked question is why there is an `elastic. ErrNoClient` error: `elastic: no Elasticsearch node available`. The reason for that is [described in detail in the Wiki](https://github.com/olivere/elastic/wiki/Connection-Problems).

It remains an open question if it would be better to 

a) make people hit the wall and make them ask about sniffing (the situation today) or
b) disable sniffing by default in 7.x so that it just works (tm) but would possibly break in production when a single node goes down in your cluster.

Feel free to comment if you have an opinion.

Comradin:
Some days ago, when reading the issue emails, I noticed the same fact. Every second issue was answered with the hint to disable sniffing or understand und use it correctly.

Maybe add a check for the node address and if it is localhost, 127.0.0.1, ::1 (ipv6) and then disable it by default?

My guess is, that the error triggers most times on someones computer during development. And at the very moment, when the code hits production, sniffing would be a great option, someone would like to have active.

But at least, with disabled sniffing, the code would run in production with the specified node, while it wouldn't fail any longer in development?

olivere:
@Comradin Yes, it would work in both environments. However, people might start to claim that it fails in production which is much worse than failing in development, of course. Maybe it would be a good idea to have the code break at compile time when switching to 7.x so that developers would be forced to make a sensible decision.

I would also like to extract the node-selection and (maybe) the scheduling algorithm, so that everyone can provide their own way to find the nodes of the cluster. Some might simply rely on DNS, some want Elastic to sniff the cluster, some have a static IP pool of nodes etc.

slawo:
I got bit by this yesterday after having not used es for some time. Despite this I really believe it should not be disabled by default.

For those who want to start using the driver with a quick example this flag could be given in the main first example with a link to documentation about it

patgmiller:
I am curious to know if the sniffing and health checks built into the client are solely for the purpose of providing reliability by default. At least that's what the documentation leads me to believe. So that if I am using Elastic to connect to an ElasticSearch cluster which is already behind a service in an environment that provides load balancing and health-checks, like kubernetes for example, then the sniffing feature should be disabled, am I correct?

olivere:
Correct. If you already have a mechanism to provide load-balancing/healthchecks, you should turn sniffing off. E.g. you could use haproxy or nginx.

olivere:
[...]

PR #1043, opened 2019-02-18
Initial support for Elasticsearch 7

telendt:
Since Elasticsearch 7-beta1 came out recently I wanted to give it a try (integrate with my app). Since I mainly use search API in my app all I had to do was to fix `SearchHits.TotalHits`, but I thought I will fix other things as well. The things I touched were mainly related to [types removal](https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html#_schedule_for_removal_of_mapping_types) or removal of previously deprecated functionalities.

Speaking of deprecated functionalities, maybe it make sense to modify `.travis.yml` so that deprecation logs from Elasticsearch (that runs in `docker-compose`) are shown? The client should avoid using deprecated functionalities if possible....

It's not really meant to be merged to `release-branch.v6`, but rather a new `release-branch.v7` should be formed.
I know it's too early to release v7 (as there's still no Elasticsearch 7 GA release) and that you like to do some API fixes with each new release, but maybe this can help you somehow. All tests pass on my machine.


olivere:
Awesome. Thanks for your work. I’ll look into it ASAP.

CHANGELOG-7.0.md
```diff
@@ -0,0 +1,5 @@
+# Changes from 6.0 to 7.0
+
+See [breaking changes](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/breaking-changes-7.0.html).
+
+## ???
```
telendt: Not sure what to put here really.

olivere: I’ll probably have the chance to look at this on Friday afternoon. Looking forward to it.

olivere:
I’ll probably have the chance to look at this on Friday afternoon. Looking forward to it.

CHANGELOG-7.0.md
olivere: i used this file to add additional comments to the list of breaking changes that are specific to this library, e.g. changes in the Go API.

telendt:
I changed status of this PR from "draft" to regular one as otherwise Travis tests don't run.
I do realize this PR is quite big, so here's what I've done:
- I've updated all links to ES doc to v 7.x (that always redirects to latest release of 7.x branch)
- I'm unmashalling `hits.total` to struct (as it changed from int to object)
- I removed types from all test mappings and some API calls that were failing tests; for those services that validate presence of type (and require it construct URL from template) I set default `_doc` value in "constructor functions"; I have a feeling that this task is not fully done and one would need to revisit all *Service APIs that allow for setting type and make a decision what to do there...
- I had to replace all `Flush()` with `Refresh()` in tests to make all index changes "searchable" (flush no longer results in refresh in 7.x)
- I had to fix some date formats as there's a [difference between `YYYY` and `yyyy` in Java](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html) and some aggregation test were breaking (I believe it's related to their migration from Joda time to java time)
- watcher restart seems to be gone, so I removed it
- there were some previously deprecated params in use (`_source_exclude`, `_source_include`) that got removed, so I replaced it with the proper params (`_source_excludes`, `_source_includes`)
- I accidentally noticed that tasks API take `nodes` param and not `node_id`, so I fixed it.

Also, `TestXPackWatcher` test seems to be flaky (I haven't touched it in my last commit, but it broke the build).

I thing the "type" situation if far from perfect - some *Service APIs might require user to manually specify `.Type("_doc")` but that should be a default (explicitly calling it should actually be deprecated, and only used when accessing an index created in 6.x with different default type name). I'll take a look at that when I'll find a time, although I'm not even familiar with many of those APIs myself.

olivere:
Sorry for the delay. I'm still looking at this, but work has been crazy lately. I'm sorry, but I hope I can get it done before 7.0 GA.

telendt:
Understood.
Speaking of type situation I would take a look how official High Level Java client handles those cases and would align with them here. I'll try to do that this weekend.

olivere:
Yes.  The Java code and API has always been my reference when in doubt.

telendt:
So I looked how they deal with deprecating types in ES Java client. In some places they return `"_doc"` (`MapperService::SINGLE_MAPPING_NAME`) in `type()` getter when type is `null`, in others they set default type value to `"_doc"` (right with field declaration or in constructor). You don't define getters for fields of `*Service` types so I went with `return &XXXService{typ: "_doc", ...}` in `NewXXXService` "constructor functions".

My last 3 commits add some missing type handling, add deprecation comments and migrate away from some deprecated APIs. I think this PR is ready at this point.

telendt:
Some side topic - I noticed that there are quite some `golint` warnings. They mostly complain about doc strings and field names - some of which are public members. How do you think about fixing those with v7 release? If you don't have anything against I can help with that.

olivere:
I think this should really be your release. Thanks for your support, once again.

My contacts at Elastic didn’t notify me yet, so I’m not really sure. But it could be about that time, you’re right.

olivere:
One thing I still have to do is fix some remaining issues in ES 6 and port them forward to the 7.x release.

olivere:
Re the lint issues and the docs: If it’s not too much work, I’m all for fixing them. The links in the docs should at least be fixed from 6.x to 7.x. I think I wrote a URL checker for that once; maybe I’ll still have it somewhere.

olivere:
Regarding the linter issues about Id vs ID, I think we should keep the former even though the latter is preferred by the community. I think at this stage of the project, I prefer stability over style.

telendt:
> I think at this stage of the project, I prefer stability over style. What do you think?

No strong opinion really. But yeah, linting warnings are quite weak reason for breaking backward-compatibility of an API...

telendt:
rebased

olivere:
We're getting there :-)  https://github.com/olivere/elastic/tree/release-branch.v7

I just added Go modules support, so you need to import `github.com/olivere/elastic/v7` for the v7 release. It probably doesn't work as we didn't push the `v7.0.0` tag. I have to look up how to do pre-releases with Go modules (if there is a way at all).

olivere:
@telendt What's you opinion on #823?

telendt:
Great! I've just added my opinion on #823. If you don't mind I will close this PR now.

olivere:
Thanks so much for your work on this.

olivere:
Just FYI: [7.0.0 was released today](https://www.elastic.co/blog/elastic-stack-7-0-0-released) and I will tag 7.0.0 shortly. Thanks again.

telendt:
I'm happy I could help.

PR #1051, opened 2019-02-26
Add SearchSource to TopHitsAggregation

ANerd:
This PR adds a `SearchSource` function to `TopHitsAggregation`, similar to the one exposed by `SearchService`. This makes it easier for users of the library to reuse code between the two cases.


olivere:
Thanks.

PR #1053, opened 2019-03-03
Add ability to specify filter path to the Bulk service

Yopi:
We can already do it for the Search and Scroll services,
and as per https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html
all REST APIs support specifying the filter path.

We would want to do this since if we are bulking a large amount of requests,
all of them needs to be serialized and passed over the network even if
we do not care about the result.

### DIFF
--- a/bulk_test.go
+++ b/bulk_test.go
@@ -524,6 +524,38 @@ func TestBulkContentType(t *testing.T) {
 	}
 }
 
+func TestBulkWithFilterPath(t *testing.T) {
+	client := setupTestClientAndCreateIndex(t)
+
+	tweet1 := tweet{User: "olivere", Message: "Welcome to Golang and Elasticsearch."}
+	tweet2 := tweet{User: "sandrae", Message: "Dancing all night long. Yeah."}
+
+	index1Req := NewBulkIndexRequest().Index(testIndexName).Type("doc").Id("1").Doc(tweet1)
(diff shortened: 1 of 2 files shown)


Yopi:
Hopefully I did this correctly! The contributing guidelines didn't specify exactly where to a PR should be merged.

olivere:
No worries. We have `release-branch.vX` where `X` is currently `6`. `release-branch.v5` is still supported, the next version will be `release-branch.v7`.

Looking at the code now, thanks for your contributions.

bulk_test.go
```diff
@@ -524,6 +524,38 @@ func TestBulkContentType(t *testing.T) {
...
+		t.Errorf("expected bulkResponse to be != nil; got nil")
+	}
+
+	// 0 because it was filtered out.
+	if want, got := 0, len(bulkResponse.Items); want != got {
+		t.Errorf("expected len(BulkResponse.Items) = %d; got %d", want, got)
+	}
+
+	// Should be more than 0 since it wasn't filtered out.
+	if want, got := 0, bulkResponse.Took; got == want {
```
olivere: Shouldn't that be `got > want`?

Yopi: I think it's correct. Since we are not filtering out `Took`, it should be more than 0. If it's 0, i.e. `0 == 0`, or `got == want` then it should fail.

Although, I agree that the naming is off. It should be something like "doNotWant" 😅

Issue #939, opened 2018-10-13
Geo shape query


slifer2015:
Hi and thanks for awesome library
u seem missing the geo shape query
can u add the support for it

olivere:
`GeoShapeQuery` was last discussed in #276. It turned out that it was the most complex thing we'd had to do if you had all the builders that are also supported by the Java source.

A workaround is to use e.g. [`RawQuery`](https://github.com/olivere/elastic/blob/release-branch.v6/search_queries_raw_string.go) and or create your own query by implementing the [`Query`](https://github.com/olivere/elastic/blob/release-branch.v6/query.go) interface.

cmitchell:
I noticed a while ago that geo query wasn't implemented and implemented my own.  It's by no means comprehensive, but it should get you started, and it follows your code patterns    It's just one file, so please excuse me for not submitting a formal PR.
```
package your_package

// GeoShapeQuery extends the elastic package in golang, which does not currently implement geo_shape querying.
type GeoShapeQuery struct {
	path    string
	shape   string
	point   []float64
	polygon [][][]float64
}

// NewGeoShapeQuery creates a new elasticsearch geo_shape query
func NewGeoShapeQuery(path string) *GeoShapeQuery {
	return &GeoShapeQuery{
		path: path,
	}
}

// Shape specifies what shape this query represents
func (q *GeoShapeQuery) Shape(shape string) *GeoShapeQuery {
	q.shape = shape
	return q
}

// AddPoint adds a point
func (q *GeoShapeQuery) AddPoint(point []float64) *GeoShapeQuery {
	q.point = point
	return q
}

// AddPolygon adds a polygon
func (q *GeoShapeQuery) AddPolygon(polygon [][][]float64) *GeoShapeQuery {
	q.polygon = polygon
	return q
}

// Source returns JSON for the query.
func (q *GeoShapeQuery) Source() (interface{}, error) {
	source := make(map[string]interface{})

	params := make(map[string]interface{})
	source["geo_shape"] = params

	shape := make(map[string]interface{})
	params[q.path] = shape

	shapeDef := make(map[string]interface{})
	shape["shape"] = shapeDef

	shapeDef["type"] = q.shape
	if q.shape == "point" {
		shapeDef["coordinates"] = q.point
	} else if q.shape == "polygon" {
		shapeDef["coordinates"] = q.polygon
	}

	return source, nil
}
```

olivere:
[...]

PR #1058, opened 2019-03-16
Implement Geo Shape Query

heltonmarx:
Based on the elasticsearch and GeoJSON specification, this PR try to
solve the #939 supporting the follow types :
 * Circle
 * Envelope
 * Linestring
 * MultiLineString
 * MultiPoint
 * MultiPolygon
 * Point
 * Polygon

Any suggestion would be appreciated.

### DIFF
--- a/geo/circle.go
+++ b/geo/circle.go
@@ -1,0 +1,37 @@
+package geo
+
+import (
+	"encoding/json"
+)
+
+// Circle specified by a center point and radius with units,
+// which default to meters.
+type Circle struct {
+	Type        string    `json:"type"`
+	Radius      string    `json:"radius"`
+	Coordinates []float64 `json:"coordinates"`
+}
+
+// NewCircle creates and initializes a new Circle.
+func NewCircle(radius string, coordinates []float64) *Circle {
+	return &Circle{
+		Type:        TypeCircle,
+		Radius:      radius,
+		Coordinates: coordinates,
+	}
+}
+
+// UnmarshalJSON decodes the circle data into a GeoJSON geometry.
+func (m *Circle) UnmarshalJSON(data []byte) error {
+	raw := &rawGeometry{}
+	if err := json.Unmarshal(data, raw); err != nil {
+		return err
+	}
+	return m.decode(raw)
+}
+
+func (m *Circle) decode(raw *rawGeometry) error {
+	m.Type = raw.Type
+	m.Radius = raw.Radius
+	return json.Unmarshal(raw.Coordinates, &m.Coordinates)
+}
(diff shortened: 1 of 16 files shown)


geo/circle.go
```diff
@@ -0,0 +1,37 @@
...
+
+// Circle specified by a center point and radius with units,
+// which default to meters.
+type Circle struct {
+	Type        string    `json:"type"`
+	Radius      string    `json:"radius"`
+	Coordinates []float64 `json:"coordinates"`
+}
+
+// NewCircle creates and initializes a new Circle.
```
avamingli: is there a method to verify the arguments? 
for example:  the order of coordinates, Lon and Lat or Lat and Lon.

heltonmarx: Good point, no, there's not, I will implement it.

olivere: There are some functions like this in [`geo_point.go`](https://github.com/olivere/elastic/blob/release-branch.v6/geo_point.go).

olivere: They are roughly based on [`GeoPoint.java`](https://github.com/elastic/elasticsearch/blob/6.7/server/src/main/java/org/elasticsearch/common/geo/GeoPoint.java).

olivere:
A few initial comments on this PR:

- Don't use documentation links with `current` or `N.x` (e.g. `6.x`). These refer to the current version or the working version for 6.x. For the former, we refer to Elasticsearch version 12.0 in ~5 years. For the latter, we refer to the state the technical writers at Elasticsearch use for deployment before releasing the actual version of the docs. Use a full version number like `6.7` in the links as that won't change, and `6.7` is actually very good because that's probably the last version in the `v6` time frame.
- What is `rawGeometry` and why do we need this? Seems like a code smell to me. I'd prefer to have e.g. `Circle` be separate from `Point`, even if they have some things in common.
- Do we really need `TypePoint`, `TypeCircle` etc.?
- I don't like the fields of e.g. `Circle` being exported and still have a constructor-like function like `NewCircle`.
- Please remove that `"github.com/stretchr/testify/require"` dependency. I'm only using `"testing"` from standard library and [`github.com/google/go-cmp/cmp`](https://github.com/google/go-cmp), and I work actively against pulling in any additional dependencies.
- The `WithXXX` functions are a nice idea. What about a separate design where you would use `MarshalJSON` on the individual shapes to add the missing `type` field to the serialized JSON. Wouldn't it be nice to write something like this from the client's perspective?

```
c := geo.Circle{Center: geo.PointFromLatLon(40.513, -70.119), Radius: 3*geo.Kilometer}
```

olivere:
Can we use this PR instead of #956, i.e. do you try to implement the same thing?

PR #1084, opened 2019-04-25
Fix NoStoredFields to not yield any stored fields

cjgu:
Currently `NoStoredFields()` has no effect as when the `storedFieldNames` is `nil` as no `stored_fields` property gets sent to Elasticsearch and thus all fields are returned. See https://github.com/olivere/elastic/blob/0534a7b1bf47b1ccf57e905491a641709f8a623d/search_source.go#L406-L413

This PR changes NoStoredFields to set `storedNamesFields` to an empty string slice and thereby send `"stored_fields": []` to Elasticsearch.


olivere:
Thank you. Will release 7.0.3 soon.

PR #1089, opened 2019-05-01
Implement XPack Security user-management API

sburton84:
These changes add support for the user-management functions of the [X-Pack security API](https://www.elastic.co/guide/en/elasticsearch/reference/7.0/security-api.html#security-user-apis).


olivere:
Sorry. While helpful, this is not ready for merging. E.g. there is no ChangeUserPassword API, it's the Change Passwords API. Also, you mixed some internal import paths into it and confused v6 with v7. Also, you made APIs like the Change Passwords API dependent on the caller providing the body; that is purely optional and the API should be usable without the user serializing the body.

I opened up a new issue #1113 for tracking the various XPack Security APIs as they are added to v7.

PR #1105, opened 2019-05-31
Index Lifecycle Management Policy API

aarontami:
Add some API calls for Index Lifecycle Management:
1. Create a policy
2. Get a policy
3. Delete a policy

These API calls are also present in the Java client.

Index Lifecycle Management is supported in version 6.7 and above (non beta), so adding this to v6.


aarontami:
@olivere

aarontami:
Travis seems to not be accepting any _ilm calls, which makes me think maybe version 6.7 isn't running on it?

olivere:
Happy to see this integrated into Elastic. The problem with the ILM API is that it's not available in the open source version AFAIK. In v6 and v7 we have [two docker images running in TravisCI](https://github.com/olivere/elastic/blob/release-branch.v7/docker-compose.yml) ([here's v6](https://github.com/olivere/elastic/blob/release-branch.v6/docker-compose.yml): the OSS docker image and the platinum version, available at different ports and with different user/password. You need to use the correct client; see the tests in e.g. [XPack Security API](https://github.com/olivere/elastic/blob/release-branch.v6/xpack_test.go#L108).

What I'm not sure is if ILM is available in the Platinum image with a trial license installed. If we can't test the XPack ILM API, we need to skip the tests on TravisCI, e.g. by checking if the given feature is available:

```
	xi, err := client.XPackInfo().Do(context.Background())
	if err != nil {
		t.Fatal(err)
	}
	if !xi.Features.Security.Enabled {
		t.Skip("skip due to deactivated xpack security")
	}
```

olivere:
Oh. According to https://www.elastic.co/de/subscriptions we should be able to at least test XPack ILM API. We should probably enable it in `docker-compose.yml` with `- xpack.ilm.enabled=true`.

aarontami:
Thanks for the review.

- Changed the file names to start with xpack_* as indicated. Also changed all the class names to start with XPack, as seemed to be the convention with all the other XPack apis.
- Changed the test to use 9210 like the other xpack tests, so it will hit the other travis docker image
- Enabled ilm in the docker yml

xpack_ilm_put_lifecycle.go
```diff
@@ -0,0 +1,163 @@
...
+	"context"
+	"fmt"
+	"net/url"
+
+	"github.com/olivere/elastic/uritemplates"
+)
+
+// See the documentation at
+// https://www.elastic.co/guide/en/elasticsearch/reference/6.7/ilm-put-lifecycle.html
+
```
olivere: Can you remove that blank line, please?

aarontami: Done.

aarontami:
@olivere thanks for quick review, do you have any idea when you'll cut new releases so we can start using this?

Thanks

olivere:
@aarontami It'll probably take at least until the weekend.

PR #1118, opened 2019-06-17
Using HTTPClienter interface instead net/http.Client

artezh:


olivere:
Seems like a good idea. I will think about it. Not sure why there’s no Doer interface in net/http. The net/http didn’t abstract anything Client-related—only RoundTripper.

artezh:
Your interface name is better)
I think that package net/http don't have Client interface because interface declare near consumer

PR #1125, opened 2019-06-25
Cannot reconnect if address changes

itizir:
ES nodes may have their address changed while keeping their ID, in which cases a client may not be able to reconnect to it. This happens for instance with ES nodes deployed as kubernetes pods, after a rolling update.

### DIFF
--- a/client.go
+++ b/client.go
@@ -723,7 +723,7 @@ func (c *Client) updateConns(conns []*conn) {
 	for _, conn := range conns {
 		var found bool
 		for _, oldConn := range c.conns {
-			if oldConn.NodeID() == conn.NodeID() {
+			if oldConn.NodeID() == conn.NodeID() && oldConn.URL() == conn.URL() {
 				// Take over the old connection
 				newConns = append(newConns, oldConn)
 				found = true


olivere:
While I see what you're after, I want this to be thoroughly tested. We have clients in a large variety of environments, and this particular area of code might be highly fragile. I think I've copied this code over from an official driver, so it might be worth to take a look at those again.

olivere:
I _think_ your change is okay. Worst case, it might lead to node re-joining the cluster although that's not necessary, which is annoying but not disastrous.

olivere:
Notice that you somehow changed the code in the `master` branch. The `master` branch is defunct. You had to change the code in the `release-branch.vX` where `X = [5,6,7]`. I will do so for the `v6` and `v7` branch manually.

itizir:
Ah nice, thanks for pulling this in.

Yeah, I believe the change is fairly minimal and harmless, but I also understand you would want it to be tested as much as possible. If you think it would be useful, I'm happy to look into at least get in some unit tests asserting the new behaviour.

Oh sorry about picking `master` as the target branch. I didn't know which branches were still active and how you were maintaining the different versions. Do you want me to cherry-pick the commit to other branches (v5 and earlier, from whichever would require the fix)?

olivere:
Unit tests are always nice, so feel free if you have the time.

Backporting into v5 is a non-issue. ES 5.x is no longer supported, so is the client. I'm only backporting serious issues like e.g. security-related things.

PR #1127, opened 2019-06-26
Update go.opencensus.io to 0.20.1

kernle32dll:
This should(tm) be a small update of go.opencensus.io. The reason for the update is a circular-dependency in older versions, leading to https://github.com/golang/lint/issues/436 . The particular version updating to is taken from https://github.com/googleapis/google-cloud-go/issues/1359#issuecomment-480988126.


olivere:
Ah, thanks for this. Will merge into 7.0.3.

kernle32dll:
Neat! But uh, did you mean 7.0.4? Because 7.0.3 already seem to be released.

olivere:
Yes, 7.0.4 of course.

PR #1128, opened 2019-06-26
fix(ingest-pipeline): use correct schema in GET

jfroche:
According to https://www.elastic.co/guide/en/elasticsearch/reference/7.x/get-pipeline-api.html

### DIFF
--- a/ingest_get_pipeline.go
+++ b/ingest_get_pipeline.go
@@ -116,6 +116,8 @@ func (s *IngestGetPipelineService) Do(ctx context.Context) (IngestGetPipelineRes
 type IngestGetPipelineResponse map[string]*IngestGetPipeline
 
 type IngestGetPipeline struct {
-	ID     string                 `json:"id"`
-	Config map[string]interface{} `json:"config"`
+	Description string                   `json:"description"`
+	Version     int                      `json:"version,omitempty"`
+	Processors  []map[string]interface{} `json:"processors"`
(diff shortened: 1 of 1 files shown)


olivere:
Okay. So in 6.x it was [like this](https://github.com/elastic/elasticsearch/blob/6.8/server/src/main/java/org/elasticsearch/ingest/PipelineConfiguration.java#L111-L118), and in 7.x it is [the same](https://github.com/elastic/elasticsearch/blob/7.2/server/src/main/java/org/elasticsearch/ingest/PipelineConfiguration.java#L110-L117).

I'm not sure with this PR as there seems to be an offset between ES documentation and the actual ES code.

olivere:
I finally checked and your PR is correct. I will release this in 7.0.5, due soon. Thanks again.

jfroche:
Thank you. Do you think it is a good idea to back port that change in v6 and v5 ?

phillbaker:
For reference & context, I think the java code is in [Pipeline.java](https://github.com/elastic/elasticsearch/blob/963c9b8092a6ba95d85c7d97b8a2d43169832367/server/src/main/java/org/elasticsearch/ingest/Pipeline.java#L39-L42), not `PipelineConfiguration.java`.

The same structure is also used in [v5](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/get-pipeline-api.html).

olivere:
> Thank you. Do you think it is a good idea to back port that change in v6 and v5 ?

Yes, that's probably a good idea.

PR #1133, opened 2019-07-05
Testing node changing its address

itizir:
Hey again! So here's a possible unit-test to complement https://github.com/olivere/elastic/pull/1125.
Feel free to do as you wish with it.


olivere:
Thanks for the code. I've adopted and changed it a bit (using `net/http/httptest` instead), but the overall test is fine.

PR #1139, opened 2019-07-11
Return `seq_no` stats with `IndicesStatsResponse`.

emil2k:
These stats provide access to the shard level sequences numbers such as
`max_seq_no`, `global_checkpoint`, and `local_checkpoint`. They can be
used to query whether the state of the index has changed or converged.

### DIFF
--- a/indices_stats.go
+++ b/indices_stats.go
@@ -236,6 +236,7 @@ type IndexStatsDetails struct {
 	Translog    *IndexStatsTranslog    `json:"translog,omitempty"`
 	Suggest     *IndexStatsSuggest     `json:"suggest,omitempty"`
 	QueryCache  *IndexStatsQueryCache  `json:"query_cache,omitempty"`
+	SeqNo       *IndexStatsSeqNo       `json:"seq_no,omitempty"`
 }
 
 type IndexStatsDocs struct {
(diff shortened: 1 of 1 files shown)


olivere:
This is looking good. I can see [`SeqNoStats`](https://github.com/elastic/elasticsearch/blob/7.2/server/src/main/java/org/elasticsearch/index/seqno/SeqNoStats.java#L77-L85) in the 7.2 tree.

It seems other parts of the `IndexStatsResponse` have changed as well, e.g. `retentionLeaseStats` have been added in 6.7+ ([see here](https://github.com/elastic/elasticsearch/blob/7.2/server/src/main/java/org/elasticsearch/action/admin/indices/stats/ShardStats.java#L154-L179) for full shard stats).

emil2k:
Hey @olivere, 

I added the `retention_leases` stats and got the shard stats up to date with their 6.8 release branch:
https://github.com/elastic/elasticsearch/commits/6.8/server/src/main/java/org/elasticsearch/action/admin/indices/stats/ShardStats.java

olivere:
I need to do forward-port this to the `release-branch.v7` as well when releasing the next version.

I'll do when I'm preparing 7.0.5.

PR #1146, opened 2019-07-23
Add rest_total_hits_as_int to Search and Msearch API

gregoryfranklin:
https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#hits-total-now-object-search-response

To aid migration from 6.x to 7.x.

In 7.x hits.total in the search response is an object instead of an int.

rest_total_hits_as_int can be used to tell a 7.x server to return hits.total as a number so that it is compatible with the 6.x search response format.  By merging this change into the v6 branch, it will be possible to use the v6 version of the client against both 6.x and 7.x servers whilst migrating to 7.x

### DIFF
--- a/msearch.go
+++ b/msearch.go
@@ -20,6 +20,7 @@ type MultiSearchService struct {
 	pretty                bool
 	maxConcurrentRequests *int
 	preFilterShardSize    *int
+	restTotalHitsAsInt    *bool
 }
 
 func NewMultiSearchService(client *Client) *MultiSearchService {
@@ -54,6 +55,11 @@ func (s *MultiSearchService) PreFilterShardSize(size int) *MultiSearchService {
 	return s
 }
 
+func (s *MultiSearchService) RestTotalHitsAsInt(v bool) *MultiSearchService {
+	s.restTotalHitsAsInt = &v
+	return s
+}
+
 func (s *MultiSearchService) Do(ctx context.Context) (*MultiSearchResult, error) {
 	// Build url
 	path := "/_msearch"
@@ -68,6 +74,9 @@ func (s *MultiSearchService) Do(ctx context.Context) (*MultiSearchResult, error)
 	}
 	if v := s.preFilterShardSize; v != nil {
 		params.Set("pre_filter_shard_size", fmt.Sprintf("%v", *v))
+	}
+	if v := s.restTotalHitsAsInt; v != nil {
+		params.Set("rest_total_hits_as_int", fmt.Sprintf("%v", *v))
 	}
 
 	// Set body
--- a/search.go
+++ b/search.go
@@ -17,21 +17,22 @@ import (
 
 // Search for documents in Elasticsearch.
 type SearchService struct {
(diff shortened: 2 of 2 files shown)


olivere:
I will review how to do this. From first sight, it looks like you successfully added the parameter, but didn't check if the search results counters are still returned accordingly. Anyway, will put it on the list for the next release.

pavolloffay:
@olivere any update on this?

olivere:
@pavolloffay Unfortunately I'm rather busy with my day job. It's still on the list of things I might review for the next release.

I'm not entirely sure that this is the way to go, unfortunately. Although I haven't looked into this in detail, `rest_total_hits_as_int` is a parameter for ES 7.x. You want to add it to ES 6.x. That leaves me with the question: Why? Because elastic.v6 is meant to be used with ES 6.x, and elastic.v7 is meant to to be used with ES 7.x. What is not supported is using elastic.v6 with ES 7.x.

pavolloffay:
We are using this client in Jaeger project https://www.jaegertracing.io/ to talk to Elasticsearch storage. Right now we are using 5.x client, which happens to be compatible with ES 5 and 6 for our use case. However it does not work with ES 7.x

We would like to add support for ES7, but at the same time keep the compatibility with older Elasticsearch versions - 6.x and preferably also 5.x.

olivere:
@pavolloffay Hmm... this was never expected to work, so you probably were just lucky or used a very small subset of the functionality of this client. Problem with `rest_total_hits_as_int` is that is changes the response of ES, and I don't like the idea of type-checking and breaking compatibility as then [this line](https://github.com/olivere/elastic/blob/release-branch.v7/search.go#L515) would be `int64` when `rest_total_hits_as_int=true`, but `*TotalHits` when `rest_total_hits_as_int=false` (or not specified).

Having said that, maybe switching to [the official client](https://github.com/elastic/go-elasticsearch/) would be a smart move anyway, as elastic.v7 will be the last version before this library will be deprecated. If you use only a small set of feature from ES, building your own HTTP-based client is also not that hard and doesn't pull in a large library like elastic.

gregoryfranklin:
rest_total_hits_as_int makes an ES7 server behave like an ES6 server.
When added to the v6 branch, it makes it possible to use the v6 branch with an ES7 server.  The types in the v6 branch will be correct and don't need to be changed.

rest_total_hits_as_int should not be added to the v7 branch.

olivere:
Yes. We could do that. And I think I will do so in the next release.

However, I'm still not happy with it because I'd add a feature that enables people to use elastic.v6 in a scenario that it's not supposed to be used with.

pavolloffay:
@olivere when are you planning to cut the new release?

olivere:
Maybe I'll find a few minutes later this evening. I'll let you know.

olivere:
Will be released in 6.2.22, which is about to be released.

pavolloffay:
thanks @olivere

Issue #911, opened 2018-09-20
Add AutoDateHistogramAggregation


olivere:
See https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/AutoDateHistogramAggregationBuilder.java

PR #1150, opened 2019-07-29
Implement auto_date_histogram like date_histogram.

soarpenguin:


olivere:
Thanks for helping with #911. The query side seems okay (although we will probably have to ignore `keyed`), but the reading side is missing, see [`search_aggs.go`](https://github.com/olivere/elastic/blob/release-branch.v7/search_aggs.go). Also, tests are missing as well.

I don't know why the other files are part of the PR, probably due to formatting issues. I will gofmt in a seperate commit, so we should remove that.

For self-reference: https://github.com/elastic/elasticsearch/blob/7.2/server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/AutoDateHistogramAggregationBuilder.java

soarpenguin:
Ok，i will fix soon.

PR #1181, opened 2019-09-09
Add Highlight option to ScrollService

jgimenez:
The library implements highlighting for the SearchService but not for the ScrollService. Here is an addition for that.


olivere:
Thanks. I will need to port this to 7.x as well.

and1truong:
@olivere https://github.com/olivere/elastic/pull/1190

olivere:
Added to both 7.0.7 and 6.2.24.

PR #1182, opened 2019-09-11
Add rollup search paramater

wuxinwei:
Based on the elasticsearch 7.0 rollup Restful API (https://www.elastic.co/guide/en/elasticsearch/reference/7.0/rollup-search.html)
Changes
- Add rollup search paramater Into `Search Service`
- Add `Rollup` method to indicate this search request is searching for rollup index

Usage
```
func TestRollupSearch(t *testing.T) {
	esCli = GetEs()

	amountAggs := es7.NewSumAggregation().Field("amount")
	typeAggs := es7.NewTermsAggregation().Field("type").SubAggregation("amount", amountAggs)

	ss := esCli.Search().
		Index("transaction-summary").
		IgnoreUnavailable(true).
		Size(0).
		Rollup(true).
		Pretty(true).
		Aggregation("type", typeAggs)

	_, err := ss.Do(context.Background())
	if err != nil {
		t.Logf("%s", err)
	}
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,7 @@ Here are a few tips on how to get used to Elastic:
 - [x] Explain API
 - [x] Profile API
 - [x] Field Capabilities API
+- [x] Rollup Search API
 
 ### Aggregations
 
--- a/search.go
+++ b/search.go
@@ -7,6 +7,7 @@ package elastic
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/url"
 	"reflect"
@@ -21,6 +22,7 @@ type SearchService struct {
 	searchSource      *SearchSource
 	source            interface{}
 	pretty            bool
+	rollup            bool
 	filterPath        []string
 	searchType        string
 	index             []string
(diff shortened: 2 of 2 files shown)


olivere:
I'm sorry, but Rollup Search API is defined as a completely separate REST endpoint, in the [`rollup.rollup_search.json`](https://github.com/elastic/elasticsearch/blob/9166311622/x-pack/plugin/src/test/resources/rest-api-spec/api/rollup.rollup_search.json) spec.

wuxinwei:
> I'm sorry, but Rollup Search API is defined as a completely separate REST endpoint, in the [`rollup.rollup_search.json`](https://github.com/elastic/elasticsearch/blob/9166311622/x-pack/plugin/src/test/resources/rest-api-spec/api/rollup.rollup_search.json) spec.
@olivere 

yeah，I see，but I need a advice about using rollup search， If I want use rollup search in elastic sdk，what should I do？thank you

olivere:
As long as we don't have a service for X-Pack Rollup Search, the easiest way is to use `Client.PerformRequest` directly and generate the JSON directly in your application. `PerformRequest` is used by all services, so you're only missing the DSL.

The best solution would be to create a service/DSL for that API, and add it to all the other services we already have.

PR #1187, opened 2019-09-17
Range query: Follow latest DSL

and1truong:
DSL for branch 7.x already changed, from/to/include_lower/include_upper are gone:

https://www.elastic.co/guide/en/elasticsearch/reference/7.0/query-dsl-range-query.html

### DIFF
--- a/search_aggs_bucket_filter_test.go
+++ b/search_aggs_bucket_filter_test.go
@@ -21,7 +21,7 @@ func TestFilterAggregation(t *testing.T) {
 		t.Fatalf("marshaling to JSON failed: %v", err)
 	}
 	got := string(data)
-	expected := `{"filter":{"range":{"stock":{"from":0,"include_lower":false,"include_upper":true,"to":null}}}}`
+	expected := `{"filter":{"range":{"stock":{"gt":0}}}}`
 	if got != expected {
 		t.Errorf("expected\n%s\n,got:\n%s", expected, got)
 	}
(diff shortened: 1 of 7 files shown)


olivere:
Although I agree that `gt` and `gte` are more intuitive, the old way of doing things are still [alive and kicking](https://github.com/elastic/elasticsearch/blob/7.4/server/src/main/java/org/elasticsearch/index/query/RangeQueryBuilder.java#L321-L341).

I won't merge this PR in its current form as it breaks existing clients that [rely on `IncludeUpper` and `IncludeLower`](https://github.com/olivere/elastic/blob/release-branch.v7/search_queries_range.go#L75-L87).

olivere:
I've checked 6.8.3 and 7.4.0, and both seem to be able to parse `gt` and `gte` correctly, so I think we can move forward with this change. Both seem to be [supported since at least 6.0.0](https://github.com/elastic/elasticsearch/blob/v6.0.0/core/src/main/java/org/elasticsearch/index/query/RangeQueryBuilder.java#L376-L387).

houz42:
@olivere thanks for your great work, and I wonder why this pr is open for nearly a year after all check passed. IMO this is a really important change, and at least in ES 7.10, the old params are already deprecated. We really need this when using newer servers. Is there any more work need to do, or any help could be useful?

olivere:
@houz42 Thanks for the kind words.

The reason is that there was just no need to rush. Even the Java code in 7.15 generates the range fields with the old syntax. [See for youself](https://github.com/elastic/elasticsearch/blob/v7.15.0/server/src/main/java/org/elasticsearch/index/query/RangeQueryBuilder.java#L312-L315). So in the end, it's just what happens on the wire. The actual outcome is the same.

PR #1190, opened 2019-09-17
7.x: Add Highlight option to ScrollService.

and1truong:


olivere:
Added to both 7.07 and 6.2.24.

PR #1191, opened 2019-09-17
Adding the support for Request Headers to the Elastic go client.

arpiagar:
Many of the ServiceMesh implementation rely on HTTP Request headers
for service discovery.
This commit is a step to add the support of service disovery via the
available ServiceMesh via HTTP request headers propogation.


and1truong:
I think we can define default headers by custom HTTP client?

https://github.com/olivere/elastic/wiki/Configuration

arpiagar:
Yes, we can and that should solve for the ServiceMesh discovery problem. However, there are  still cases where we would need Request Headers support for introducing failure injections via ServiceMesh and for that custom Request Headers might need to be set

olivere:
I agree that having a http.Header for each service/endpoint is a good idea and doesn’t destroy anything. Once I have the time, I will review and merge this. Thanks for your patience.

arpiagar:
Any update on this ?

olivere:
I will add this in the next release. Currently I can only work on elastic on the weekends, so things slow down a bit. Sorry.

olivere:
Please use #1200 for further discussion.

PR #1193, opened 2019-09-23
Fixes #1192 Close every valid res.Body to stop network leak

### LINKED ISSUE #1192: Possible network connection leak

[x] elastic.v6 (for Elasticsearch 6.x)

Elasticsearch shouldn't have a network connection leak

On one of our nodes, we are seeing more than a million open network connections to elasticsearch

I believe our requests where being returned in a error state, (not sure the cause).
But when my IT guy said we have a million open network connections to elasticsearch.
I started looking for possible causes the the leak.
I believe I found it.
https://github.com/olivere/elastic/blob/release-branch.v7/client.go#L958
```
       res, err := c.c.Do((*http.Request)(req).WithContext(ctx))
	if err != nil {
		return nodes
	}
	if res == nil {
		return nodes
	}

	if res.Body != nil {
		defer res.Body.Close()
	}
```

This code has the same bug that I discovered a couple years ago in my own applications.
Basically it's possible for the network call Do()  to return err != nil, and res.Body != nil
It seemed to be rare, (many of the exit conditions in Do() ) return a nil Body, but I saw at least on case where it can return an error and a Body.

This code and any other code like it needs to be re-written at
```
       res, err := c.c.Do((*http.Request)(req).WithContext(ctx))
       if res != nil && res.Body != nil {
		defer res.Body.Close()
	}
	if err != nil {
		return nodes
	}
	if res == nil {
		return nodes
	}	
```

nlacey:
Fix network leak when res.Body co-exists with a non-nil err object

### DIFF
--- a/aws/v4/aws_v4.go
+++ b/aws/v4/aws_v4.go
@@ -6,6 +6,7 @@ package v4
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -71,11 +72,17 @@ func (st Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 		return nil, err
 	}
 	resp, err := st.client.Do(req)
+
+	//Make sure resp.Body is closed, if it exists, regardless of error state
+	if resp != nil && resp.Body != nil {
+		defer resp.Body.Close()
+	}
+
 	if err != nil {
 		return nil, err
 	}
+
 	if resp.Body != nil {
-		defer resp.Body.Close()
 		buf := new(bytes.Buffer)
 		_, err = buf.ReadFrom(resp.Body)
 		if err != nil {
--- a/client.go
+++ b/client.go
@@ -949,15 +949,13 @@ func (c *Client) sniffNode(ctx context.Context, url string) []*conn {
 	c.mu.RUnlock()
 
 	res, err := c.c.Do((*http.Request)(req).WithContext(ctx))
+
+	//Make sure resp.Body is closed, if it exists, regardless of error state
+	if res != nil && res.Body != nil {
+		defer res.Body.Close()
+	}
 	if err != nil {
 		return nodes
-	}
-	if res == nil {
-		return nodes
-	}
-
-	if res.Body != nil {
-		defer res.Body.Close()
 	}
 
 	var info NodesInfoResponse
@@ -1342,6 +1340,12 @@ func (c *Client) PerformRequest(ctx context.Context, opt PerformRequestOptions) 
 
 		// Get response
 		res, err := c.c.Do((*http.Request)(req).WithContext(ctx))
+
+		//Make sure res.Body is closed, if it exists, regardless of error state
+		if res != nil && res.Body != nil {
+			defer res.Body.Close()
+		}
+
 		if IsContextErr(err) {
 			// Proceed, but don't mark the node as dead
 			return nil, err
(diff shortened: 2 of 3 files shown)


aws/v4/aws_v4.go
```diff
@@ -71,11 +72,21 @@ func (st Transport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	if resp != nil && resp.Body != nil {
+		defer resp.Body.Close()
+	}
+
 	if err != nil {
 		return nil, err
 	}
+
+	if resp == nil {
+		return nil, fmt.Errorf("Network response was empty")
```
olivere: I do not agree with returning an error here. We should return what `st.Client.Do(...)` returns.

client.go
```diff
@@ -949,17 +949,18 @@ func (c *Client) sniffNode(ctx context.Context, url string) []*conn {
 	c.mu.RUnlock()
 
 	res, err := c.c.Do((*http.Request)(req).WithContext(ctx))
+
+	//Make sure resp.Body is closed, if it exists, regardless of error state
+	if res != nil && res.Body != nil {
```
olivere, line 954: Actually that's not necessary with `*http.Client.Do`. Its documentation [clearly states](https://golang.org/pkg/net/http/#Client.Do):

> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

I'm not sure why it's a problem with the AWS signing transport, but I've never seen an issue with non-AWS clusters (and haven't seen any reports of such problems from others).

olivere: Also, see [`http.Response`](https://golang.org/pkg/net/http/#Response):

```
// The http Client and Transport guarantee that Body is always
// non-nil, even on responses without a body or responses with
// a zero-length body. It is the caller's responsibility to
// close Body.
```

aws/v4/aws_v4.go
nlacey: I concur, I can change that

client.go
nlacey: You are mis-reading Client.Do()  It says if error is nil, than there is a body.
What it DIDN'T say, is if error is NOT nil, than there is no body to close.

Your network leak, can come if you get a 404 error, with a non-nil body, because original code aborted on error (before closing Body)

olivere: The `net/http` gives an explicit example of how to close the response body in its Overview section. The code is:

```
resp, err := http.Get("http://example.com/")
if err != nil {
	// handle error
}
defer resp.Body.Close()
```

olivere: Below that `If the returned error is nil ...` section, documentation goes on with:

> On error, any Response can be ignored. A non-nil Response with a non-nil error only occurs when CheckRedirect fails, and even then the returned Response.Body is already closed.

... which Brad Fitzpatrick [confirms here](https://github.com/golang/go/issues/17780#issuecomment-258293699).

nlacey:
Looks like you are correct, I went through the current net package and can't find a case where body isn't closed on error.

Guess my memory is all the way back from go 1.3, ...
 doFollowingRedirects(ireq *Request, shouldRedirect func(int) bool)
	if redirectFailed {
		// Special case for Go 1 compatibility: return both the response
		// and an error if the CheckRedirect function failed.
		// See http://golang.org/issue/3795
		return resp, urlErr
	}

But actually looking at the code again, it looks like even there, the body was closed?

Thanks for your time

olivere:
No worries. It’s been a long week, and sometimes we can’t see the forest for the trees. What I still wonder is how the AWS transport doesn’t follow those guarantees.

Issue #1202, opened 2019-10-04
Distance Feature Query


raynigon:
Since Elasticsearch 7.2 there is a distance_feature query available, see: https://www.elastic.co/guide/en/elasticsearch/reference/7.2/query-dsl-distance-feature-query.html

### Which version of Elastic are you using?

[x] elastic.v7 (for Elasticsearch 7.x)
[ ] elastic.v6 (for Elasticsearch 6.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v2 (for Elasticsearch 1.x)

### Please describe the expected behavior
Able to using the Distance Feature Query

### Please describe the actual behavior
Distance Feature Query not implemented

olivere:
Seems like we were working on the same issue...

PR #1203, opened 2019-10-04
Implemented Distance Feature Query

raynigon:
Added the Distance Feature Query

See #1202


olivere:
Seems like we were working on the same thing at the same time. Added this already in 13dd156dabeed67c79d06c66661533697c7c8912 and e4cd810912a7387a965286b6e18d06e12dba7f5f.

Issue #1113, opened 2019-06-08
Add XPack Security APIs


olivere:
See [Security APIs](https://www.elastic.co/guide/en/elasticsearch/reference/7.1/security-api.html) and the [REST API spec](https://github.com/elastic/elasticsearch/blob/7.1/x-pack/plugin/src/test/resources/rest-api-spec/api/security.authenticate.json).

While #1089 is a good start, it is incomplete and has a few issues.

- [ ] Authenticate
- [ ] Clear cache
- [ ] Clear cached roles
- [x] Change passwords
- [ ] Create API key
- [ ] Delete privileges
- [x] Delete role
- [x] Delete role mapping
- [x] Delete user
- [x] Disable user
- [x] Enable user
- [ ] Get API key
- [ ] Get privileges
- [x] Get role
- [ ] Get role mapping
- [ ] Get token
- [x] Get user
- [ ] Get user privileges
- [ ] Has privileges
- [ ] Invalidate API key
- [ ] Invalidate token
- [ ] Put privileges
- [x] Put role
- [x] Put role mapping
- [x] Put user

sangeethdba:
@olivere is it possible to have xpack put user for elastic version 6

olivere:
@sangeethdba Sure, that's probably very easy to achieve (if there are not a lot of changes between 6.x and 7.x). To backport, you only need to copy over [`xpack_security_put_user*.go`](https://github.com/olivere/elastic/blob/release-branch.v7/xpack_security_put_user.go) to 6.x and add the [`XPackSecurityPutUser` function](https://github.com/olivere/elastic/blob/release-branch.v7/client.go#L2162-L2165) into `client.go` of 6.x.

If you want to try, I'm happy to get a PR. Otherwise, I'll mark this as an open issue.

sangeethdba:
https://github.com/olivere/elastic/pull/1517

sangeethdba:
@olivere https://github.com/olivere/elastic/pull/1517

olivere:
A lot of them are implemented now. So closing this an creating an issue per missing API.

PR #1204, opened 2019-10-08
feat: implement xpack security user api

jfroche:
Implements:

  * Create/Update User
  * Delete User
  * Get User
  * Disable User
  * Enable User

### DIFF
--- a/xpack_security_get_user.go
+++ b/xpack_security_get_user.go
@@ -1,0 +1,177 @@
+// Copyright 2012-2019 Oliver Eilhard. All rights reserved.
+// Use of this source code is governed by a MIT-license.
+// See http://olivere.mit-license.org/license.txt for details.
+
+package elastic
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/olivere/elastic/v7/uritemplates"
+)
+
+// XPackSecurityGetUserService retrieves a user by its name.
+// See https://www.elastic.co/guide/en/elasticsearch/reference/7.0/security-api-get-user.html.
+type XPackSecurityGetUserService struct {
+	client *Client
+
+	pretty     *bool       // pretty format the returned JSON response
+	human      *bool       // return human readable values for statistics
+	errorTrace *bool       // include the stack trace of returned errors
+	filterPath []string    // list of filters used to reduce the response
+	headers    http.Header // custom request-level HTTP headers
+
+	username []string
+}
+
+// NewXPackSecurityGetUserService creates a new XPackSecurityGetUserService.
+func NewXPackSecurityGetUserService(client *Client) *XPackSecurityGetUserService {
+	return &XPackSecurityGetUserService{
+		client: client,
+	}
+}
+
+// Name is name of the user to retrieve.
+func (s *XPackSecurityGetUserService) Username(username string) *XPackSecurityGetUserService {
+	s.username = []string{username}
+	return s
+}
+
+// Name is name of the user to retrieve.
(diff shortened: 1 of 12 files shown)


jfroche:
refs #1113

olivere:
Looking great. Thank you! Will merge in 7.0.8.

xpack_security_get_user.go
```diff
@@ -0,0 +1,120 @@
...
+// XPackSecurityUser is the user object.
+//
+// The Java source for this struct is defined here:
+// https://github.com/elastic/elasticsearch/blob/7.3/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/user/User.java
+type XPackSecurityUser struct {
+	Username     string                 `json:"username"`
+	Roles        []string               `json:"roles"`
+	Fullname     string                 `json:"full_name"`
+	Email        string                 `json:"email"`
+	Password     string                 `json:"password"`
```
ckaye: I don't think the `Password` and `PasswordHash` are returned in the GetUser API? docs here: https://www.elastic.co/guide/en/elasticsearch/reference/7.3/security-api-get-user.html

olivere: Correct. It seems like [these fields will be serialized](https://github.com/elastic/elasticsearch/blob/7.4/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/user/User.java#L165-L175).

xpack_security_get_user.go
```diff
@@ -0,0 +1,120 @@
...
+	"net/url"
+
+	"github.com/olivere/elastic/v7/uritemplates"
+)
+
+// XPackSecurityGetUserService retrieves a user by its name.
+// See https://www.elastic.co/guide/en/elasticsearch/reference/7.0/security-api-get-user.html.
+type XPackSecurityGetUserService struct {
+	client *Client
+	pretty bool
```
olivere: The other standard fields are missing in this API as well, e.g. `filter_path`, `header` etc.

xpack_security_get_user.go
```diff
@@ -0,0 +1,120 @@
...
+	name   string
```
olivere: The `name` is not a valid parameter for this API. See the [REST API spec](https://github.com/elastic/elasticsearch/blob/7.4/x-pack/plugin/src/test/resources/rest-api-spec/api/security.get_user.json) for a list of parameters: It should be one or more `username`.

olivere (changes requested):
Can you please check that all services have all [common fields](https://github.com/elastic/elasticsearch/blob/7.4/rest-api-spec/src/main/resources/rest-api-spec/api/_common.json) (except `source`) and [the official XPack REST API spec](https://github.com/elastic/elasticsearch/tree/7.4/x-pack/plugin/src/test/resources/rest-api-spec/api) matches the parameters in the Go services in this PR.

xpack_security_get_user.go
jfroche: Thank you for the review,
I have added the common options (based on https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html)

xpack_security_get_user.go
jfroche: Right, I have added username as a list of strings. Are you ok with that ?

xpack_security_get_user.go
jfroche: Indeed, well seen ! Thank you

olivere:
I changed a few things like aligning strongly to the [REST API Spec](https://github.com/elastic/elasticsearch/tree/master/x-pack/plugin/src/test/resources/rest-api-spec/api), keeping names the same everywhere (e.g. `Username`), and adding missing common options.

phillbaker:
@olivere would you accept a PR backporting this to v6?

PR #1205, opened 2019-10-11
add AllowPartialSearchResults for SearchService

NikitaSerenko:


NikitaSerenko:
Hello, olivere!
This little parameter can save my day!
I made it like SearchRequest::allowPartialSearchResults

olivere:
Aw yeah, [here it is](https://github.com/elastic/elasticsearch/blob/6.8/rest-api-spec/src/main/resources/rest-api-spec/api/search.json#L154-L158). I will do this in the next release and do a forward-port to 7.x. When I'm in the process of doing this, I will see if there are other parameters missing as well.

NikitaSerenko:
Thanks for the fast reply, great news to hear!
Can you give an approximate date of release for this feature, please?

olivere:
Maybe over the weekend. But I'm still doing this in my spare time, so I tend to not take deadlines too seriously.

olivere:
Implemented in 6.2.25 and 7.0.8.

Issue #1006, opened 2019-01-07
Fields dropped from Watch


abaehre:
### Which version of Elastic are you using?

[ x ] elastic.v6 (for Elasticsearch 6.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v2 (for Elasticsearch 1.x)

### Please describe the expected behavior
https://github.com/olivere/elastic/blob/release-branch.v6/xpack_watcher_get_watch.go#L74
(the Do function on *XPackWatcherGetWatchService) returns all fields of the watch.

### Please describe the actual behavior
Fields that are not included in the struct XPackWatch (https://github.com/olivere/elastic/blob/release-branch.v6/xpack_watcher_get_watch.go#L118) are truncated during the json.Unmarshal in Do (https://github.com/olivere/elastic/blob/release-branch.v6/xpack_watcher_get_watch.go#L98)

### Any steps to reproduce the behavior?
Here's an example of "throttlePeriod" being dropped in a Go Playground. https://play.golang.org/p/s2d0ikMEDg-

You can also reproduce by doing a "Put" and then a "Get" of a watch with a field that isn't currently in the XPackWatch struct.

### Notes
This should be able to be fixed by adding all fields that a Watch may have to the XPackWatch struct. Doing so will stop the truncation of those fields.

Example fields that are found in https://github.com/elastic/examples - 

https://github.com/elastic/examples/blob/master/Alerting/Sample%20Watches/cpu_iowait_hosts/watch.json#L2

https://github.com/elastic/examples/blob/master/Alerting/Sample%20Watches/errors_in_logs/watch.json#L53

https://github.com/elastic/examples/blob/master/Alerting/Sample%20Watches/filesystem_usage/watch.json#L69

olivere:
Thanks for reporting. That part of the package has been PR'd by a contributor, so I will have to dig into it. Will do so in the next release.

abaehre:
I wouldn't necessarily describe `XPackWatch` as "flexible". Right now because it is a struct it will drop all watcher options not included in the struct. This makes the `Do` method (https://github.com/olivere/elastic/blob/release-branch.v6/xpack_watcher_get_watch.go#L74) unusable   if you need to retrieve any of those options. 

[...]

PR #1211, opened 2019-10-21
XPackWatchStatus.ExecutionState should be string

hstern:
Elastic 7.3.2 and all earlier versions that I have seen do not use an
object for `execution_status`, but instead use a string. Execution state
is not mentioned at all in the current documentation.

#1006 f438c83fc introduced decoding of the `_watches/watch/$NAME` response. 

XPackWatchGetService.Do fails with `json: cannot unmarshal string into
Go struct field XPackWatchStatus.status.execution_state of type
elastic.XPackWatchActionExecutionState`.

This commit changes the type of XpackWatchStatus to string so that the
XPackWatchGetService.Do method succeeds.


olivere:
Correct. In 7.4, [`ExecutionState` gets build here](https://github.com/elastic/elasticsearch/blob/7.4/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/watcher/watch/WatchStatus.java#L260-L262), and only writes the for `ExecutionState.id`.

PR #1214, opened 2019-10-22
Set XPackWatchActionStatus members to type XPackWatchActionExecutionState

hstern:
XPackWatchActionStatus.LastExecution and LastSuccessfulExecution members
were of type `*time.Time`, but Elastic returns an object here. This was
the opposite case with XPackWatchStatus.ExecutionState which we changed
to a string because Elastic emits strings in #1211.

Without this change we get a parse error:
```
parsing time \"{\"timestamp\":\"2019-10-22T15:07:37.519Z\",\"successful\":true}\" as \"\"2006-01-02T15:04:05Z07:00\"\": cannot parse \"{\"timestamp\":\"2019-10-22T15:07:37.519Z\",\"successful\":true}\" as \"\"\"
```

### DIFF
--- a/xpack_watcher_get_watch.go
+++ b/xpack_watcher_get_watch.go
@@ -179,15 +179,21 @@ type XPackWatchExecutionState struct {
 }
 
 type XPackWatchActionStatus struct {
-	AckStatus               *XPackWatchActionAckStatus `json:"ack"`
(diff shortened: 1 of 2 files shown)


olivere:
Thanks!

PR #1219, opened 2019-10-26
Support partitioning in the SignificantTerms aggregation

panychek:
This patch adds the missing methods to the SignificantTerms aggregation.

### DIFF
--- a/search_aggs_bucket_significant_terms.go
+++ b/search_aggs_bucket_significant_terms.go
@@ -111,6 +111,22 @@ func (a *SignificantTermsAggregation) ExcludeValues(values ...interface{}) *Sign
 	return a
 }
 
+func (a *SignificantTermsAggregation) Partition(p int) *SignificantTermsAggregation {
+	if a.includeExclude == nil {
+		a.includeExclude = &TermsAggregationIncludeExclude{}
+	}
+	a.includeExclude.Partition = p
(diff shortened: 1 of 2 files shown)


olivere:
I'd prefer the more generic:

```
func (a *SignificantTermsAggregation) IncludeExclude(includeExclude *TermsAggregationIncludeExclude) *SignificantTermsAggregation {
	a.includeExclude = includeExclude
	return a
}
```

panychek:
The thing is that these methods are present in the [Terms](https://github.com/olivere/elastic/blob/release-branch.v7/search_aggs_bucket_terms.go#L123) and [SignificantText](https://github.com/olivere/elastic/blob/release-branch.v7/search_aggs_bucket_significant_text.go#L129) aggregations.

So we can use them like:
`agg := NewSignificantTextAggregation().Field("field_name").Partition(0).NumPartitions(20)`

or:
`agg := NewTermsAggregation().Field("field_name").Partition(0).NumPartitions(20)`

But this will fail:
`agg := NewSignificantTermsAggregation().Field("field_name").Partition(0).NumPartitions(20)`

Seems inconsistent.

olivere:
Oh my. Yes, then we should add the `Partition` and `NumPartitions` and mark them as deprecated for removal in the future.

Issue #1207, opened 2019-10-16
missing ExtendedStatsAggregation.Sigma()


diacone:
### Which version of Elastic are you using?

[x] elastic.v7 (for Elasticsearch 7.x)
[ ] elastic.v6 (for Elasticsearch 6.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v2 (for Elasticsearch 1.x)

Hi, 
first of all, nice job on this library.

It would be nice to add configuration (Sigma()) for [standard deviation bounds](https://www.elastic.co/guide/en/elasticsearch/reference/7.0/search-aggregations-metrics-extendedstats-aggregation.html#_standard_deviation_bounds)

Thank you.

PR #1220, opened 2019-10-27
#1207 Expose sigma in ExtendedStatsAggregation

kevinconaway:
This PR adds the `sigma` field to `ExtendedStatsAggregation`.

Addresses #1207


olivere:
I prefer to get an error if you use invalid values of `sigma`. It tells you directly that you've done something wrong without hiding it.

Issue #1179, opened 2019-09-09
[PerformRequestWithOptions] - Healthcheck followed by elasticsearch request only if retry is enabled


askterm:
### Which version of Elastic are you using?
[ ] elastic.v5 (for Elasticsearch 5.x)

### Please describe the expected behavior
If all nodes are marked as dead, PerformRequestWithOptions fires another healthcheck; if it finds a healthy node, it should fire the elasticsearch query irrespective of whether retry is configured or not.

### Please describe the actual behavior
If all nodes are marked as dead, PerformRequestWithOptions fires another healthcheck; if now it finds a healthy node, the query will be fired only if retry is configured. Since no request to elasticsearch was made until this point, it should not be considered as a retry. Its just that the healthcheck (which runs every one minute as per the default configuration), needs to be run again because elasticsearch may be alive by this time. Why should the request be considered a retry, when request has not been fired even once?

### Any steps to reproduce the behavior?
Bring down elasticsearch then wait for healthcheck (check logs - it should mark all nodes as dead), then bring up elasticsearch. If a request is fired (using the library) before next healthcheck, and retry is not configured, then request fails. Subsequent requests will succeed because nodes are marked alive by then.

olivere:
I see. Will try.

Any reason why you didn't enable retries? It's good practice anyway.

askterm:
We are planning to add retries now, but it seemed like there were some intermittent elasticsearch connection failures, which could have been resolved without retry, with a query fired after healthcheck was successful. I created an issue to get clarity on whether its an intentional design decision and wanted to know the reason if that was the case.

mscheifer:
We have also run into this issue after turning off retries. We intentionally disabled retries because they just make minor traffic spikes turn into larger outages.

mscheifer:
@olivere is there anything we can do to help get this PR merged in? :point_up:

olivere:
@mscheifer I‘m sincerely sorry that I’m not very responsive lately. My daytime job keeps me severely busy for months now. I will try my best to look into it over Xmas. Again, sorry.

mscheifer:
@olivere No worries! I totally understand that you have other obligations.

olivere:
Will be fixed in 7.0.10, 6.2.27 and 5.0.84.

PR #1235, opened 2019-11-15
Perform request if healthcheck finds a healthy node

kyangtt:
Issue #1179


olivere:
I'm currently fixing this for v5 to v7. Thank you.

PR #1236, opened 2019-11-17
feature: column aliases in cat indices API

larrycinnabar:
#### Why this should be part of the repo?

This repository allows to call any elastic request, so if elastic can execute `http://localhost:9200/_cat/indices?h=idx`, then this repo should too. (having `idx` instead of `index`)

#### Implementation

- in `cat_indices.go`, when url is built - given columns are looked up in the alias map and replaced if any
- an alias can be given to multiple original fields because of backwards compatibility (depending on which of 7.* version, you may have `refresh.time` or `refresh.external_time`, and they both have `rti` alias)
- alias map is constructed on the init stage (using `init()` function). It uses `reflect`, but as it's one-time on-startup call, I consider this not harmful.
Other option (to avoid reflect) is to manually declare the alias map. But this approach has a drawback as well: we declare fields in two places manually: the `CatIndicesResponseRow` struct and the aliases map. And when updating further, developer is required to update BOTH things, so it will become a danger place that can admit bugs when developer updates one thing and forgets about other.
- Tests are added:
  + very simple test to check if alias map exists
  + test on `?h=h` to be the same as `?h=health`
  + test on double alias with `?h=rti`

### DIFF
--- a/cat_indices.go
+++ b/cat_indices.go
@@ -10,7 +10,6 @@ import (
 	"github.com/olivere/elastic/v7/uritemplates"
 	"net/http"
 	"net/url"
-	"reflect"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


larrycinnabar:
@olivere  any ideas, arguments, suggestions?

cat_indices.go
```diff
@@ -240,135 +256,176 @@ type CatIndicesResponse []CatIndicesResponseRow
...
+		for _, alias := range strings.Split(aliasVal, ",") {
+			if existed, ok := catIndicesResponseRowAliasesMap[alias]; ok {
+				catIndicesResponseRowAliasesMap[alias] = existed + "," + jsonVal
+			} else {
+				catIndicesResponseRowAliasesMap[alias] = jsonVal
+			}
+		}
+	}
+
+	return
```
olivere: Although this works, I think it's overkill to use reflection for it. Can't we just use a simple map?

```
var catIndicesResponseRowAliasesMap = map[string]string{
  "i":   "index",
  "idx": "index",
  ...
}
```

Keep it simple and stupid.

olivere (changes requested):
Just a comment to simplify things a bit.

cat_indices.go
larrycinnabar: Yeap, as I described above - I understand that reflect maybe a complication, so I offer two options: reflect and a simple hardcoded map.

The only drawback of a map, as I see, is - more possible case that it can be out-to-date when the main struct is updated (so e.g. a field is added/removed and the alias was not updated, as it is stored in a separate map). 

If you're OK with it too, I'll switch to a single map then

olivere: Yeah. We simply add it then. It's just a simple bug. You could've forgotten the alias as well.

larrycinnabar:
Upd: switched from alias reflect-generation to manual hard-coded map

Ready for merging

olivere:
Thanks. Will be released in 7.0.10.

PR #1237, opened 2019-11-18
ignore FieldSecurity when it is null

Jack0005:
ignore FieldSecurity when it is null

### DIFF
--- a/xpack_security_get_role.go
+++ b/xpack_security_get_role.go
@@ -181,6 +181,6 @@ type XPackSecurityApplicationPrivileges struct {
 type XPackSecurityIndicesPermissions struct {
 	Names         []string    `json:"names"`
 	Privileges    []string    `json:"privileges"`
-	FieldSecurity interface{} `json:"field_security"`
+	FieldSecurity interface{} `json:"field_security,omitempty"`
 	Query         string      `json:"query"`
 }


olivere:
How does this change the semantics? `XPackSecurityIndicesPermissions` is deserialized from the ES response. If ES doesn't return the field, it's `nil` in Go. If ES returns the field as `null`, it is still `nil`. The `omitempty` attribute would make sense when _serializing_ the struct, which we don't do anywhere, or do we?

Jack0005:
when i use XPackSecurityRole  to create ES Role, it return err like this:

```
elastic: Error 400 (Bad Request): failed to parse indices privileges for role [cl2917_deployment_c3]. expected START_OBJECT or START_ARRAY but got VALUE_NULL in "field_security". [type=parse_exception]]
```
json data is 
```
{
	"cluster": [],
	"indices": [{
		"names": ["cloud_container*"],
		"privileges": ["read"],
		"field_security": null,
		"query": ""
	}],
	"applications": [{
		"application": "kibana-.kibana",
		"privileges": ["feature_dev_tools.read", "feature_discover.read", "feature_visualize.read", "feature_dashboard.read", "feature_indexPatterns.all"],
		"resources": ["space:default", "space:ocean"]
	}],
	"run_as": [],
	"global": {},
	"metadata": {},
	"transient_metadata": {
		"enable": true
	}
}
```
when i delete  field_security in XPackSecurityIndicesPermissions and test in kibana，Successfully created a role。

1. there is no XPackSecurityPutRoleServiceRequest in xpack_security_put_role.go. So i chose XPackSecurityRole. when XPackSecurityIndicesPermissions   is serialized, field_security exist and is null.
2.  May be ES’field_security default this can't be empty or doesn‘t  exist. 
3. I think the easiest way is to get rid of it. Or add XPackSecurityPutRoleServiceRequest  in xpack_security_put_role.go.

olivere:
Okay, I see. I will have a detailed look at it in the next release.

PR #1244, opened 2019-11-24
feature(query-interval): QueryIntervals implemented

### LINKED ISSUE #1228: intervals query

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-intervals-query.html

The official website has a new function:intervals query. 
Have you achieved it?

larrycinnabar:
fix #1228 

WIP:
- [ ] automated tests are not there yet
- [ ] not full documentation comments are there yet
- [ ] some doubts on FilterRule (there is a `todo` placed there)
- [x] @olivere are you OK with the approach of declaring a new interface `IntervalRule`, that actually is just a `Query`? It is made just for the reason: so some methods like `NewIntervalsQuery` accept only those 6 types of rules (but not all the queries). `IntervalsQuery` is the first time library meets such thing

### DIFF
--- a/search_queries_interval.go
+++ b/search_queries_interval.go
@@ -1,0 +1,43 @@
+// Copyright 2012-present Oliver Eilhard. All rights reserved.
+// Use of this source code is governed by a MIT-license.
+// See http://olivere.mit-license.org/license.txt for details.
+
+package elastic
+
+// IntervalRule represents the generic matching rule interface. Matching Rule is
+// just a Query, but may be used only inside IntervalQuery. An extra method is added
+// just to separate *Rule objects from other queries
+type IntervalRule interface {
+	Query
+
+	// IsIntervalRule is never actually called, and is used just for Rule to differ from standard Query
+	IsIntervalRule() bool
+}
+
+// IntervalQuery returns documents based on the order and proximity of matching terms
+//
+// For more details, see
+// https://www.elastic.co/guide/en/elasticsearch/reference/7.4/query-dsl-intervals-query.html
+type IntervalQuery struct {
+	field string
+	rule  IntervalRule
+}
+
+// NewIntervalQuery creates and initializes a new IntervalQuery.
+func NewIntervalQuery(field string, rule IntervalRule) *IntervalQuery {
+	return &IntervalQuery{field: field, rule: rule}
+}
+
+// Source returns JSON for the function score query.
+func (q *IntervalQuery) Source() (interface{}, error) {
+	source := make(map[string]interface{})
+
+	ruleSrc, err := q.rule.Source()
+	if err != nil {
+		return nil, err
+	}
+
+	source[q.field] = ruleSrc
+
+	return source, nil
+}
--- a/search_queries_interval.go
+++ b/search_queries_interval.go
@@ -4,14 +4,14 @@
 
 package elastic
 
-// IntervalRule represents the generic matching rule interface. Matching Rule is
+// IntervalQueryRule represents the generic matching interval rule interface. Interval Rule is actually
 // just a Query, but may be used only inside IntervalQuery. An extra method is added
-// just to separate *Rule objects from other queries
-type IntervalRule interface {
+// just to separate all its implementations (*Rule objects) from other query objects
+type IntervalQueryRule interface {
 	Query
 
-	// IsIntervalRule is never actually called, and is used just for Rule to differ from standard Query
-	IsIntervalRule() bool
+	// isIntervalQueryRule is never actually called, and is used just for Rule to differ from standard Query
+	isIntervalQueryRule() bool
 }
 
 // IntervalQuery returns documents based on the order and proximity of matching terms
@@ -20,11 +20,11 @@ type IntervalRule interface {
 // https://www.elastic.co/guide/en/elasticsearch/reference/7.4/query-dsl-intervals-query.html
 type IntervalQuery struct {
 	field string
-	rule  IntervalRule
+	rule  IntervalQueryRule
 }
 
 // NewIntervalQuery creates and initializes a new IntervalQuery.
-func NewIntervalQuery(field string, rule IntervalRule) *IntervalQuery {
+func NewIntervalQuery(field string, rule IntervalQueryRule) *IntervalQuery {
 	return &IntervalQuery{field: field, rule: rule}
 }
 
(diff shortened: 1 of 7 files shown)


olivere:
I typically design queries after the Java classes. For InternalsQuery, it looks like this probably: https://github.com/elastic/elasticsearch/blob/v7.4.2/server/src/main/java/org/elasticsearch/index/query/IntervalQueryBuilder.java

Will look into your code tomorrow.

larrycinnabar:
Actually, it's pretty close. Naming is different. They have `SourceProvider` and I called it as in elastic documentation: `MatchingRule`. And I splitted every matching rule in its own file (can be merged to one file if needed)

And, do you handle the version differences inside? for example from 7.0 to 7.4 the IntervalsQuery differs (more matching rule, more fields, etc)

olivere:
I’d use the most complete version you can find. If features were simply added, it’s up to the consumer of the API to only use those features available for the Elasticsearch server at hand.

larrycinnabar:
> I’d use the most complete version you can find. If features were simply added, it’s up to the consumer of the API to only use those features available for the Elasticsearch server at hand.

Ok, makes sense. In this branch it's done by the latest 7.4 documentation, and yeap - the new features are simply added, so no conflicts.

search_queries_interval.go
```diff
@@ -0,0 +1,43 @@
...
+package elastic
+
+// IntervalRule represents the generic matching rule interface. Matching Rule is
+// just a Query, but may be used only inside IntervalQuery. An extra method is added
+// just to separate *Rule objects from other queries
+type IntervalRule interface {
+	Query
+
+	// IsIntervalRule is never actually called, and is used just for Rule to differ from standard Query
+	IsIntervalRule() bool
```
olivere: If it's only used internally, you could mark it as unexported: `isIntervalRule() bool`?

search_queries_interval.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2012-present Oliver Eilhard. All rights reserved.
+// Use of this source code is governed by a MIT-license.
+// See http://olivere.mit-license.org/license.txt for details.
+
+package elastic
+
+// IntervalRule represents the generic matching rule interface. Matching Rule is
+// just a Query, but may be used only inside IntervalQuery. An extra method is added
+// just to separate *Rule objects from other queries
+type IntervalRule interface {
```
olivere: Can we maybe rename that to `IntervalQueryRule`, so that they share a common suffix? Will probably help with intellisense in such a huge package that is `elastic.Interval___`.

search_queries_interval.go
```diff
@@ -0,0 +1,43 @@
...
+//
+// For more details, see
+// https://www.elastic.co/guide/en/elasticsearch/reference/7.4/query-dsl-intervals-query.html
+type IntervalQuery struct {
+	field string
+	rule  IntervalRule
+}
+
+// NewIntervalQuery creates and initializes a new IntervalQuery.
+func NewIntervalQuery(field string, rule IntervalRule) *IntervalQuery {
```
olivere: There can be only one `IntervalQueryRule`, correct?

search_queries_interval_rules_all_of.go
```diff
@@ -0,0 +1,66 @@
+package elastic
+
+type AllOfRule struct {
```
olivere: That should better be `IntervalQueryRuleAllOf` or `IntervalQueryAllOf`.

search_queries_interval_rules_any_of.go
```diff
@@ -0,0 +1,47 @@
+package elastic
+
+type AnyOfRule struct {
```
olivere: Same here: I think it'd be better to use `IntervalQueryAnyOf`.

search_queries_interval_rules_filter.go
```diff
@@ -0,0 +1,142 @@
+package elastic
+
+type FilterRule struct {
```
olivere: It's too broad a term. We now have hundreds of types in `elastic` package already, so it'd be better to at least use a prefix to know where it belongs.

search_queries_interval_rules_filter.go
```diff
@@ -0,0 +1,142 @@
...
+	if r.script != nil {
+		src, err := r.script.Source()
+		if err != nil {
+			return nil, err
+		}
+		source["script"] = src
+	}
+
+	// todo: not so clear from docs, if filter can be a top-level rule, and so
+	// 	if it does need a wrapper map[string]interface{}{"filter": ..} like other rules do
```
olivere, line 135: Always use the Java source as the ultimate reference. The online docs are pretty good, but targeted to a broader understanding of the concepts, not hard facts.

olivere (changes requested):
Looks good overall. I like it.

We probably only have to change type names accordingly and double-check that we're complete, plus: Setup some tests that the query creates the correct JSON. I typically use the examples from the online documentation for that.

search_queries_interval.go
larrycinnabar: maybe `IntervalRuleQuery` then?

search_queries_interval.go
larrycinnabar: Do you mean if `rule ..IntervalRule` should be possible?
I guess no, but I'll check in java sources too.

The idea is that you add just one rule object. If you need complex logic: you use `all_of`, `any_of`, etc then can add OR/ADD logic to your multiple rules

search_queries_interval.go
olivere: I'd go for `IntervalQueryRule` so that everything "interval query"-related has the same prefix: `IntervalQuery`.

search_queries_interval.go
olivere: Yes, that's what I meant. Maybe you can do `rule ...IntervalQueryRule`... I haven't looked into the Java source yet.

larrycinnabar:
issue status upd: having busy week, will handle on weekend

olivere:
Same here. Take your time. I'm probably even busy over the weekend.

search_queries_interval.go
larrycinnabar: in Java sources it is called `IntervalsSourceProvider` and it can be just one. So nothing is changed here.

olivere:
I've reviewed the PR, added tests and documentation, and fixed a few things. It will be released in 7.0.10.

PR #1256, opened 2019-12-18
(fix) DateHistogramAggregation.CalendarInterval: update documentation…

larrycinnabar:
CalendarInterval doesn't allow multiple units (not obvious from the first sight, so need to be added in the comment). Also, units less than day are not allowed too

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html


olivere:
Thanks.

FYI I've changed the link to use an exact version (`7.5`) instead of `current`. The `current` will lead users to the 8.x docs in a few weeks...

PR #1279, opened 2020-02-04
Update link to doc

rbUUbr:
due to: https://blog.golang.org/go.dev
I've updated link to the documentation page to make it more actual.
I haven't found valid badge for it, so leave it as it is :) Hope it will help this project


olivere:
It's https://pkg.go.dev/github.com/olivere/elastic/v7 actually, but I already fixed.

PR #1289, opened 2020-02-16
Add clear cache api

amanjain97:
This PR is regarding adding `_cache/clear` API to elastic client. 
Doc: https://www.elastic.co/guide/en/elasticsearch/reference/6.8/indices-clearcache.html

Please suggest if anything else is required.

### DIFF
--- a/indices_clear_cache.go
+++ b/indices_clear_cache.go
@@ -1,0 +1,227 @@
+package elastic
+
+import (
+	"context"
+	"fmt"
+	"github.com/olivere/elastic/uritemplates"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+type IndicesClearCacheService struct {
+	client *Client
+
+	pretty     *bool       // pretty format the returned JSON response
+	human      *bool       // return human readable values for statistics
+	errorTrace *bool       // include the stack trace of returned errors
+	filterPath []string    // list of filters used to reduce the response
+	headers    http.Header // custom request-level HTTP headers
+
+	index             []string
+	ignoreUnavailable *bool
+	allowNoIndices    *bool
+	expandWildcards   string
+	fieldData         *bool
+	fields            string
+	query             *bool
+	request           *bool
+}
+
+func NewIndicesClearCacheService(client *Client) *IndicesClearCacheService {
+	return &IndicesClearCacheService{client: client}
+}
+
+// Pretty tells Elasticsearch whether to return a formatted JSON response.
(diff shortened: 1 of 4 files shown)


indices_clear_cache.go
```diff
@@ -0,0 +1,227 @@
+package elastic
```
olivere, line 5: The copyright notice is missing here.

olivere (changes requested):
The copyright notice is missing in the file headers, but the code looks good.

Please double-check that the service has all properties defined in the [REST API spec](https://github.com/elastic/elasticsearch/blob/6.7/rest-api-spec/src/main/resources/rest-api-spec/api/indices.clear_cache.json) for v6. It looks complete, but I haven't checked in depth.

Thanks!

olivere:
I will also need to port it to v7 as well.

indices_clear_cache.go
amanjain97: @olivere Sorry, I am not much familiar with adding copyrights. I have added my name. Please check.

amanjain97:
Can you take a look at https://github.com/elastic/elasticsearch/blob/6.7/rest-api-spec/src/main/resources/rest-api-spec/api/indices.clear_cache.json#L49

I added `request_cache` field which seems duplicated with `request`.

indices_clear_cache.go
olivere: No. Simply copy the header from any of the other files.

amanjain97: Sure ok.

olivere:
> Can you take a look at https://github.com/elastic/elasticsearch/blob/6.7/rest-api-spec/src/main/resources/rest-api-spec/api/indices.clear_cache.json#L49
> 
> I added `request_cache` field which seems duplicated with `request`.

Oh, one of them seems to be a deprecated field. I typically look up in the Java code which of the fields is used. The high-level Java client [seems to use the `request` field](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/rest/action/admin/indices/RestClearIndicesCacheAction.java#L65).

amanjain97:
@olivere I shall add comment on the function as deprecated. Maybe we can remove in v7

olivere:
Will be part of the next v6 and v7 release. Thank you.

amanjain97:
Sorry, why did you close when you write `Will be part of the next v6 and v7 release. Thank you.`

olivere:
Because it's already part of the release branches, but not tagged yet.

Issue #1306, opened 2020-04-08
AWS Auth Signing doesn't work for ElasticSearch record Updates


veqryn:
### Which version of Elastic are you using?

[x] elastic.v7 (for Elasticsearch 7.x)
[x] elastic.v6 (for Elasticsearch 6.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v2 (for Elasticsearch 1.x)

### Please describe the expected behavior
I am running an update using this library, against elasticsearch service in aws.
All my writing and reading of records works, but the update doesn't.

```go
	// Example:
	updateScript := elastic.NewScriptInline("ctx._source.platform = params.platform").
		Lang("painless").
		Param("platform", params.Value)

	query := pixelFilterParamsQuery(start, end, params.Filter)
	query = query.Filter(elastic.NewExistsQuery(params.Field))

	update := es.client.UpdateByQuery().
		Index(indices).
		Type(typePixel).
		IgnoreUnavailable(true).
		WaitForCompletion(true).
		Refresh("true").
		Slices("auto").
		AbortOnVersionConflict().
		Query(query ).
		Script(updateScript)

	response, err := update.Do(ctx)
```

I have tracing logging turned on with the olivere/elastic client, and I get the following in my logs:

```
POST /ecm-2020-03-29%2Cecm-2020-03-30%2Cecm-2020-03-31/pixel/_update_by_query?conflicts=abort&ignore_unavailable=true&refresh=true&slices=auto&wait_for_completion=true HTTP/1.1
Host: xxxxxxxxx.us-east-1.es.amazonaws.com
User-Agent: elastic/6.2.29 (linux-amd64)
Transfer-Encoding: chunked
Accept: application/json
Content-Type: application/json
Accept-Encoding: gzip
fe
{\"query\":{\"bool\":{\"filter\":[{\"range\":{\"ts\":{\"from\":1585526400,\"include_lower\":true,\"include_upper\":false,\"to\":1585527300}}},{\"exists\":{\"field\":\"plat\"}}]}},\"script\":{\"lang\":\"painless\",\"params\":{\"plat\":\"desktop\"},\"source\":\"ctx._source.plat = params.plat\"}}
0
```

```
HTTP/2.0 403 Forbidden
Content-Length: 192
Access-Control-Allow-Origin: *
Content-Type: application/json
Date: Wed, 08 Apr 2020 22:29:18 GMT
[...]

PR #1307, opened 2020-04-09
Fix aws v4 signer to not read body if already seeker, not read response

veqryn:
Related to comments in: https://github.com/olivere/elastic/issues/1306
1. Do not read the request body if it is already a ReaderSeeker
2. Do not read the whole response, let the caller do that.

---
This change is [Reviewable](https://reviewable.io/reviews/olivere/elastic/1307)

### DIFF
--- a/aws/v4/aws_v4.go
+++ b/aws/v4/aws_v4.go
@@ -6,6 +6,7 @@ package v4
 
 import (
 	"bytes"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


aws/v4/aws_v4.go
```diff
@@ -53,37 +54,29 @@ func (st Transport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		switch body := req.Body.(type) {
+		case io.ReadSeeker:
+			_, err = st.signer.Sign(req, body, "es", st.region, now)
+		default:
+			buf, err := ioutil.ReadAll(req.Body)
+			if err != nil {
+				return nil, err
+			}
+			req.Body = ioutil.NopCloser(bytes.NewReader(buf))
+			_, err = st.signer.Sign(req, bytes.NewReader(buf), "es", st.region, time.Now().UTC())
```
olivere, line 75: To be honest, I don't think all of this is necessary at all. If I read [the documentation](https://github.com/aws/aws-sdk-go/blob/v1.30.7/aws/signer/v4/v4.go#L251-L275) correctly, the body should be automatically [wrapped in an `io.ReadCloser`](https://github.com/aws/aws-sdk-go/blob/v1.30.7/aws/signer/v4/v4.go#L261-L265):

```go
// Sign will set the request's Body to be the `body` parameter passed in. If
// the body is not already an io.ReadCloser, it will be wrapped within one. If
// a `nil` body parameter passed to Sign, the request's Body field will be
// also set to nil. Its important to note that this functionality will not
// change the request's ContentLength of the request.
```

Also, by default, the request body should be overwritten by the `Sign` method by default (as long as [`DisableRequestBodyOverwrite`](https://github.com/aws/aws-sdk-go/blob/v1.30.7/aws/signer/v4/v4.go#L190-L198) is `false`, which is the default):

```go
// Disables the automatical setting of the HTTP request's Body field with the
// io.ReadSeeker passed in to the signer. This is useful if you're using a
// custom wrapper around the body for the io.ReadSeeker and want to preserve
// the Body value on the Request.Body.
//
// This does run the risk of signing a request with a body that will not be
// sent in the request. Need to ensure that the underlying data of the Body
// values are the same.
DisableRequestBodyOverwrite bool
```

The code snippet that overrides the body is [here](https://github.com/aws/aws-sdk-go/blob/v1.30.7/aws/signer/v4/v4.go#L354-L366).

Do you have a test cluster on AWS where you could actually test this? I think it would help a lot in terms of removing allocation overhead if we could skip reading the request body.

olivere (changes requested):
Maybe we can optimize a bit more.

Did you test the changes in an actual test environment on AWS, or is this a dry run?

aws/v4/aws_v4.go
veqryn: So, yes, I did just test this against an actual AWS ES cluster, and it does work.

I believe you are right, that we do not need to set the req.Body (line 74).
However, we do still need to read in the entire req.Body, and then wrap it in a bytes.Reader, because the AWS signer requires a io.ReadSeeker
The only way to get around that, is to have the main olivere/elastic library use ReadSeekers whenever possible (ie: bytes.Reader instead of bytes.Buffer).

olivere: I can't assess the impact of changing `bytes.Reader` to `bytes.Buffer`, so I don't want to do this in the scope of this PR.

So, OK, if we can remove line 74 (wrapping with a nop closer) and the integration test with an actual AWS ES cluster worked, I'm happy to squash and merge. What do you think?

veqryn: I take that back, we can't remove line 74, because if DisableRequestBodyOverwrite is set to true, we will end up sending an empty request body (because it was already read).

So I think this PR is as good as it will get, without touching the main olivere/elastic files.

olivere: Ok, well... I will merge when I'm back at my computer.

olivere (approved):
LGTM 🚀

PR #1308, opened 2020-04-09
Fix aws v4 signer to not read body if already seeker, not read response

veqryn:
Related to comments in: https://github.com/olivere/elastic/issues/1306
1. Do not read the request body if it is already a ReaderSeeker
2. Do not read the whole response, let the caller do that.

---
This change is [Reviewable](https://reviewable.io/reviews/olivere/elastic/1308)


olivere (approved):
LGTM 🚀

PR #1302, opened 2020-03-31
add error field to response to tasks_get_tasks

huyhvq:
add error field to response to tasks_get_tasks

Closes #1301

Issue #1301, opened 2020-03-31
missing error field in struct TasksGetTaskResponse


huyhvq:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

[x] elastic.v7 (for Elasticsearch 7.x)
[ ] elastic.v6 (for Elasticsearch 6.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v2 (for Elasticsearch 1.x)
go mod github.com/olivere/elastic/v7 v7.0.6
elasticsearch version:7.3.0

### Please describe the expected behavior
API get task by ID `_tasks/{id}` must return error field.
Ex:
```json
{
  "completed" : true,
  "task" : {
    "node" : "MO-PoeuPSX66JG1Vp02e8A",
    "id" : 128023221,
    "type" : "transport",
    "action" : "indices:data/write/update/byquery",
    "status" : {
      "total" : 154435,
      "updated" : 0,
      "created" : 0,
      "deleted" : 0,
      "batches" : 1,
      "version_conflicts" : 0,
      "noops" : 0,
      "retries" : {
        "bulk" : 0,
        "search" : 0
      },
      "throttled_millis" : 0,
      "requests_per_second" : -1.0,
      "throttled_until_millis" : 0
    },
    "description" : "update-by-query [ahjhjdongok_alias] updated with Script{type=inline, lang='painless', idOrCode='asddds', options={}, params={ahjhj='do ngok'}}",
    "start_time_in_millis" : 1585629334222,
    "running_time_in_nanos" : 11128042,
    "cancellable" : true,
    "headers" : { }
  },
  "error" : {
    "type" : "script_exception",
    "reason" : "compile error",
    "script_stack" : [
      "asddds",
      "^---- HERE"
    ],
    "script" : "asddds",
    "lang" : "painless",
    "caused_by" : {
      "type" : "illegal_argument_exception",
      "reason" : "Variable [asddds] is not defined."
    }
  }
}
```

### Please describe the actual behavior
Missing error field on _tasks api response `TasksGetTaskResponse`
https://github.com/olivere/elastic/blob/0e429faaa934317f89b23fd6005f91364fb5c1a0/tasks_get_task.go#L164-L168
### Any steps to reproduce the behavior?
POST wrong script using `_update_by_query` API
```
[...]

PR #1309, opened 2020-04-09
Add fields for script exceptions to error details

olivere:
When a task fails when run asynchronously, it returns the error in the
Task Get API. If it was a script exception, it returns four additional
fields that we added now.

Close #1302


olivere:
Closes #1301

PR #1313, opened 2020-04-15
Fix SearchResult struct so it is compatible with cross-cluster search results

tdinucci:
The `Clusters` member of the `SearchResult` struct is currently an array and I don't believe it should be.  This causes the following error when results of a cross-cluster search are unmarshalled:

```
cannot unmarshal object into Go struct field SearchResult._clusters of type []*elastic.SearchResultCluster"
```
The Elasticsearch documentation shows that the `_clusters` element is indeed not an array:
https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-cross-cluster-search.html#ccs-search-remote-cluster

This PR resolves this issue.

### DIFF
--- a/search.go
+++ b/search.go
@@ -641,20 +641,20 @@ func (s *SearchService) Do(ctx context.Context) (*SearchResult, error) {
 
 // SearchResult is the result of a search in Elasticsearch.
 type SearchResult struct {
-	Header          http.Header            `json:"-"`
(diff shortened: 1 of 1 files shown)


olivere:
Already fixed. See https://github.com/olivere/elastic/blob/release-branch.v7/search.go#L648.

PR #1314, opened 2020-04-16
fix comment

langleegelang:


olivere:
Thanks. Fixed in both v6 and v7.

PR #1326, opened 2020-05-08
No need to force https in aws signer

### LINKED ISSUE #1328: Do not force https on AWS signer
We do not need to force HTTPS when doing the AWS signing. This makes working with the library harder in development.

veqryn:
There is no need to force https in the aws signer.
This already happens by default, and forcing it ruins the ability to test in dev against an Elasticsearch docker container.

Close #1328


olivere (approved):
LGTM

PR #1338, opened 2020-06-02
span near query was added

pzentenoe:
I added span near query for need of use. you could check if it is feasible to upload this and generate a new olivere release

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ Here are a few tips on how to get used to Elastic:
   - [ ] Span Term Query
   - [ ] Span Multi Term Query
   - [ ] Span First Query
-  - [ ] Span Near Query
+  - [x] Span Near Query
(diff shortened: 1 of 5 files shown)


olivere:
It goes into the right direction. I always use the Java source as the ultimate source of truth. For `SpanNearQuery`, that's [this file](https://github.com/elastic/elasticsearch/blob/v7.7.0/server/src/main/java/org/elasticsearch/index/query/SpanNearQueryBuilder.java).

pzentenoe:
modify the name of the files based on what you sent me and create the builder methods

pzentenoe:
could you check again? please

olivere:
Thanks for your work. I will add a few more Span queries in a different branch with 7.0.17 and 6.2.33.

PR #1340, opened 2020-06-03
BulkProcessor flusher goroutine panic recover avoid process die

Jeff885:
image


olivere:
Sorry, but your PR is just hiding the root cause of _why_ the panic occurred in the first place. Also, in the screenshot, you're using v5, in the PR you're using v7. Can you instead file an issue with a reproducible test that causes the problem?

PR #1345, opened 2020-06-07
Fix XpackWatch struct

wellshs:
I found `XPackWatch` struct's Action fields have wrong type specification.

XPackWatchActionStatus is representing Status corresponding action, which represent status of watch actions. However, [watch get api](https://www.elastic.co/guide/en/elasticsearch/reference/current/watcher-api-get-watch.html) returns specification of actions in `XpackWatch.Actions` position.

### DIFF
--- a/xpack_watcher_get_watch.go
+++ b/xpack_watcher_get_watch.go
@@ -202,13 +202,13 @@ type XPackWatchActionThrottle struct {
 }
 
 type XPackWatch struct {
-	Trigger                map[string]map[string]interface{}  `json:"trigger"`
-	Input                  map[string]map[string]interface{}  `json:"input"`
-	Condition              map[string]map[string]interface{}  `json:"condition"`
-	Transform              map[string]interface{}             `json:"transform,omitempty"`
-	ThrottlePeriod         string                             `json:"throttle_period,omitempty"`
-	ThrottlePeriodInMillis int64                              `json:"throttle_period_in_millis,omitempty"`
-	Actions                map[string]*XPackWatchActionStatus `json:"actions"`
-	Metadata               map[string]interface{}             `json:"metadata,omitempty"`
-	Status                 *XPackWatchStatus                  `json:"status,omitempty"`
+	Trigger                map[string]map[string]interface{} `json:"trigger"`
+	Input                  map[string]map[string]interface{} `json:"input"`
+	Condition              map[string]map[string]interface{} `json:"condition"`
+	Transform              map[string]interface{}            `json:"transform,omitempty"`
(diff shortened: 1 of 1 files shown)


phillbaker:
I also ran into this recently. @olivere would you mind taking a look?

xpack_watcher_get_watch.go
```diff
@@ -202,13 +202,13 @@ type XPackWatchActionThrottle struct {
...
-	Actions                map[string]*XPackWatchActionStatus `json:"actions"`
-	Metadata               map[string]interface{}             `json:"metadata,omitempty"`
-	Status                 *XPackWatchStatus                  `json:"status,omitempty"`
+	Trigger                map[string]map[string]interface{} `json:"trigger"`
+	Input                  map[string]map[string]interface{} `json:"input"`
+	Condition              map[string]map[string]interface{} `json:"condition"`
+	Transform              map[string]interface{}            `json:"transform,omitempty"`
+	ThrottlePeriod         string                            `json:"throttle_period,omitempty"`
+	ThrottlePeriodInMillis int64                             `json:"throttle_period_in_millis,omitempty"`
+	Actions                map[string]interface{}            `json:"actions"`
```
phillbaker, line 211: To follow the trigger/input/condition pattern above, this could also be `map[string]map[string]interface{}` I believe.

wellshs: I agree, `map[string]map[string]interface{}` looks fine for Actions' type.
However, I can't find strict type specification of watch api in [this api documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/actions.html). 
If you find strict type specification, please share me!

phillbaker:
@olivere any thoughts here?

olivere:
> @olivere any thoughts here?

I'm a bit cautious with something like `map[string]map[string]interface{}` because although okay, it's very cumbersome to use. The basic benefit of this library is to provide a Go-like DSL, and I'd really like to make it at least a bit easier than such a generic access pattern.

I didn't have the time to look into XPack APIs too much. But most of it is available just like any other part of ES, and it seems to use the same logic to create request/response structures. E.g. [here is the code](https://github.com/elastic/elasticsearch/blob/7.8/x-pack/plugin/watcher/src/main/java/org/elasticsearch/xpack/watcher/actions/slack/SlackAction.java#L58-L69) for the Slack action. It's just a bit of work to translate all of this into Go structs, and I didn't have the time for it recently.

phillbaker:
@olivere what do you think of merging this PR to fix the bug with the wrong type being used?

olivere:
Sorry for the delay. I've found a different issue with the API, added a test, and moved it over to #1362. It will be fixed in 7.0.18, hopefully due later today.

Issue #1357, opened 2020-07-08
Incoherent error if basic auth is not provided


adrianlzt:
### Which version of Elastic are you using?

[x] elastic.v7 (for Elasticsearch 7.x)

### Please describe the expected behavior
In case of a elasticsearch server with basic auth activated, if we are not providing it, the error should notify this problem.

### Please describe the actual behavior
Currently the error is:
```
health check timeout: no Elasticsearch node available
```

### Any steps to reproduce the behavior?
I am using [skydive](http://skydive.network/) with elasticsearch as the backend.

The elasticsearch server has TLS and authorization via HTTP basic auth.

If I don't configure the basic auth, the error reported id:
```
2020-07-08T10:01:49.846+0200    ERROR   elasticsearch/client.go:426 (*Client).Start     archer: Elasticsearch not available: health check timeout: no Elasticsearch node available
```

That error is generated when calling ``elastic.NewClient()``

The HEAD call to elasticsearch return a ``401 Unauthorized``

olivere:
Thanks for reporting. Could it be that the actual error is not because HTTP Basic Auth failed (it's actually [in the code](https://github.com/olivere/elastic/blob/81e9b9e88dcb5e23cebadf7262130f9c60b0b175/client.go#L1162-L1164)), but due to a context timeout when checking health at startup? That would mean the error is correct in that the timeout came before the 401.

adrianlzt:
Using http and sniffing traffic with wireshask:

Skydive (elasticsearch client) logs:
```
2020-07-09T10:24:06.826+0200    INFO    client/election.go:105 (*MasterElector).start   archer: starting as the master for /master-analyzer-es-graph-flush: archer
2020-07-09T10:24:11.839+0200    ERROR   elasticsearch/client.go:448 (*Client).Start     archer: Elasticsearch not available: creating elasticsearch client: health check timeout: no Elasticsearch node available
2020-07-09T10:24:17.855+0200    ERROR   elasticsearch/client.go:448 (*Client).Start     archer: Elasticsearch not available: creating elasticsearch client: health check timeout: no Elasticsearch node available
```

Connections seen by wireshark:
![image](https://user-images.githubusercontent.com/3237784/87016009-6f1fcc00-c1ce-11ea-8dcc-de52fe7264d2.png)

So, looks like there is no timeout.

olivere:
[...]

PR #1367, opened 2020-07-13
Handle unauthorized messages from elasticsearch

adrianlzt:
If elasticsearch has configured basic auth, HEAD call will return a HTTP
401 from elasticsearch.
This message should generate an specific error to the user.

Also, non expected messages now generates an error message to the user.


adrianlzt:
Fixes #1357

olivere:
Thanks so much. Closing in favour of 42d2a9d.

PR #1377, opened 2020-07-31
adding pinned query

dungnx:
https://www.elastic.co/guide/en/elasticsearch/reference/7.8/query-dsl-pinned-query.html

### DIFF
--- a/search_queries_pinned.go
+++ b/search_queries_pinned.go
@@ -1,0 +1,60 @@
+// Copyright 2012-present Oliver Eilhard. All rights reserved.
+// Use of this source code is governed by a MIT-license.
+// See http://olivere.mit-license.org/license.txt for details.
+
+package elastic
+
+// PinnedQuery is a query that promotes selected documents to rank higher than those matching a given query.
+//
+// For more details, see:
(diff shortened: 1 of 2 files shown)


olivere:
Wonderful. Thank you. I will merge this in the next release, maybe already over the weekend. Matches exactly the [Java code](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/index/query/PinnedQueryBuilder.java).

Good job!

ivila:
@olivere When use pinned query, sometimes score will be "Infinity" (wrapped it in function score query), but with definition of *float64 in SearchHit.Score(search.go), there will be an UnmashalError.
` json: cannot unmarshal string into Go struct field SearchHit._score of type float64`
I wanna have a pull request but I think this change will break some of the users' codes (maybe they export the score data and define it as float64 too), so I think I should just report this, and leave it to your decision.

I think it is a common problem of many query type, Java support NaN and Infinity, and the server side of elasticsearch will sometimes return this 2 cases, for example, wrong math calculation for NaN, geo for Infinity, and this will result in UnmarshalError of the golang Client SearchHit type.

olivere:
Thanks @ivila for reminding me of this. Will have to write a test to illustrate and fix this problem.

PR #1380, opened 2020-08-17
Update errors.go

ramky1982:
added support for reporting actual error for search/write failure

Error if search request contains sorting on text column without keyword:
request:
{
 "query": {"match_all": {}},
 "sort": [
    {
      "comment": {
        "order": "asc"
      }
    }
  ]
}

ES error:
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "Fielddata is disabled on text fields by default. Set fielddata=true on [comment] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead."
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "tim",
        "node": "YW-tkMpZRfeh_DfeZTFAdg",
        "reason": {
          "type": "illegal_argument_exception",
          "reason": "Fielddata is disabled on text fields by default. Set fielddata=true on [comment] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead."
        }
      }
    ],
    "caused_by": {
      "type": "illegal_argument_exception",
      "reason": "Fielddata is disabled on text fields by default. Set fielddata=true on [comment] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead.",
      "caused_by": {
        "type": "illegal_argument_exception",
        "reason": "Fielddata is disabled on text fields by default. Set fielddata=true on [comment] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead."
      }
    }
  },
  "status": 400
}

Error reported by library:
search_phase_execution_exception : elastic: Error 400 (Bad Request): all shards failed [type=search_phase_execution_exception]

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -61,7 +61,7 @@ func createResponseError(res *http.Response) error {
 		if errReply.Details != nil && len(errReply.Details.CausedBy) > 0 {
 			reason := errReply.Details.CausedBy["reason"]
 			if reason != nil {
-				return fmt.Errorf(fmt.Sprintf("Error 400 (Bad Request): %v %v", errReply.Details.Type, reason))
+				return fmt.Errorf(fmt.Sprintf("Error %v: %v %v", res.StatusCode, errReply.Details.Type, reason))
 			}
 		}
 		return errReply
--- a/errors.go
+++ b/errors.go
@@ -58,6 +58,12 @@ func createResponseError(res *http.Response) error {
 		if errReply.Status == 0 {
 			errReply.Status = res.StatusCode
 		}
+		if errReply.Details != nil && len(errReply.Details.CausedBy) > 0 {
+			reason := errReply.Details.CausedBy["reason"]
+			if reason != nil {
+				return fmt.Errorf(fmt.Sprintf("%v : %v", errReply.Details.Type, reason))
+			}
+		}
 		return errReply
 	}
 	return &Error{Status: res.StatusCode}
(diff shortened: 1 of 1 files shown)


olivere:
I won't merge this PR because it breaks Elastic in two specific ways:

1. You return a generic error by doing `fmt.Errorf(...)`. By doing so, you hide the fact that the library will try to return an `*elastic.Error` whenever possible. There are other error checks like [`elastic.IsStatusCode`](https://github.com/olivere/elastic/blob/release-branch.v7/errors.go#L172-L187) and friends that no longer work because of the `fmt.Errorf`.
2. I like to keep errors short, as does the standard library in general. However, it allows to extract further details about the error, as does this library. To get all details about the error, you can do a type check on `*elastic.Error`: `if elasticErr, ok := err.(*elastic.Error); ok { ... }`. Then you can freely access all details returned by Elasticsearch.

There are also some minor issues: E.g. you use a combination of `fmt.Errorf(fmt.Sprintf(...))`. The inner `fmt.Sprintf` is not required.

It is debatable if the error reason should be added when available (where do you draw the line with when [so much detail](https://github.com/olivere/elastic/blob/release-branch.v7/errors.go#L75-L91) is available). There are two ways to do that: Either we extend the [`*elastic.Error.Error()`](https://github.com/olivere/elastic/blob/release-branch.v7/errors.go#L102-L108) method, or we add a function like `elastic.ErrorReason(err error) string` to make things easier for consumers.

PR #1384, opened 2020-08-26
Add IgnoreThrottled to count and scroll

quixoten:


olivere:
That's looking good. Thank you.

PR #1390, opened 2020-09-14
set DefaultSnifferEnabled false

kotaroooo0:
These days, many people use docker, so we use Elasticsearch on Docker.
In this Elasticsearch client, many people encount error `no active connection found`.
I am also one of them and wasted a few hours trying to resolve.
I want to change Default value!

### DIFF
--- a/client.go
+++ b/client.go
@@ -54,7 +54,7 @@ const (
 	DefaultHealthcheckInterval = 60 * time.Second
 
 	// DefaultSnifferEnabled specifies if the sniffer is enabled by default.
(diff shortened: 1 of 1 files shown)


kotaroooo0:
@olivere How about this idea? If ok, I fix unit test to pass.

olivere:
We had the same discussion before v7, and decided against it. The main reason was that while it might be helpful for new users, it would change the defaults for all existing users--in the worst possible way.

If we'd change it, we can only do so with a major version change.

PR #1434, opened 2020-11-18
use unsafeBytesToString to reuse memory

yjhmelody:
I found that some operations are very expensive about mem alloc when string(bytes).
Such as: 

image

So I try to reuse some mems by using unsafe.


mitar:
Just for the record, this is done inside standard library as well: https://go.dev/src/strings/builder.go#L46

mitar:
Although there will be soon official public API to do this: https://github.com/golang/go/issues/53003

PR #1437, opened 2020-11-19
Add the RestTotalHitsAsInt parameter to the scroll API

### LINKED ISSUE #1436: [v6.2.35] Allow RestTotalHitsAsInt on Scroll API

[x] elastic.v6 (for Elasticsearch 6.x)

svc := client.Scroll(testIndexName).Size(1).RestTotalHitsAsInt(true) should work 

Current Scroll service fails in ES 7 with the following error 

```
cannot unmarshal object into Go struct field SearchHits.hits.total of type int64
```

behos:
This backports the change from https://github.com/olivere/elastic/commit/40e4e98362590a34c09b61ad82ffd57bcb7aec68 to add the RestTotalHitsAsInt to the scroll API.

RestTotalHitsAsInt is required to do online upgrades from ESv6 to ESv7.

example_test.go doesn't compile on the release.v6 branch anymore now that v7 is the default so I removed it to be able to run tests. Let me know if you would prefer that we add it back.

Fixes #1436


olivere:
Thanks.

I will have to revert the removal of `example_test.go`. If they don't work any more, then we should fix them instead of removing tests.

olivere:
Fixed in bfa6e786e7d3613a20c55c2ebeb099e8b04be8b0.

PR #1441, opened 2020-12-02
Request with Content-Length instead of Transfer-Encoding chunked

olivier4576:
Request with Content-Length instead of Transfer-Encoding chunked

### DIFF
--- a/request.go
+++ b/request.go
@@ -113,7 +113,7 @@ func (r *Request) setBodyReader(body io.Reader) error {
 		switch v := body.(type) {
 		case *strings.Reader:
 			r.ContentLength = int64(v.Len())
-		case *bytes.Buffer:
+		case *bytes.Reader:
 			r.ContentLength = int64(v.Len())
 		}
 	}


chrisharrisonkiwi:
Oh my gosh, I'm so glad this is coming.
So many headaches over Docker when it tries to hit an ElasticSearch instance using TransferEncoding: Chunked.

olivere:
@chrisharrisonkiwi I'm always interested in the kinds of problems people face out there: What kind of problem do you see with chunked encoding?

chrisharrisonkiwi:
> @chrisharrisonkiwi I'm always interested in the kinds of problems people face out there: What kind of problem do you see with chunked encoding?

Mostly local development issues using ElasticSearch with "localstack".
Detail: I've been using Wireshark to capture requests from within docker containers made by this library and noticed the transfer encoding approach being used.
If I replay the request from within Docker and replace the chunked encoding for a Content-Length it seems to have a lot more luck networking between local services for some reason.

Attempting: "Transfer-Encoding: chunked" in a localstack Docker environment results in:
`<!doctype html>`
`<title>408 Request Timeout</title>`
`<h1>Request Timeout</h1>`
`Request timed out; try again later`

It could be unrelated, but it certainly seems to help.

olivier4576:
Resolves issue https://github.com/olivere/elastic/issues/1526

Issue #1438, opened 2020-11-25
Unable to set User-Agent with `client.SetHeaders`


flowchartsman:
### Which version of Elastic are you using?

[X] elastic.v7 (for Elasticsearch 7.x)
[ ] elastic.v6 (for Elasticsearch 6.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v2 (for Elasticsearch 1.x)

### Please describe the expected behavior
if a client is instantiated like so:
```go
agentHeader := http.Header{}
agentHeader.Set("User-Agent", "test/1.0")
ec, err := elastic.NewClient(
	elastic.SetHeaders(agentHeader)

)
```
the user agent `test/1.0` should be sent with all requests

### Please describe the actual behavior
The User-Agent header that's sent will still be set to `"elastic/"+Version+" ("+runtime.GOOS+"-"+runtime.GOARCH+")")`
by [NewRequest](https://github.com/olivere/elastic/blob/60d62e5b2d1c728d7cbbeb7ed9a284303ea4acd4/request.go#L27)

olivere:
Will be solved in 7.0.24.

PR #1442, opened 2020-12-02
Solve "Unable to set User-Agent with `client.SetHeaders`"

olivier4576:
Solves https://github.com/olivere/elastic/issues/1438


olivere:
This has hopefully been fixed in 7.0.24.

Issue #1240, opened 2019-11-21
elastic/go-elasticsearch vs olivere/elastic


harish2048:
I am using Elasticsearch 7.x, I am fairly new to Go, I want to migrate away from python to go and hence would need an Elasticsearch Data Access Layer. I was wondering what the differences were between the official client and this one, and which once is better for the long term.

olivere:
Hi! This library was started around 2012 as an MIT-based, free Go driver for Elasticsearch. I've kept it up-to-date to the best of my capabilities. Earlier this year (I think), the good people at Elastic decided to have an official Elasticsearch client. I've been in contact with @karmi and the team at Elastic developing the official client, and I still am from time to time. So there's no bad vibes or tension between the two projects—quite the opposite tbh.

This unofficial library is more of a DSL than the official driver. E.g. it specifies all requests and responses from ES as Go structs, which are not part of the REST API spec as of now. The official client doesn't go as far. You can use a generic approach instead, like sending a `map[string]interface{}`, or anything serializable to JSON.

I don't know if there will ever be a v8 of olivere/elastic. I'd love to put more time into this project. But as an individual it's hard to keep up with the pace that Elastic, as a company, puts into their product.

Long story short: In the long run, you're probably better off with the official driver. If you're using only a fraction of ES for your project, it can also be a viable approach to skip the driver altogether and simply use the REST API directly.

karmi:
I agree with everything @olivere says above.

I'd be more cautious about this, though:

>  (...) it can also be a viable approach to skip the driver altogether and simply use the REST API directly (...)

From my own experience, I would advise against this. The `elastic/go-elasticsearch` client has a small footprint, and provides a reasonable abstraction for calling the Elasticsearch APIs, without you implementing something like `myclient.Search()` one week, then `myclient.Cluster.Health()` another week, and so on.

(I've been working on an internal service in Go before writing `elastic/go-elasticsearch`, and eg. its ability to retry failed requests is something which the "simple wrapper" I wrote didn't provide, and which was super-important for user experience of said service.)

harish2048:
Thank you both of you, your answers help a lot.

PR #1458, opened 2021-02-03
Add new component template APIs

yann-soubeyrand:
This is an attempt to add the new component template APIs based on the code for the new index template APIs.

### DIFF
--- a/indices_component_templates_test.go
+++ b/indices_component_templates_test.go
@@ -1,0 +1,67 @@
+// Copyright 2012-present Oliver Eilhard. All rights reserved.
+// Use of this source code is governed by a MIT-license.
+// See http://olivere.mit-license.org/license.txt for details.
+
+package elastic
+
+import (
+	"context"
+	"testing"
+)
+
+func TestComponentTemplatesLifecycle(t *testing.T) {
+	client := setupTestClientAndCreateIndex(t)
+
+	const templateName = "template_1"
+
+	// Always make sure the component template is deleted
+	defer func() {
+		_, _ = client.IndexDeleteComponentTemplate(templateName).Pretty(true).Do(context.Background())
+	}()
+
+	// Create an component template
+	{
+		resp, err := client.IndexPutComponentTemplate(templateName).Pretty(true).BodyString(`{
+			"template": {
+				"settings": {
+					"number_of_shards": 2,
+					"number_of_replicas": 0
+				},
+				"mappings": {
+					"_source": { "enabled": true }
+				}
+			}
+		}`).Do(context.Background())
+		if err != nil {
+			t.Fatalf("expected to successfully create component template, got %v", err)
+		}
+		if resp == nil {
+			t.Fatal("expected response on creating component template")
+		}
+		if want, have := true, resp.Acknowledged; want != have {
+			t.Errorf("expected Acknowledged=%v, got %v", want, have)
+		}
+	}
+
+	// Get the component template
+	{
+		resp, err := client.IndexGetComponentTemplate(templateName).Pretty(true).Do(context.Background())
+		if err != nil {
+			t.Fatalf("expected to successfully get component template, got %v", err)
+		}
+		if resp == nil {
+			t.Fatal("expected response on getting component template")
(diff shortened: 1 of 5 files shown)


indices_delete_component_template.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil, err
+	}
+	return ret, nil
+}
+
+// IndicesDeleteComponentTemplateResponse is the response of IndicesDeleteComponentTemplateService.Do.
+type IndicesDeleteComponentTemplateResponse struct {
+	Acknowledged       bool   `json:"acknowledged"`
+	ShardsAcknowledged bool   `json:"shards_acknowledged"`
+	Index              string `json:"index,omitempty"`
```
yann-soubeyrand, line 181: I copied this from indices_delete_index_template.go but I didn't know where to check if these fields can appear in a response (they don't when I do a simple request to the API).

indices_put_component_template.go
```diff
@@ -0,0 +1,221 @@
...
+		return nil, err
+	}
+	return ret, nil
+}
+
+// IndicesPutComponentTemplateResponse is the response of IndicesPutComponentTemplateService.Do.
+type IndicesPutComponentTemplateResponse struct {
+	Acknowledged       bool   `json:"acknowledged"`
+	ShardsAcknowledged bool   `json:"shards_acknowledged"`
+	Index              string `json:"index,omitempty"`
```
yann-soubeyrand, line 220: Same comment as above.

ghost:
@olivere This would be a really nice feature for the terraform provider, could you take a look on this PR, please :)

indices_component_templates_test.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		if resp == nil {
+			t.Fatal("expected response on creating component template")
+		}
+		if want, have := true, resp.Acknowledged; want != have {
+			t.Errorf("expected Acknowledged=%v, got %v", want, have)
+		}
+	}
+
+	// Get the component template
```
ghost, line 46: A Test Case for a missing template would be nice :)

indices_delete_component_template.go
ghost: I guess only the`acknowledged` field is filled here, since we do not change something on indices here

indices_put_component_template.go
ghost: see above

ghost (changes requested):

phillbaker:
@olivere any thoughts on this?

olivere:
This is looking really good. Thank you for working on it.

Did you check that the request parameters are the ones specified by the [REST API](https://github.com/elastic/elasticsearch/tree/v7.11.1/rest-api-spec/src/main/resources/rest-api-spec/api)? I've seen that contributors often start by copying an existing service but then leave a lot of fields that are actually not allowed for the new service, or miss some that are. It's pretty hard to figure out because some fields are allowed on every service, and are still not specified in the REST API spec for that specific service.

It's also worth checking the Java code as the ultimate source of truth, when in doubt. E.g. those [`toXContent`](https://github.com/elastic/elasticsearch/blob/v7.11.1/server/src/main/java/org/elasticsearch/cluster/metadata/ComponentTemplate.java#L129-L141) methods describe how ES serializes the JSON—documentation, while really good, is not enough in many cases.

yann-soubeyrand:
Hi @olivere, sorry for the delayed answer… I think I did some mistakes on the parameters, since, as you mentioned, it’s hard to find the ultimate source of truth.

I’m wondering what’s the status of this project compared to the official Go client for Elasticsearch (https://github.com/elastic/go-elasticsearch)?

olivere:
Hi @yann-soubeyrand! The source of trust the the REST API spec. However, they don't spec out the body of the request/response. That's when I look into the Java source. Those `(from|to)XContent` methods in the `.*Builder` classes are a safe bet.

With regards to the official Elasticsearch client, I've pinned an issue with my perspective on that. See [here](https://github.com/olivere/elastic/issues/1240).

PR #1469, opened 2021-03-20
[xpack role] Fix spelling of resources

phillbaker:
Not sure if this would be considered a breaking change to the API, but thought it may help others.

May be related to https://github.com/phillbaker/terraform-provider-elasticsearch/issues/128


olivere:
Yeah, once I see these things I simply cannot unsee them. So, yeah, it probably is a breaking change, but it generates a compile-time error and takes around 1s to fix.

Thanks for finding.

PR #1472, opened 2021-03-24
Support for new case_insensitive wildcard queries

cacts:
Elasticsearch 7.10.0 added support for case_insensitive wildcard queries:

https://www.elastic.co/guide/en/elasticsearch/reference/7.x/query-dsl-wildcard-query.html

This provides support for the new option, and renames wildcard field to value to match the current API docs. I'm not really sure what the process is for mid-major-version feature support, so let me know if I need to submit this in a different way!


olivere:
That's great. Let me check.

PR #1473, opened 2021-03-25
fix(TotalHits): check for nil on TotalHits() method pointer receiver

qilingzhao:
if `Do`method return a `nil` *SearchResult, call `func (r *SearchResult) TotalHits() int64` method will got a panic.
Adding an if condition to prevent panic caused by nil pointer.


olivere:
Yeah, this will make it more robust. Thanks.

However, I can't see why this might be happening. Because `SearchResult` might only be `nil` if you're forgetting to check for an error in the Search API call.

PR #1478, opened 2021-03-29
Add stream option to PerformRequest()

### LINKED ISSUE #1444: Direct access to HTTP response
Background: We're using the client in an application where concurrent memory usage is a significant concern, and queries can return many megabytes of data in some cases. To reduce memory usage and increase pipelining of throughout our query handling, we've developed a solution where we stream the byte stream from Elasticsearch instead of unmarshalling of the entire response into a struct.

This works really well, but we had to use a custom JSON parser to accomplish this, and we also needed direct access to the HTTP response instead of getting a struct back. Unfortunately, the client doesn't permit this; we encountered some roadblocks:

* Every `Do()` method unmarshals the response.
* The client has a `PerformRequest()`, but it reads the entire body and does not return a stream.
* There's no way to ask a client for its configuration so that one can set up a new request
* All the logic to build an HTTP request are buried in the `Do()` methods

We ended up creating a "wrapper client" that contains an `*elastic.Client`, but also carries around with it the URL of the server and some other configuration. Then we use `net/http` to make an HTTP request, bypassing the client altogether. Fortunately, we are able to reuse structs like `elastic.SearchRequest`, but we have to construct the HTTP envelope (e.g. `scroll`) ourselves. We also have to do things like URL construction, logging, status handling, retrying, etc. ourselves.

Here's a snippet from our current logic, greatly simplified:

```go
b, err := search.Body()  // search is a *elastic.SearchRequest we've prepared
if err != nil {
	return err
}

url := baseURL
url.Query = url.Values{"scroll": []string{keepAlive}}

req, err := http.NewRequest(http.MethodPost, url.String(), bytes.NewReader(b))
if err != nil {
	...
}

resp, err := http.DefaultClient.Do(ctx, req)
if err != nil {
	...
}

return newJSONStream(resp.Body)
```

It would be much better if we could do something like:

```go
scroll := client.Scroll().Index(indeXName).KeepAlive(keepAlive) // etc.

resp, err := scroll.DoStream(ctx)  // Returns an *http.Response
if err != nil {
	...
}

return newJSONStream(resp.Body)
```

Or at least something like:

```
resp, err := client.PerformRequest(ctx, elastic.PerformRequestOptions{
	Body: search,
	Params: url.Values{"scroll": []string{keepAlive}},
	Stream: true,  // Signals that body should be returned as a stream
})
if err != nil {
	...
}

return newJSONStream(resp.Body)
```

Thoughts?

atombender:
Add an option `Stream` to `PerformRequestOptions`, allowing a response to be parsed as a stream. Fixes #1444.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1286,6 +1286,7 @@ type PerformRequestOptions struct {
 	RetryStatusCodes []int
 	Headers          http.Header
 	MaxResponseSize  int64
+	Stream           bool
 }
 
 // PerformRequest does a HTTP request to Elasticsearch.
@@ -1294,6 +1295,9 @@ type PerformRequestOptions struct {
 // Optionally, a list of HTTP error codes to ignore can be passed.
 // This is necessary for services that expect e.g. HTTP status 404 as a
 // valid outcome (Exists, IndicesExists, IndicesTypeExists).
+//
+// If Stream is set, the returned BodyReader field must be closed, even
+// if PerformRequest returns an error.
 func (c *Client) PerformRequest(ctx context.Context, opt PerformRequestOptions) (*Response, error) {
 	start := time.Now().UTC()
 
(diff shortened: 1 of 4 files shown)


olivere:
That's looking good. I like it. 👍 

Let me check a few things, but it's likely to get into the next release.

atombender:
Glad to hear it!

atombender:
Awesome, thanks!

PR #1486, opened 2021-05-02
feat(_cat/master) Add CatMaster API

qilingzhao:
This commit adds the Cat Master API, test cases and update README.md


olivere:
Thank you.

PR #1493, opened 2021-05-21
search: specify the length when declaring the slice

bestgopher:
Specify the length when declaring the slice to reduce the number of memory allocations.


olivere:
Thanks.

PR #1497, opened 2021-05-28
Match boolean prefix query

mbalabin:
Implements "match boolean prefix" query as described in:

https://www.elastic.co/guide/en/elasticsearch/reference/7.13/query-dsl-match-bool-prefix-query.html


olivere:
Thank you. Will be part of 7.0.25.

PR #1499, opened 2021-06-05
Add support for multi terms aggregation

lechnertech:
https://www.elastic.co/guide/en/elasticsearch/reference/7.12/search-aggregations-bucket-multi-terms-aggregation.html


olivere:
I changed a few bits, but the core is intact. Thanks. Will be released with the next version.

PR #1512, opened 2021-07-20
Implement Combined Fields Query

mbalabin:
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-combined-fields-query.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,8 @@ Here are a few tips on how to get used to Elastic:
   - [x] Common Terms Query
   - [x] Query String Query
   - [x] Simple Query String Query
+  - [x] Combined Fields Query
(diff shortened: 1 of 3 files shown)


olivere:
Thanks. Will need to reformat the code and check if the Java source has some additional properties. Otherwise: looking good. :-)

mbalabin:
I did go fmt. It would be handy if CI on github checked formatting though.

I am not a Java developer, but as far as I understand implementation in `elasticsearch/server/src/main/java/org/elasticsearch/index/query/CombinedFieldsQueryBuilder.java` all the parameters are implemented.

olivere:
Thank you.

Issue #1524, opened 2021-08-17
Allow not to extend keep alive of Point in Time


pongraczgabor87:
### Which version of Elastic are you using?

[x] elastic.v7 (for Elasticsearch 7.x)
[ ] elastic.v6 (for Elasticsearch 6.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v2 (for Elasticsearch 1.x)

### Please describe the expected behavior
Based on the elasticsearch [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/point-in-time-api.html#point-in-time-keep-alive) during the `_search` the `keep_alive` parameter is optional, it's only needed if you would like to extend the TTL of the given point in time.

### Please describe the actual behavior
- I couldn't leave out the `keep_alive` from the request.     
- If I try to set it to empty string I receive an error from elasticsearch: `Error 400 (Bad Request): [1:762] [pit] failed to parse field [keep_alive] [type=x_content_parse_exception]`.    
- If I try to set it to 0, something strange happens in the elasticsearch, the PIT is not removed as expected.

### Any steps to reproduce the behavior?
```
searchResult, err := client.Search().
		// Index(testIndexName). // <-- you may not use indices with PointInTime!
		Query(NewMatchAllQuery()).
		PointInTime(NewPointInTime(pitResp.Id, "")). // or "0m", "0s", etc...
		Size(100).
		Pretty(true).
		Do(context.TODO())
```

### Proposal
Make `keep_alive` optional.

olivere:
Yep. Fixed in 7.0.28. Thanks.

PR #1525, opened 2021-08-17
Fix point in time keep alive handling

pongraczgabor87:
Relates https://github.com/olivere/elastic/issues/1524


olivere:
Fixed in 7.0.28.

PR #1529, opened 2021-08-31
Master

superb80s:


superb80s:
getSource

superb80s:
getSource

olivere:
You don't want to do it this way. If we need the `SearchSource`, you [create one](https://github.com/olivere/elastic/blob/release-branch.v7/search_source.go#L49-L57) and [pass it into the `SearchService`](https://github.com/olivere/elastic/blob/release-branch.v7/search.go#L104-L111). It's the other way around.

superb80s:
thx

PR #1538, opened 2021-10-06
Set ID in the struct if structure passes SetIDInterface test.

sergeyglazyrindev:
Hey @olivere 
I am working on admin framework for go.
And now I want to develop solution for admin part of the project to read/edit/add new documents to ES index.
For this, I need to know ID of the document, so I added corresponding fix for your package.
Lemme know what do you think.

### DIFF
--- a/client.go
+++ b/client.go
@@ -26,7 +26,7 @@ import (
 
 const (
 	// Version is the current version of Elastic.
-	Version = "7.0.29"
+	Version = "7.0.31"
 
 	// DefaultURL is the default endpoint of Elasticsearch on the local machine.
 	// It is used e.g. when initializing a new Client without a specific URL.
--- a/client.go
+++ b/client.go
@@ -26,7 +26,7 @@ import (
 
 const (
 	// Version is the current version of Elastic.
-	Version = "7.0.31"
+	Version = "7.0.29"
 
 	// DefaultURL is the default endpoint of Elasticsearch on the local machine.
 	// It is used e.g. when initializing a new Client without a specific URL.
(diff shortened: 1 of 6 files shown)


olivere:
May I ask what problem this PR solves?

olivere:
This won't be merged, I'm sorry. If necessary at all, it should be solved in the application code, not in this library.

sergeyglazyrindev:
@olivere , how do I know ID of the record from ElasticSearch Index if I use Each method ?

olivere:
The [ID is per `Hit`](https://github.com/olivere/elastic/blob/9e03315e5dfecde4caaf8deb49e76dbe74a19026/search.go#L763), and is stored inside Elasticsearch [in the `_id` field](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-id-field.html).

sergeyglazyrindev:
ok, but this ID is not available in the structure that I got using Each method. But ideally it should available, no ?

olivere:
No, it can't, because it only deserializes the `_source` property. If you want the ID, you shouldn't use `Each`, but do it like this:

```go
for _, hit := range searchResponse.Hits.Hits {
	var v YourEntity
	if err := json.Unmarshal(hit.Source, &v); err != nil {
		// Handle error
	}
	// Use v
}
```

`Each` is just for convenience.

sergeyglazyrindev:
ah, thanks for explanation.

olivere:
BTW: If you accept non-autogenerated IDs (and the slight overhead), you can of course also add an `ID` field to your entity and use its value as the `_id` as well. ES will only create an `_id` for you if you don't give one to it. Then `Each` would also work.

sergeyglazyrindev:
yeah, thx man. Sorry, was like blind. Focused on Each method....

PR #1542, opened 2021-10-14
Add context to logger

### LINKED ISSUE #1541: logger with context ?

[ x ] elastic.v7 (for Elasticsearch 7.x)
[ ] elastic.v6 (for Elasticsearch 6.x)

Can Logger support Context ? Because it's usefully for tracing，add requestId etc for a request.

olivere:
This commit adds a `LoggerWithContext` interface that extends the
`Logger` interface by a method `PrintfWithContext` that, when
implemented, is called instead of the `Printf` method of the `Logger`
interface.

The purpose of `PrintfWithContext` is to receive the current context
under which the logging happens. Notice that this doesn't always
have to be request-scoped, i.e. an actual API call from a user. It may
also be from an internal state or process, e.g. Bulk processor or node
health.

Close #1541

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -4,7 +4,15 @@
 
 package elastic
 
+import "context"
+
 // Logger specifies the interface for all log operations.
 type Logger interface {
 	Printf(format string, v ...interface{})
+}
+
+// LoggerWithContext extends the Logger interface by a context.
+type LoggerWithContext interface {
+	Logger
+	PrintfWithContext(ctx context.Context, format string, v ...interface{})
 }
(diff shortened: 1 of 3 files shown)


logger.go
```diff
@@ -4,7 +4,15 @@
...
+import "context"
+
 // Logger specifies the interface for all log operations.
 type Logger interface {
 	Printf(format string, v ...interface{})
 }
+
+// LoggerWithContext extends the Logger interface by a context.
+type LoggerWithContext interface {
+	Logger
```
lichunqiang, line 16: I don't think about LoggerWithContext implements Logger interface is necessarily.

olivere: Otherwise you can’t specify the logger in a backwards compatible manner with SetErrorLog etc.

lichunqiang: you are right :)

Issue #1459, opened 2021-02-05
OpenTracing spans reveal URL auth credentials


atombender:
### Which version of Elastic are you using?

[X] elastic.v7 (for Elasticsearch 7.x)
[ ] elastic.v6 (for Elasticsearch 6.x)
[ ] elastic.v5 (for Elasticsearch 5.x)
[ ] elastic.v3 (for Elasticsearch 2.x)
[ ] elastic.v2 (for Elasticsearch 1.x)

### Please describe the expected behavior

We expected the spans reported to hide the user name and password.

### Please describe the actual behavior

Full URL was exposed. Example from Jaeger UI (scrubbed):

![Screen Shot 2021-02-05 at 08 55 09](https://user-images.githubusercontent.com/50314/107042732-129a8480-6790-11eb-8eb8-d54eba8326ad.jpg)

### Any steps to reproduce the behavior?

Set the client up with a URL with a user name and password.

I'll see about creating a PR for this.

olivere:
Oops. That's a bug for sure.

olivere:
I improved tests in 180a7ca4b8668f948a4da5a31f641f7c78eabb85 but cannot reproduce. Am I missing something?

atombender:
Great. Just to be clear, that logging goes to the OpenTracing span?

olivere:
Whoops. I'm sorry...

olivere:
Have to double-check.

olivere:
I'll give it another go in 7.0.31. It replaced the occurrence of `req.URL.String()` with `req.URL.Redacted()` which hopefully, finally, fixes this issue. Let me know if it still happens to be logged. I do not doubt it happens, but I'm still not sure how to reproduce it.

atombender:
Thank you!

PR #1559, opened 2021-12-13
Redact password from URL in infof logging

igorwwwwwwwwwwwwwwwwwwww:
The infof call that logs requests currently leaks the basic auth password through the URL. By using [URL.Redacted()](https://pkg.go.dev/net/url#URL.Redacted), we can replace the password with "xxxxx" in the log message.


igorwwwwwwwwwwwwwwwwwwww:
There may be a similar issue in https://github.com/olivere/elastic/issues/1459.

olivere:
Thanks for this. 🙏 Will merge it in the next release.

PR #1550, opened 2021-11-17
[index template] Parse datastream object

phillbaker:
Briefly mentioned in https://github.com/olivere/elastic/issues/1464 and related to https://github.com/olivere/elastic/issues/1484, but we saw this in https://github.com/phillbaker/terraform-provider-elasticsearch/issues/159. 

The data_stream element isn't really documented well by elasticsearch, the only docs I could find were tangential in the simulate API here: https://www.elastic.co/guide/en/elasticsearch/reference/7.12/indices-simulate-template.html#simulate-template-api-request-body. This seems to match this code in ES: https://github.com/elastic/elasticsearch/blob/9958c3c2fc49a3e253b04e96cfec2e653c39d2e7/client/rest-high-level/src/main/java/org/elasticsearch/client/indices/DataStream.java#L22

PR #1570, opened 2022-01-13
[index template] Fix data stream timestamp field

phillip-dd:
Related to https://github.com/olivere/elastic/pull/1550 and https://github.com/olivere/elastic/commit/3e406842c91fb5f7dfd1904dff173db10440983e.

Although this is a just a string on the Java object, it's expressed as an object in the JSON api, see: https://github.com/elastic/elasticsearch/blob/5c4330292ffead1218ddba896431ca04408f766c/client/rest-high-level/src/main/java/org/elasticsearch/client/indices/DataStream.java#L162

The documentation concurs: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-data-stream.html#get-data-stream-api-example

(Looks like I missed this in my original PR as well)

Edit: Sorry, opened with work account instead of personal.

### DIFF
--- a/indices_get_index_template.go
+++ b/indices_get_index_template.go
@@ -231,16 +231,20 @@ type IndicesGetIndexTemplateData struct {
 }
 
 type IndicesDataStream struct {
-	Name               string                 `json:"name,omitempty"`
-	TimestampField     string                 `json:"timestamp_field,omitempty"`
-	Indices            []string               `json:"indices,omitempty"`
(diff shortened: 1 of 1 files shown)


olivere:
Yeah, looks like it. Will do this in 7.0.32.

phillbaker:
👋 @olivere just checking in on this, would you be able to review? Related to https://github.com/phillbaker/terraform-provider-elasticsearch/issues/159

PR #1576, opened 2022-01-27
fix: keep sort in search hit as is

houz42:
'sort' in search hit should be kept as is.

In some cases, when search documents while sort by an int64 field (date_nano for example), server returns a sort value in each hit, which contains a long number. If we represent sort as []interface{}, the long number will be treated as float64 by std json decoder, and if the number is great enough (and date_nano value of tody is great enough), we will get an incorrect int64 value in go.

And when the incorrect sort value is used for next search reqeuest (set as search_after parameter), we may see duplicated or missing documents in pagingated search requests.

To keep backward compatable, I did not remove or change the type of `Sort` field in `SearchHit`, but add a new field `RawSort` as `[]json.RawMessage` and json unmarshaller instread.

### DIFF
--- a/search.go
+++ b/search.go
@@ -767,7 +767,8 @@ type SearchHit struct {
 	Version        *int64                         `json:"_version,omitempty"` // version number, when Version is set to true in SearchService
 	SeqNo          *int64                         `json:"_seq_no"`
 	PrimaryTerm    *int64                         `json:"_primary_term"`
-	Sort           []interface{}                  `json:"sort,omitempty"`            // sort information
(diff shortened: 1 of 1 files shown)


houz42:
@olivere I have found an issue on search request and proposed a quick fix, may you be available to have a look? thanks.

olivere:
I don't like to deprecate the `Sort` field like that, but I see your point.

RaphaelL2e:
I think can replcae json.unmarshal to json.decoder

RaphaelL2e:
@houz42 Ask a question, I used this mr but the order is still wrong.

PR #1577, opened 2022-01-27
Fix: PIT Search Recipe NewPointInTime Call

grahamplace:
In [this commit](https://github.com/olivere/elastic/commit/d0ea877a6e58d9b607e4216f928f74e65af0921c), `keepAlive` was removed from `NewPointInTime` (and instead supported by `NewPointInTimeWithKeepAlive`), but the PIT Search recipe was not updated to reflect that change


olivere:
Thanks.

Issue #1600, opened 2022-03-16
 BottomLeft and TopRight combo is broken for BoundingBoxQuery


elmir-b:
Please use the following questions as a guideline to help me answer
your issue/question without further inquiry. Thank you.

### Which version of Elastic are you using?

[X] elastic.v7 (for Elasticsearch 7.x)

### Please describe the expected behavior
After upgrading from 7.0.22 to 7.0.31 BottomLeft and TopRight combo is broken for BoundingBoxQuery. 

I would expect that I can use the following to get proper BoundingBox:
`elastic.NewGeoBoundingBoxQuery("location.coordinates").
BottomLeft(float64(box.BottomLeft.Lat), float64(box.BottomLeft.Lon)).
TopRight(float64(box.TopRight.Lat), float64(box.TopRight.Lon)),`

### Please describe the actual behavior
I don't get the desired bounding box but a much larger area. It seems to make a switch between lat and lon somewhere when it tries to convert between BottomLeft and TopRight combo to TopLeft and BottomRight combination

### Any steps to reproduce the behavior?
Just use the code from above. If I do the conversion between the two myself it works fine, i.e.:
`elastic.NewGeoBoundingBoxQuery("location.coordinates").
TopLeft(float64(box.TopRight.Lat), float64(box.BottomLeft.Lon)).
BottomRight(float64(box.BottomLeft.Lat), float64(box.TopRight.Lon)),`

olivere:
At first sight, I can't see what's wrong. Can you write a test to complete the [test suite](https://github.com/olivere/elastic/blob/release-branch.v7/search_queries_geo_bounding_box_test.go) that illustrates the problem?

elmir-b:
> At first sight, I can't see what's wrong. Can you write a test to complete the [test suite](https://github.com/olivere/elastic/blob/release-branch.v7/search_queries_geo_bounding_box_test.go) that illustrates the problem?

Hi, so I think this test should show it:
```
func TestGeoBoundingBoxQueryBottomLeftTopRight(t *testing.T) {
	q := NewGeoBoundingBoxQuery("pin.location")
	q = q.BottomLeft(48.010597, 11.486893)
	q = q.TopRight(48.14822, 11.650314)
	q = q.Type("memory")
	src, err := q.Source()
	if err != nil {
		t.Fatal(err)
	}

	data, err := json.Marshal(src)
	if err != nil {
		t.Fatalf("marshaling to JSON failed: %v", err)
	}
[...]

PR #1610, opened 2022-04-14
Fixed GeoBoundingBox BottomLeft args order

pistatium:
Hello.

The order of BottomLeft arguments in GeoBoundingBoxQuery was swapped, which has been corrected.

before: `"bottom_left":[40.01,-71.12],"top_right":[-74.1,40.73]}`

after: `bottom_left":[-71.12,40.01],"top_right":[-74.1,40.73]}`

### DIFF
--- a/search_queries_geo_bounding_box.go
+++ b/search_queries_geo_bounding_box.go
@@ -65,7 +65,7 @@ func (q *GeoBoundingBoxQuery) BottomRightFromGeoHash(bottomRight string) *GeoBou
 
 // BottomLeft position from longitude (left) and latitude (bottom).
 func (q *GeoBoundingBoxQuery) BottomLeft(bottom, left float64) *GeoBoundingBoxQuery {
-	q.bottomLeft = []float64{bottom, left}
+	q.bottomLeft = []float64{left, bottom}
 	return q
 }
 
--- a/search_queries_geo_bounding_box_test.go
+++ b/search_queries_geo_bounding_box_test.go
@@ -29,6 +29,26 @@ func TestGeoBoundingBoxQuery(t *testing.T) {
 	}
 }
 
+func TestGeoBoundingBoxQueryInverted(t *testing.T) {
+	q := NewGeoBoundingBoxQuery("pin.location")
+	q = q.TopRight(40.73, -74.1)
+	q = q.BottomLeft(40.01, -71.12)
+	q = q.Type("memory")
+	src, err := q.Source()
+	if err != nil {
+		t.Fatal(err)
+	}
+	data, err := json.Marshal(src)
+	if err != nil {
+		t.Fatalf("marshaling to JSON failed: %v", err)
+	}
+	got := string(data)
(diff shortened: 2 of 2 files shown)


olivere:
@pistatium Can you check if that fixes all cases as described in #1600?

pistatium:
I see...

`bottom_right` is used in the document in the order `lat,lon`.

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-bounding-box-query.html#_lat_lon_as_array_2

So this seems to be the correct order for `BottomLeft`.
`BottomLeft` is also called from `BottomRightFromGeoPoint` and `BottomRightFromGeoHash`, but this one is used just like the other methods and has no problem.

 Please let me know if there are any other assumptions that need to be made.

chchaffin:
Won't this fix break existing clients?

search_queries_geo_bounding_box.go
```diff
@@ -65,7 +65,7 @@ func (q *GeoBoundingBoxQuery) BottomRightFromGeoHash(bottomRight string) *GeoBou
 
 // BottomLeft position from longitude (left) and latitude (bottom).
 func (q *GeoBoundingBoxQuery) BottomLeft(bottom, left float64) *GeoBoundingBoxQuery {
-	q.bottomLeft = []float64{bottom, left}
+	q.bottomLeft = []float64{left, bottom}
```
chchaffin, line 68: Clients, like me, already swapped this and are using this in production. This fix will break unsuspecting clients that think this is a bug fix. If this change is accepted then it would warrant a major version change.

olivere: I think this falls into the bucket of a very unfortunate bug on my side. However, if you were to fix all of these bugs by creating a new major version, we would probably be way beyond version 100 now. So I think we will fix it in a minor version.

If you fixed it, you are probably aware of the bug. Thankfully, Go modules won't silently update your dependency to the latest version automatically. However, I would be very thankful for a way to post a warning when the Go module is updated, but I'm not aware of anything like that. Changelogs are good, but people generally don't read them. A final option would be to remove `BottomLeft` and replace it with e.g. `ButtomLeft2` to make people aware of the issue, but that also seems rather awkward to me.

So: It will be a minor version update, and watch out when you update.

chchaffin: I think awkward beats breaking backwards compatibility, but that's just my opinion. I'd opt to make a new exposed function and deprecate the old one. This will certainly be a headache for some one down the road otherwise. It's your repo, so your call.

jizhihaoSAMA: When do we merge this PR? Should I use the older version?

jonathansharman: I second @chchaffin - deprecate the old function and fix the bug in a new function. This is technically a bug fix, but anyone still using the current, bugged version of this function _will_ have swapped the order to compensate, so there's going to be a lot more breakage if this goes out as a minor or patch version.

PR #1627, opened 2022-06-27
fix range query use old dsl isses

RyouZhang:


RyouZhang:
@olivere please check this pr

mitar:
I think this is duplicate of #1453?

PR #1635, opened 2022-07-23
Removed Redacted() function from req.url

javed2214:
Removed Redacted() function from from req.URL as it is creating issue while creating the go container. Getting compilation error while building the app. Seems like it doesn't support newer version of Go. My current version is 1.18. This issue is facing by a lot of users. Kindly fix that.


mitar:
It seems that it exists since 1.15: https://pkg.go.dev/net/url#URL.Redacted

And it compiles for me using Go 1.18 without this PR. So I am not sure why you are getting your error. Maybe you have too old Go version when building the container?

PR #1661, opened 2023-02-02
Fix: Bulk processor retries indefinitely on failure

### LINKED ISSUE #1278: Bulk processor retries indefinitely on failure

- [ ] elastic.v7 (for Elasticsearch 7.x)
- [x] elastic.v6 (for Elasticsearch 6.x)
- [ ] elastic.v5 (for Elasticsearch 5.x)
- [ ] elastic.v3 (for Elasticsearch 2.x)
- [ ] elastic.v2 (for Elasticsearch 1.x)

This issue is similar to #1247 but with a bigger scope.

When an individual request fails in a bulk processor, it should not retry when the backoff returns `false`.

When the backoff returns `false` and there is an error in the individual request , the failing request is added back to the bulkprocessor before the commit function returns. This causes the indefinite retries.

The following code in bulk_processor.go is responsible for this behaviour. 

<pre>
// commitFunc will commit bulk requests and, on failure, be retried
// via exponential backoff
commitFunc := func() error {
    var err error
    // Save requests because they will be reset in service.Do
    reqs := w.service.requests
    res, err = w.service.Do(ctx)
    if err == nil {
	// Overall bulk request was OK.  But each bulk response item also has a status
	if w.p.retryItemStatusCodes != nil && len(w.p.retryItemStatusCodes) > 0 {
		// Check res.Items since some might be soft failures
	    if res.Items != nil && res.Errors {
		// res.Items will be 1 to 1 with reqs in same order
		for i, item := range res.Items {
		    for _, result := range item {
	                if _, found := w.p.retryItemStatusCodes[result.Status]; found {
			    <b>w.service.Add(reqs[i]) //here the failing request is added back to the processor regardless of the backoff flag</b>
			    if err == nil {
				err = ErrBulkItemRetry
	                     }
			}
	            }
		}
	    }
        }
    }
    return err
}
</pre>

One easy way to recreate this is to retry on client side errors, and send wrong data type to a field in the index.

rafaeljusto:
When all retries are exhausted the worker internal requests buffer needs to be cleared in failure scenarios. That is required because the `commitFunc` (and consequently the underlying `BulkService.Do` call) doesn't reset it when some error happens. Without clearing the internal buffer the worker will continue sending the same requests on the following rounds of execution.

Kudos for this solution goes to @rwynn and @raiRaiyan .

Resolves #1278

### DIFF
--- a/bulk_processor.go
+++ b/bulk_processor.go
@@ -580,6 +580,14 @@ func (w *bulkWorker) commit(ctx context.Context) error {
 	err := RetryNotify(commitFunc, w.p.backoff, notifyFunc)
 	w.updateStats(res)
 	if err != nil {
+		// After all retry attempts clear the requests for the next round. This is
+		// important when backoff is disabled or limited to a number of rounds, and
+		// the aftermath is a failure., because the commitFunc (and consequently the
+		// underlying BulkService.Do call) will not clear the requests from the
+		// worker. Without this the same requests will be used on the next round of
+		// execution of the worker.
+		w.service.Reset()
+
 		w.p.c.errorf("elastic: bulk processor %q failed: %v", w.p.name, err)
 	}
 
--- a/bulk_processor.go
+++ b/bulk_processor.go
@@ -580,12 +580,12 @@ func (w *bulkWorker) commit(ctx context.Context) error {
 	err := RetryNotify(commitFunc, w.p.backoff, notifyFunc)
 	w.updateStats(res)
 	if err != nil {
-		// After all retry attempts clear the requests for the next round. This is
-		// important when backoff is disabled or limited to a number of rounds, and
-		// the aftermath is a failure., because the commitFunc (and consequently the
-		// underlying BulkService.Do call) will not clear the requests from the
-		// worker. Without this the same requests will be used on the next round of
-		// execution of the worker.
+		// After exhausting all retry attempts, the worker must clear the requests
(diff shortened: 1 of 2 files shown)


rafaeljusto:
Hey @olivere , as this repository is deprecated, is it not accepting patches anymore?

olivere:
Yes. This repository is deprecated and basically dormant. You are strongly encouraged to switch to the official client.

jstasiak:
I suspect merging this PR would solve this problem in Jaeger: https://github.com/jaegertracing/jaeger/issues/2192.

olivere:
@jstasiak I understand your frustration. But please see it through my eyes. My decision to stop this project was to send a clear signal to the users of this library to move to the official client supported by Elastic, the company. They took over years ago, now. Sure, I could merge the PR. But that might open up a new stream of changes, or have side-effects that I can't foresee and don't have the energy to work on any more.

Another choice for jaeger would be to fork this repository and keep their own version of the library. I'm of course perfectly OK with this.

Thanks for your understanding.

jstasiak:
Hey @olivere! I totally understand this repository is not maintained and I didn't want to appear as if I was putting any pressure on anyone or expecting anything - I'm not. No need to explain anything.

I left links with notes here and in #1278 because I thought they could be useful to someone else investigating this. Not much frustration here, just sharing what I found.

Thank you for all your work on this and have a great day.

PR #1668, opened 2023-05-03
fix: attempt to retry and mark connection as dead

mwain:
When nodes are being shutdown quickly we have a window when a connection is alive in the pool, but erroring when a request is performed.

It does retry, however there is an edge case on the next pass a connection could be chosen which should be dead.

This change alters the retry logic, when an error is returned from the retrier and continue == true, it will mark the connection as dead and retry.

### DIFF
--- a/client.go
+++ b/client.go
@@ -160,9 +160,9 @@ type Client struct {
 //
 // Example:
 //
-//   client, err := elastic.NewClient(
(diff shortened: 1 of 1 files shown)


olivere:
I appreciate your time and effort to push this PR. However, this library is deprecated for over a year now, and won't be worked on. Please move to the [official Elasticsearch client for Go](https://github.com/elastic/go-elasticsearch). Thanks for your understanding.

Issue #1669, opened 2023-05-19
Support for Version=0 in bulk_delete_request.go


garrison-stauffer:
### Which version of Elastic are you using?

- [X] elastic.v7 (for Elasticsearch 7.x)
- [ ] elastic.v6 (for Elasticsearch 6.x)
- [ ] elastic.v5 (for Elasticsearch 5.x)
- [ ] elastic.v3 (for Elasticsearch 2.x)
- [ ] elastic.v2 (for Elasticsearch 1.x)

### Please describe the expected behavior
When we are using the Bulk API to issue delete requests, we should be able to use a `version_type=EXTERNAL_GTE` with `version=0`

### Please describe the actual behavior
When we use this in the Bulk API, the version is not being forwarded to ES and is having `-3` as its default (the `MATCH_ANY` behavior). This leads to an error being returned by ElasticSearch: 
```
elastic: Error 400 (Bad Request): Validation Failed: 1: illegal version value [-3] for version type [EXTERNAL_GTE]; [type=action_request_validation_exception]
```

When I curl elasticsearch directly, it is able to support Version=0:
```bash
$ curl https://my-elasticsearch-cluster.com/_bulk \
-H 'content-type: application/json' \
-X POST \
-d '{"delete": {"_index": "my-index", "_id": "my_id", "version": 0, "version_type": "external_gte"} }
'
{"took":13,"errors":false,"items":[{"delete":{"_index":"my-index","_type":"_doc","_id":"my_id","_version":0,"result":"deleted","_shards":{"total":4,"successful":3,"failed":0},"_seq_no":1,"_primary_term":3,"status":200}}]}
```

### Any steps to reproduce the behavior?
The code is pretty messy, but we're basically just sending bulk requests to create + delete a record. Example code for the request:
```golang
var bulkService *elastic.BulkService // defined elsewhere

// Add a new document at Version=0 and VersionType=External
bulkService.Add(
	elastic.NewBulkIndexRequest().
		Index("some-index").
		Id("some-id").
		Doc(stuff).
		Version(0).
		VersionType("external").
		RetryOnConflict(0)
).Do(ctx)

// Now delete that document referencing Version=0 and VersionType=External_GTE
bulkService.Add(
        elastic.NewBulkDeleteRequest().
		Index("some-index").
		Id("some-id").
		Version(0).
		VersionType("external_gte")
).Do(ctx)
```

[...]

Issue #1533, opened 2021-09-22
Announcement: Future directions (updated on 2022-07-22)


olivere:
Writing this in an issue is probably not the right way, but at least it's near the artifact that matters most to me—the code.

Disclaimer: Please consider this as my very own personal view. I'm a complete outsider, have no internal information and no investments in either Amazon or Elastic. I'm just a dev with an opinion.

I've started this project in 2012. First commit was on December 6, 2012. That was the same year that [Searchworkings Global](https://www.sec.gov/Archives/edgar/data/1707753/000119312518266861/d588632ds1.htm), the predecessor of the company we now know as Elastic, was founded, and way before Elasticsearch 1.0.0 was a thing (that was [Feb 12, 2014](https://en.wikipedia.org/wiki/Elasticsearch#Release_history)). Since then this library supported all versions of Elasticsearch with the help of many [contributors](https://github.com/olivere/elastic/blob/release-branch.v7/CONTRIBUTORS), but also because I ❤️ Elasticsearch and Go and I was using it extensively in my projects and at work. I've probably invested thousands of hours into building and supporting it. I've tried very hard to not break things. Elasticsearch has never disappointed me (and it [made us faster—literally](https://www.elastic.co/blog/how-elasticsearch-made-us-faster-literally) ;-)). My deepest respect and thanks to all the developers that made it such a success (and good companion). But in the end, I think, I wrote the library for myself—because I'm a techie at heart.

Over the recent months though, a strange controversy arose between Elastic and Amazon over the use of Elasticsearch, the technology.

First of all I understand both sides. But still I'm disappointed about the end result for us developers. On the one hand, there's Amazon offering Elasticsearch under their brand, now forking and rebranding it under the term [OpenSearch](https://aws.amazon.com/de/blogs/opensource/introducing-opensearch/). As a developer, I think Amazon shouldn't have done it the way they did. But it's Amazon. You and me have to decide if their way of doing business is having a positive impact on all of us in the long run. On the other hand, in a move to protect their stakes, Elastic [introduced a new license](https://www.elastic.co/blog/why-license-change-AWS) and made [technical changes](https://github.com/elastic/go-elasticsearch/commit/e57702f714d6ecf810150bd13fbe281a78403e82) to their API clients. Although it's quite easy to work around it, again—as a developer—I can't see this heading into the right direction.

[...]

PR #1674, opened 2023-07-20
ISSUES-1669: Allow 0 value for Version in Bulk Delete Request

garrison-stauffer:
This change addresses the issue outlined in https://github.com/olivere/elastic/issues/1669

0 is a valid value for the external version, but when submitting it for bulk deletes - it is not serialized in json. This PR brings the BulkDeleteRequest to parity with the BulkIndexRequest which was updated in [this commit](https://github.com/olivere/elastic/commit/07d32edbf0aca2e283726a15f8c2a663d3ae202e).


garrison-stauffer:
@olivere apologies for another tag, but this we'd really like to get this change merged in if possible, but acknowledge that this repo is deprecated

olivere:
@garrison-stauffer Sorry for the late reply. I see what you're after, and if this project would really be actively maintained, I would merge this in a second. However, I've moved on, and doing so would send the signal that this project is still maintained, and people will join the queue and ask for their features to be added (because: "After all, it's just a simple change.").

I've dedicated 7 years of free time from my life to maintain this client, and I'm 💔 to tell people to move on to the official client. But it is like it is: I don't have the capacity to work on this and keep-up or even compete with the Elastic people. So your best bet short-term is to maintain your own fork of the client, and change your code to use the official Go client long-term.

Again, I'm sorry.

P.S.:
See [here](https://github.com/olivere/elastic/issues/1533) for a longer discussion in the past.

Edit: typos.

garrison-stauffer:
Hey, just wanted to say thanks for the response. I understand the direction, and while (selfishly) unfortunate, I totally get it.

I appreciate all the work put into this library, easily the best go integration for Elasticsearch by far, and I'm happy to see that its living on through the ES library. Hope we can get some guarantees about where Opensearch will/wont diverge so we'd feel more comfortable using the client going forward, but alas.

Issue #1696, opened 2025-10-27
Bug in v7.0.32: GeoBoundingBoxQuery.BottomLeft() swaps latitude and longitude parameters


wstriveh:
## Description

In version **v7.0.32**, the `BottomLeft()` method in `GeoBoundingBoxQuery` has a critical bug where it swaps the latitude and longitude parameters, causing incorrect geo bounding box queries and Elasticsearch errors.

## Version

- **elastic/v7: v7.0.32**

## Problem

In the file `search_queries_geo_bounding_box.go` (lines 68-71), the `BottomLeft` method stores parameters in the wrong order:

**Example coordinates:**
- TopRight: latitude=22.559341, longitude=113.946491
- BottomLeft: latitude=22.507294, longitude=113.914329

Due to the bug, BottomLeft is sent as `[22.507294, 113.914329]` (lat, lon) instead of `[113.914329, 22.507294]` (lon, lat), making Elasticsearch think the bounding box is invalid.

PR #1697, opened 2025-12-04
fix: BottomLeftFromGeoPoint flipping lat and lng values

danielyuenhx:
using `BottomLeftFromGeoHash` results in error as the lat lng values are flipped. this PR reverts it to the correct ordering


danielyuenhx:
fixes https://github.com/olivere/elastic/issues/1696

